PR #145, opened 2022-01-24
Adding batch-predict

### LINKED ISSUE #143: [Models lib] Batch-predict

marcromeyn:
Closes #143


merlin_models/tf/core.py
```diff
@@ -2270,14 +2337,137 @@ def test_step(self, inputs):
...
+    def batch_predict(self, dataset, batch_size=None, **kwargs):
+        if not set(self.schema.column_names).issubset(set(dataset.schema.column_names)):
+            raise ValueError(
+                f"Model schema {self.schema.column_names} does not match dataset schema"
+                + f" {dataset.column_names}"
+            )
+
+        # Check if merlin-dataset is passed
+        if hasattr(dataset, "to_ddf"):
+            dataset = dataset.to_ddf()
```
benfred, line 2402: I'm wondering if we should enforce that the `dataset` param is a merlin.io.Dataset class. It seems like if it isn't, the code on line 2394 will fail (since a dask ddf won't have a .schema member on it).

benfred (commented):
I think this mostly looks good! Left some minor comments iniline

merlin_models/tf/__init__.py
```diff
@@ -100,6 +103,11 @@
 )
 from .utils import repr_utils
 
+# configure_tensorflow()
+
+
+Tags.__hash__ = lambda self: hash(str(self))
```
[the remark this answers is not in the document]
marcromeyn: It’s needed for merlin-ops. Might be better to make this change in merlin-core though

merlin_models/tf/core.py
marcromeyn: Yeah you’re right! I added an extra check for this. I just want to avoid checking with `isinstance` since we could get a `nvt.Dataset` as well.

merlin_models/tf/core.py
```diff
@@ -2161,6 +2188,30 @@ def get_config(self):
         return config
 
 
+@tf.keras.utils.register_keras_serializable(package="merlin_models")
+class ModelBlock(Block, tf.keras.Model):
```
gabrielspmoreira, line 2192: Why is the `ModelBlock` needed? Just to make a `Block` serializable?

gabrielspmoreira (approved):
This is great Marc. The final API for `model.batch_predict()` is very clean. 
I just suggest adding the docstrings to the new classes introduced in `merlin_models/tf/prediction/batch.py`

merlin_models/tf/core.py
marcromeyn: Yeah it’s so that we can save each tower to disk.

PR #164, opened 2022-02-02
Introducing TupleAggregator & use it for CosineSimilarity + ElementWiseMultiply

marcromeyn:
This was a left TODO from the PR of @bschifferer.


gabrielspmoreira (approved):
The PR sounds good to me. 
I think the only missing things are docstrings for `CosineSimilarity` and `ElementWiseMultiply` and a test for `CosineSimilarity` similar to `test_elementwisemultiply`

PR #180, opened 2022-02-11
fix references to item-id column

### LINKED ISSUE #178: [BUG] Models shouldn't assume that the column tagged with `Tag.ITEM_ID` is also called "item_id"
**Describe the bug**
When attempting to train a `YoutubeDNNRetrieval` model on a dataset that has a column called "movie_ids" that is tagged with `Tag.ITEM_ID`, `SequentialBlock.build()` fails with the error:
```
tests/unit/inference/test_poc.py:254: in create_retrieval_model
    history = model.fit(train_dataset_tf, callbacks=[tensorboard_callback], epochs=1)
/usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py:67: in error_handler
    raise e.with_traceback(filtered_tb) from None
/usr/local/lib/python3.8/dist-packages/merlin_models/tf/core.py:2050: in call
    outputs = self.body(inputs, **kwargs)
/usr/local/lib/python3.8/dist-packages/merlin_models/config/schema.py:50: in __call__
    return super().__call__(*args, **kwargs)
/usr/local/lib/python3.8/dist-packages/merlin_models/tf/core.py:210: in _maybe_build
    super()._maybe_build(inputs)
/usr/local/lib/python3.8/dist-packages/merlin_models/tf/core.py:470: in build
    self._maybe_propagate_context(input_shape)
/usr/local/lib/python3.8/dist-packages/merlin_models/tf/core.py:388: in _maybe_propagate_context
    self.context.build(input_shapes)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = BlockContext(
  (_feature_names): List(
    (0): 'item_id'
  )
  (_feature_dtypes): Dict(
    (search_terms): tf.int32
    (movie_ids): tf.int32
    (genres): tf.int32
    (movie_id_count): tf.int32
  )
)
input_shape = {'genres': (TensorShape([436, 1]), TensorShape([16, 1])), 'movie_id_count': TensorShape([16, 1]), 'movie_ids': (TensorShape([212, 1]), TensorShape([16, 1])), 'search_terms': (TensorShape([212, 1]), TensorShape([16, 1]))}

    def build(self, input_shape):
        for feature_name in self._feature_names:
            if feature_name not in self.named_variables:
>               shape = input_shape[feature_name]
E               KeyError: 'Exception encountered when calling layer "model" (type Model).\n\nitem_id\n\nCall arguments received:\n  • inputs={\'search_terms\': (\'tf.Tensor(shape=(212, 1), dtype=int32)\', \'tf.Tensor(shape=(16, 1), dtype=int64)\'), \'movie_ids\': (\'tf.Tensor(shape=(212, 1), dtype=int32)\', \'tf.Tensor(shape=(16, 1), dtype=int64)\'), \'genres\': (\'tf.Tensor(shape=(436, 1), dtype=int32)\', \'tf.Tensor(shape=(16, 1), dtype=int64)\'), \'movie_id_count\': \'tf.Tensor(shape=(16, 1), dtype=int32)\'}\n  • kwargs={\'training\': \'False\'}'

/usr/local/lib/python3.8/dist-packages/merlin_models/tf/core.py:117: KeyError
```

**Steps/Code to reproduce bug**
Tag a column that isn't called "item_id" with `Tag.ITEM_ID` in the schema, and pass the adjusted schema to a model creation method like `YoutubeDNNRetrieval()`.

**Expected behavior**
Models should check the schema to find the name of the column tagged as `ITEM_ID` and use the name of that column instead of the string "item_id" for e.g. shape look-ups.

**Additional context**
Searching the repo for `Tag.ITEM_ID` turns up a few different places where it looks like the string value of `Tag.ITEM_ID` is used as a feature name, so the issue may affect more than `YoutubeDNNRetrieval`.

sararb:
This PR fixes #178 : 

- [x] Enable selecting user-id or item-id from the context with a schema
- [x] Replacing all the fixed references to the item-id column name (`str(Tag.ITEM_ID)`) to a dynamic parameter called `item_id_feature_name`: When possible, this is done implicitly by calling the schema-object set during the build method. 
- [x] Update sequence testing synthetic dataset by changing the  item-id column name to `item_id_seq` and user_id to `test_user_id`


marcromeyn:
I don’t think this is the right way to solve the issue, since we would need too many specialized methods in the schema-object. I prototyped another solution [here](https://github.com/NVIDIA-Merlin/models/tree/select-by-tag), this is using the schema-object that’s part of each `Block`. We could enable to select from the context with a Schema ([source](https://github.com/NVIDIA-Merlin/models/blob/select-by-tag/merlin_models/tf/core.py#L78-L81)).

sararb:
Thanks for the suggestions! I missed the possibility of calling the schema within the `call`  of any block, which makes getting the itme-id column name much easier!  

But we'll still need an explicit call to select the item-id column from the schema when we need to access some information in the __init__() method of a block like [here](https://github.com/NVIDIA-Merlin/models/blob/2cf6cb52a02ac4f7f3dc1b40657a8138ea38d0f2/merlin_models/tf/prediction/item_prediction.py#L97):  where we retrieve the parameter `num_classes` related to items catalog size, right?

So, we could maybe set the explicit call to `select_by_tag` in these types of blocks if we don't want to add specialized methods to the schema class.

marcromeyn (approved):

karlhigley:
Thanks for fixing this! 🙏🏻

PR #184, opened 2022-02-15
Unified Schema

benfred:
This modifies merlin-models to be compatabile with the new 'core' schema class that
will be shared with NVTabular


merlin_models/loader/backend.py
```diff
@@ -27,6 +27,8 @@
 except ImportError:
     cp = np
 
+from merlin.graph.tags import Tags
```
marcromeyn, line 30: I would prefer it if this would be moved to either the top-level `from merlin import Tags` or to core: `from merlin.core import Tags`

merlin_models/tf/utils/tf_utils.py
```diff
@@ -22,9 +22,9 @@
 
 def get_output_sizes_from_schema(schema, batch_size=0, max_sequence_length=None):
     sizes = {}
-    for feature in schema.feature:
+    for feature in schema:
         name = feature.name
-        if feature.HasField("value_count"):
+        if feature._is_list:
```
marcromeyn, line 27: Shouldn’t this be a public property?

merlin_models/loader/backend.py
benfred: Agreed - I was thinking we should go `from merlin.schema import Schema, Tags`

merlin_models/tf/utils/tf_utils.py
benfred: Yeah - I think it probably should be .  I just left as a protected property so I wouldn't have to make changes to NVT as well. I'll update merlin-core and make the `_is_list` version deprecated

karlhigley: Yeah, we've been thinking about changing this on the NVT side too. `_is_ragged` should probably also become public.

merlin_models/loader/backend.py
karlhigley: I'm good with any of those. We've dragged our feet on shortening imports on the NVT side, but anything we use in multiple libraries should be quick and clean to import 👍🏻

benfred: I've moved here https://github.com/NVIDIA-Merlin/core/pull/11 (also made is_list and is_ragged public)

benfred: made that update -

merlin_models/tf/utils/tf_utils.py
benfred: done

tests/tf/test_core.py
```diff
@@ -72,31 +72,31 @@ def test_serialization_continuous_features(
...
-        return self.context.get_embedding(Tag.ITEM_ID)
+        return self.context.get_embedding(Tags.ITEM_ID)
 
 
 def test_block_context(ecommerce_data: SyntheticData):
     inputs = ml.InputBlock(ecommerce_data.schema)
     dummy = DummyFeaturesBlock()
     model = inputs.connect(ml.MLPBlock([64]), dummy, context=ml.BlockContext())
-
+    print(ecommerce_data.tf_tensor_dict.keys())
```
benfred, line 93:
```suggestion
```

marcromeyn (approved):

merlin_models/data/synthetic.py
```diff
@@ -99,9 +104,9 @@ def read_schema(cls, path: Union[str, Path]) -> Schema:
             os.path.join(str(path), "schema.json") if os.path.isdir(str(path)) else str(path)
         )
         if _schema_path.endswith(".pb") or _schema_path.endswith(".pbtxt"):
-            return Schema().from_proto_text(_schema_path)
+            TensorflowMetadata.from_from_proto_text(_schema_path).to_merlin_schema()
```
karlhigley, line 107: I think this should be `from_proto_text_file`? Also not sure this works with the way that method is written, since I think the current implementation wants the directory and assumes the file is called `schema.pbtxt` unless you pass a different value of the `filename` param. Now wondering if this is covered by tests too...

merlin_models/tf/block/retrieval.py
```diff
@@ -100,14 +100,12 @@ def TwoTowerBlock(
...
-    item_id_tag=Tag.ITEM_ID,
+    query_id_tag=Tags.USER_ID,
+    item_id_tag=Tags.ITEM_ID,
     embeddings_initializers: Optional[Dict[str, Callable[[Any], None]]] = None,
     **kwargs,
 ):
-    query_item_schema = schema.select_by_tag(
-        lambda tags: query_id_tag in tags or item_id_tag in tags
-    )
+    query_item_schema = schema.select_by_tag(query_id_tag) + schema.select_by_tag(item_id_tag)
```
karlhigley, line 108: This seems like a reasonable way to implement the same behavior on the other version of the `Schema` API. I do wonder if we should think about integrating the `lambda` functionality into core though.

merlin_models/tf/core.py
```diff
@@ -87,10 +92,18 @@ def __getitem__(self, item):
...
+        else:
+            item = str(item)
+        return self.named_variables[item]
 
     def get_embedding(self, item):
-        return self.named_variables[f"{str(item)}/embedding"]
+        if isinstance(item, Tags):
+            item = item.value
+        else:
+            item = str(item)
```
karlhigley, line 105: I wonder if this could be a helper on `Tags` to keep the behavior encapsulated. Something like:
```
class Tags(Enum):
    ...
    @classmethod
    def value(cls, maybe_tag):
        if isinstance(maybe_tag, Tags):
            return item.value
        else:
            return str(item)
```

karlhigley (commented):
Looks pretty good to me. Left a few comments on things we might want to abstract further, and one spot where I think there's a method name typo that didn't get caught by the tests.

PR #185, opened 2022-02-15
Adding some basic doc-strings

marcromeyn:


merlin_models/tf/block/inputs.py
```diff
@@ -52,52 +52,67 @@ def InputBlock(
     seq_aggregator: Block = SequenceAggregator(SequenceAggregation.MEAN),
     **kwargs,
 ) -> Block:
-    """Input Block to process categorical and continuous input features.
+    """The entry block of the model to process input features from a schema.
+
+    This function creates continuous and embedding layers, and connects them via `ParallelBlock`.
```
sararb, line 57: Maybe we could add a short description of the sequential case (seq=True)? In this case, the features are split into context and sparse groups, then two InputBlocks are generated, and finally, these blocks are connected via a `ParallelBlock`.

merlin_models/tf/core.py
```diff
@@ -299,6 +366,22 @@ def connect_with_shortcut(
         aggregation: Optional["TabularAggregationType"] = None,
         block_outputs_name: Optional[str] = None,
     ) -> "SequentialBlock":
+        """Connect the block to other blocks sequentially with a shortcut connection.
```
sararb, line 369: Maybe we could add a brief definition of shortcut connection?

sararb (approved):

PR #187, opened 2022-02-16
Adding models

marcromeyn:
Fixes # (issue)


merlin_models/tf/blocks/retrieval.py
```diff
@@ -111,8 +112,16 @@ def MatrixFactorizationBlock(
     embedding_options = EmbeddingOptions(
         embedding_dim_default=dim, embeddings_initializers=embeddings_initializers
     )
+
+    rename_features = RenameFeatures({str(query_id_tag): "query", str(item_id_tag): "item"})
```
sararb, line 116: We need to retrieve the name of user_id and item_id features from the schema instead of using the string representation of their tags.

sararb (approved):

merlin_models/tf/blocks/retrieval.py
marcromeyn: Good catch! Fixed now

PR #195, opened 2022-02-22
Dataloader/targets

marcromeyn:
Fixes # (issue)
- #211 
### Goals :soccer:
Standardize our APIs to make `merlin.io.Dataset` the main abstraction.
We now under the hood create the data-loader. The blocks will automatically convert list columns to sparse/dense..

### Testing Details :mag:
All our tests are now making use of `merlin.io.Dataset` to train/evaluate.


merlin/models/tf/blocks/core/transformations.py
```diff
@@ -67,11 +67,11 @@ def call(self, inputs: TabularData, **kwargs) -> TabularData:
                 row_lengths = val[1][:, 0]
                 ragged = tf.RaggedTensor.from_row_lengths(values, row_lengths)
                 if self.max_seq_length:
-                    outputs[name] = ragged.to_tensor(shape=[None, self.max_seq_length])
+                    outputs[name] = tf.squeeze(ragged.to_tensor(shape=[None, self.max_seq_length]))
```
marcromeyn: Do we have to check the num dims herte to know whether to squeeze or not?

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -37,6 +37,12 @@ def call(self, inputs: TabularData, **kwargs) -> TabularData:
             if isinstance(val, tuple):
                 values = val[0][:, 0]
                 row_lengths = val[1][:, 0]
+
+                if values.dtype == tf.float32:
```
gabrielspmoreira, line 41: Are we always converting values from float to int? Couldn't the sparse features be floats?

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -37,6 +37,12 @@ def call(self, inputs: TabularData, **kwargs) -> TabularData:
...
+                    values = tf.cast(values, tf.int32)
+                if row_lengths.dtype == tf.float32:
```
gabrielspmoreira, line 43: I am not sure if when mixed precision is enabled (like below) that would after the dtypes here to tf.float16.
In any case, it would be safer to check if the number is a float by using `row_lengths.dtype.is_floating`

```
policy = mixed_precision.Policy('mixed_float16')
mixed_precision.set_global_policy(policy)
```

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -50,20 +56,44 @@ def compute_output_shape(self, input_shape):
...
         for name, val in inputs.items():
             if isinstance(val, tuple):
                 values = val[0][:, 0]
                 row_lengths = val[1][:, 0]
-                outputs[name] = tf.RaggedTensor.from_row_lengths(values, row_lengths).to_tensor()
+                ragged = tf.RaggedTensor.from_row_lengths(values, row_lengths)
+                if self.max_seq_length:
+                    outputs[name] = ragged.to_tensor(shape=[None, self.max_seq_length])
+                else:
+                    outputs[name] = tf.squeeze(ragged.to_tensor())
```
gabrielspmoreira, line 73: Should we raise an exception if `max_seq_length` is not defined for a sparse feature?

tests/tf/test_core.py
```diff
@@ -78,9 +78,14 @@ def call(self, inputs, **kwargs):
         items = self.context[Tags.ITEM_ID]
         emb_table = self.context.get_embedding(Tags.ITEM_ID)
         item_embeddings = tf.gather(emb_table, tf.cast(items, tf.int32))
+        if len(item_embeddings.shape) == 3:
```
gabrielspmoreira, line 81: `tf.rank(item_embeddings) == 3` ?

gabrielspmoreira (approved):
This PR sounds good to me. Just added a few comments.

What do we gain by setting the batch_size in the `model.fit()` method instead of having it defined in the data loader? Does that mean that TF [data_adapter](https://github.com/keras-team/keras/blob/v2.8.0/keras/engine/data_adapter.py), will now be responsible for shuffling and iterating over the batches?

marcromeyn:
@gabrielspmoreira The big advantage is that we standardize our APIs. Our current API is confusing because `.fit` & `.evaluate` take in the data-loader but batch-predict & user/item-embeddings a dataset (since it’s using Dask). This PR will standardize it accross the board.

Another advantage is that we remove boilerplate. Since we control the entire model we can infer which columns should be sparse/dense etc. Therefore the user doesn’t have to specify it manually anymore.

merlin/models/tf/blocks/core/transformations.py
marcromeyn: I think it’s fine without since we can still convert it, then the max-lenght within the batch will be used as the max which is OK for certain scenarios I think. Does that make sense?

merlin/models/tf/blocks/core/transformations.py
gabrielspmoreira: Yeah. If batch size or `self.max_seq_length` are 1 we might have an issue with tf.squeeze here

merlin/models/tf/blocks/core/transformations.py
gabrielspmoreira: Ok. It makes sense. And the users can always reshape the tensors on their side too

PR #199, opened 2022-02-24
Pin `merlin-core` version to specific commit to avoid breaking changes

karlhigley:
There are some upcoming changes to the package names in `merlin-core` (e.g. `graph->dag`) that would break the tests in `merlin-models` if we continued depending directly on the latest commit in `merlin-core`. Pinning the version to a specific commit allows us to bump the version here in the same PR that propagates upstream changes from `merlin-core` without breaking the tests/CI on everyone else's PRs.

Fixes # (issue)

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -1,5 +1,4 @@
 numpy>=1.17.0
 tqdm>=4.27
 pyarrow>=1.0
-merlin-core@git+https://github.com/NVIDIA-Merlin/core.git
-
+merlin-core@git+https://github.com/NVIDIA-Merlin/core.git@2c980f7f949e378f6fc8a8bee4c3a7dc9d021020


benfred:
Are there other changes aside from the merlin.graph -> merlin.dag coming in merlin-core that will cause breaks here?

merlin models doesn't depend on merlin.graph at all, so I'm wondering if we can just leave pinned to the latest main

karlhigley:
I was thinking of NVIDIA-Merlin/core#4 and NVIDIA-Merlin/core#5.  We can deal with those changes when they happen, I guess?

marcromeyn:
@benfred @karlhigley Since we’re cutting a release this week, should we hold off on this and pin merlin-core to the right version (through pip)?

karlhigley:
Yeah, the delay getting it merged has made this PR pointless, so closing it

Issue #173, opened 2022-02-08
[BUG]  getting AssertionError from  ItemRetrievalScorer() class 


rnyak:
I am getting the following error from training a mf-retrieval model. The issue is that ItemRetrieval scorer requires `query` and `item` tags from the schema, whereas they dont exist in the schema.pbtxt. This can be solved either
- user will define `user_id` as `query` and `item_id` as `item` features during the ETL process.
-  or needs to be fixed in the ItemRetrievalScorer(Block) class.

```
import merlin_models.tf as mm
mf = mm.MatrixFactorizationBlock(schema, dim=128)
model = mf.connect(mm.ItemRetrievalTask(samplers = [mm.InBatchSampler()]))
model.compile(optimizer="adam", run_eagerly=False)
train_loader = get_dataloader(nvt.Dataset(train_paths, part_size="256MB"), shuffle=True) 
losses = model.fit(train_loader, epochs=3)
model.reset_metrics()
```
```
2022-02-08 19:59:03.494259: W tensorflow/python/util/util.cc:368] Sets are not currently considered sequences, but this may change in the future, so consider avoiding using them.
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
Input In [10], in <module>
     10 print('*'*20 + '\n')
     11 train_loader = get_dataloader(nvt.Dataset(train_paths, part_size="256MB"), shuffle=True) 
---> 12 losses = model.fit(train_loader, epochs=3)
     13 model.reset_metrics()

File /usr/local/lib/python3.8/dist-packages/keras/engine/training.py:1134, in Model.fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq, max_queue_size, workers, use_multiprocessing)
   1128   self._cluster_coordinator = tf.distribute.experimental.coordinator.ClusterCoordinator(
   1129       self.distribute_strategy)
   1131 with self.distribute_strategy.scope(), \
   1132      training_utils.RespectCompiledTrainableState(self):
   1133   # Creates a `tf.data.Dataset` and handles batch and epoch iteration.
-> 1134   data_handler = data_adapter.get_data_handler(
   1135       x=x,
   1136       y=y,
   1137       sample_weight=sample_weight,
   1138       batch_size=batch_size,
   1139       steps_per_epoch=steps_per_epoch,
[...]

PR #200, opened 2022-02-24
Fix ItemRetrieval assertion and output errors

rnyak:
- This fix the error reported here https://github.com/NVIDIA-Merlin/models/issues/173

- we also find another issues in the shape of `positive_item_ids`  and fixed it.

### DIFF
--- a/merlin/models/tf/core.py
+++ b/merlin/models/tf/core.py
@@ -730,7 +730,7 @@ class SequentialBlock(Block):
     def call_targets(self, predictions, targets, training=False, **kwargs):
         outputs = targets
         for layer in self.layers:
-            if isinstance(outputs, tuple):
+            if isinstance(outputs, tuple) and len(outputs) > 0:
                 outputs, predictions = outputs
             outputs = layer.call_targets(predictions, outputs, training=training, **kwargs)
 
--- a/merlin/models/tf/core.py
+++ b/merlin/models/tf/core.py
@@ -730,7 +730,7 @@ class SequentialBlock(Block):
     def call_targets(self, predictions, targets, training=False, **kwargs):
         outputs = targets
         for layer in self.layers:
-            if isinstance(outputs, tuple) and len(outputs) > 0:
+            if isinstance(outputs, (list, tuple)) and len(outputs) > 0:
                 outputs, predictions = outputs
             outputs = layer.call_targets(predictions, outputs, training=training, **kwargs)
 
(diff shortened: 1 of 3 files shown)


merlin/models/tf/core.py
```diff
@@ -730,7 +730,7 @@ def compute_loss(self, inputs, targets, **kwargs):
     def call_targets(self, predictions, targets, training=False, **kwargs):
         outputs = targets
         for layer in self.layers:
-            if isinstance(outputs, tuple):
+            if isinstance(outputs, tuple) and len(outputs) > 0:
```
marcromeyn, line 733:
```suggestion
            if isinstance(outputs, (list, tuple)) and len(outputs) > 0:
```

merlin/models/tf/prediction/item_prediction.py
```diff
@@ -143,6 +143,8 @@ class ItemRetrievalScorer(Block):
     item_id_feature_name: str
         Name of the column containing the item ids
         Defaults to `item_id`
+    query_name: str = 'query'
```
marcromeyn, line 146: Could use a single line of explanation, same for `item_name`

merlin/models/tf/prediction/item_prediction.py
```diff
@@ -179,7 +186,7 @@ def add_features_to_context(self, feature_shapes) -> List[str]:
         return self._required_features
 
     def _check_input_from_two_tower(self, inputs):
-        assert set(inputs.keys()) == {"query", "item"}
+        assert set(inputs.keys()) == set([self.query_name, self.item_name])
```
marcromeyn, line 189:
```suggestion
        if set(inputs.keys()) != set([self.query_name, self.item_name]):
            raise ValueError(f“Wrong input-names, expected: {[self.query_name, self.item_name]} but got: {inputs.keys()}")
```

marcromeyn (changes requested):
Looks good, I left some small comments

merlin/models/tf/prediction/item_prediction.py
rnyak: yes for sure.

merlin/models/tf/core.py
rnyak: thanks.

gabrielspmoreira (approved):
Looks good to me. 
Have included here fixes to deal with the zero loss (bug in the ItemRetrievalScorer) and nan loss when using mixed precision (fp16) due to min_float being smaller than the min float32 value.

marcromeyn (approved):

PR #194, opened 2022-02-21
MLP Model

bschifferer:
Provide MLPModel as abstraction layer

PR #205, opened 2022-03-01
MLP Model

rnyak:
This is Benedikt's MLP model. Original PR is here https://github.com/NVIDIA-Merlin/models/pull/194, recreating it due to CI error.


merlin/models/tf/models/ranking.py
```diff
@@ -7,9 +7,58 @@
 from ..blocks.inputs import InputBlock
 from ..blocks.mlp import MLPBlock
 from ..core import Block, Model, ParallelPredictionBlock, PredictionTask
+from ..features.embedding import EmbeddingOptions
 from .utils import _parse_prediction_tasks
 
 
+def MLPModel(
+    schema: Schema,
+    embedding_dim: int,
```
marcromeyn, line 16: What’s the rational behind only exposing `embedding_dim` and not the rest of the options of `EmbeddingOptions `?

marcromeyn:
I am worried that this will lead to an explosion of these type of model functions for each block we create. To circumvent this, my proposal is outlined in the following PR: 
- #209

merlin/models/tf/models/ranking.py
marcromeyn: Ah so to be clear, the proposal I posted in the comment below was meant to be a replacement this of the PR. So if we agree with the proposal, we can close this PR.

rnyak: We can close ok.

PR #207, opened 2022-03-01
Add a helper function to get the movielens dataset

benfred:
We already had a helper function to download and do ETL on the movielens dataset,
but this adds a higher level function where we can also just get the downloaded
movielens dataset, and download it if it doesn't exist.

This will allow us to go:
```python
from merlin.models.utils.data_etl_utils import get_movielens

train, valid = get_movielens(variant="ml-25m")
```

Which will return two merlin.io.Dataset objects for the training and validation datasets.
The schema for these datasets can be gotten from the schema member (so like w/
```train.schema```).

This should avoid some complexity in the first tutorial notebooks, and let our
notebooks get to the training step right away

### CODE REVIEW

File: merlin/models/utils/data_etl_utils.py [hunk 1]
```diff
@@ -19,6 +20,23 @@
 logger.setLevel(logging.INFO)
 
 
+def get_movielens(path=None, variant="ml-25m"):
+    """Returns a tuple of train/test Dataset objects for the movielens dataset"""
```
marcromeyn, line 24: Could use docs for the parameters (mostly to show the options for variant). I’m wondering if it would be better to rename `variant` to `size` and allow for `{100k, 25m}`

benfred: I've added a better docstring in the last commit - 

I left the parameter name as variant, size there are dataset differences beyond just the number of rows (like the ml-100k dataset has extra user metadata that the ml-25m dataset doesn't have). I don't have a strong opinion on this though

PR #209, opened 2022-03-01
Adding wrap_as_model to Block

marcromeyn:
### Goals :soccer:
We recently introduced high-level functions to create models to make it easier to create models. This PR tries to introduce new functionality to each block to be converted to a model. This because otherwise we would need a high-level model-function for each block we create (like for instance #205 does for `MLPBlock`).

This PR adds `wrap_as_model` to a block, which allows you to wrap a block in-between a `InputBlock` and prediction-tasks. This can be used for instance to convert a `MLPBlock` to a model by doing:

```python
model = ml.MLPBlock([512, 256]).wrap_as_model(schema)
```

Another potential solution would be to add `from_block` to the `Model` class, to API would then look as follows:

```python
model = ml.Model.from_block(ml.MLPBlock([512, 256]), schema)
```

### Testing Details :mag:
Added a simple test to ensure wrap_as_model works as expected.

### DIFF
--- a/merlin/models/tf/core.py
+++ b/merlin/models/tf/core.py
@@ -382,7 +382,7 @@ class Block(SchemaMixin, ContextMixin, Layer):
 
         return output
 
-    def wrap_as_model(
+    def to_model(
         self,
         schema: Schema,
         input_block: Optional[BlockType] = None,
@@ -407,6 +407,11 @@ class Block(SchemaMixin, ContextMixin, Layer):
         """
         from merlin.models.tf.blocks.inputs import InputBlock
         from merlin.models.tf.models.utils import parse_prediction_tasks
+
+        if is_input_block(self.first):
+            if input_block is not None:
+                raise ValueError("The block already includes an InputBlock")
+            input_block = self.first
 
         aggregation = kwargs.pop("aggregation", "concat")
         input_block = input_block or InputBlock(schema, aggregation=aggregation, **kwargs)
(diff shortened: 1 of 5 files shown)


gabrielspmoreira:
@marcromeyn  Both API solutions seems reasonable and useful for me!
Could not we have both (`block.to_model()` and `model.from_block()`) ?

rnyak:
> @marcromeyn Both API solutions seems reasonable and useful for me! Could not we have both (`block.to_model()` and `model.from_block()`) ?

@marcromeyn I liked Gabriel's suggestion. I think `to_model()` sound better than `wrap_as_model()`.

marcromeyn:
I think `to_model` makes sense. We do need to add some extra checks there though, since some blocks (like `DLRMBlock`) already contains the input-block. So then we shouldn’t add it again, and if the user provides it we need to throw an error.

merlin/models/tf/core.py
```diff
@@ -374,6 +374,44 @@ def connect(
...
+            Schema to use for the model.
+        input_block: Optional[Block]
+            Block to use as input.
+        prediction_tasks: Optional[
+            Union[PredictionTask, List[PredictionTask], ParallelPredictionBlock]
+        ]
+            Prediction tasks to use.
+
+        """
+        from merlin.models.tf.blocks.inputs import InputBlock
```
rnyak, line 400: How do we set `embedding_dim` in `to_model()`?

tests/tf/test_core.py
```diff
@@ -141,6 +138,42 @@ def test_simple_model(ecommerce_data: SyntheticData):
     )
 
 
+def test_block_to_model(ecommerce_data: SyntheticData):
+    model = ml.MLPBlock([64]).to_model(
```
rnyak, line 142: Similar comment: Better to showcase how a user can define embedding_dim here.

tests/tf/test_core.py
```diff
@@ -141,6 +138,42 @@ def test_simple_model(ecommerce_data: SyntheticData):
...
+
+    copy_model = testing_utils.assert_serialization(model)
+    testing_utils.assert_loss_and_metrics_are_valid(
+        copy_model, ecommerce_data.tf_features_and_targets
+    )
+
+
+def test_model_from_block(ecommerce_data: SyntheticData):
+    model = ml.Model.from_block(
+        ml.MLPBlock([64]),
```
rnyak, line 154: same comment with the above. Show how embedding_dim can be set.

tests/tf/test_core.py
marcromeyn: Added it [here](https://github.com/NVIDIA-Merlin/models/blob/tf/wrap-as-model/tests/tf/test_core.py#L142-L147), works the same for `Model.from_block`

PR #219, opened 2022-03-02
Add shim to adapt the lightfm library

benfred:
Similar to #218, this change adds an adaptor to adapt lightfm
to work with the high level merlin models api


benfred:
> Could be good to add `requirements/lightfm.txt` & add it to `requires_extra` in the setup.py.

done -

EvenOldridge (approved):
Looks good as a start.  We'll probably want to flesh out the testing in more detail for future releases but this looks fine to me for now.

merlin/models/lightfm/__init__.py
```diff
@@ -0,0 +1,103 @@
...
+        print(model.evaluate(valid))
+    """
+
+    def __init__(self, *args, epochs=10, num_threads=0, **kwargs):
+        self.lightfm_model = lightfm.LightFM(*args, **kwargs)
+        self.epochs = epochs
+        self.num_threads = num_threads or multiprocessing.cpu_count()
+
+    def fit(self, train: Dataset):
+        """Trains the implicit model
```
benfred, line 52:
```suggestion
        """Trains the lightfm model
```

PR #222, opened 2022-03-04
fix for tf when creating cpu dataloader

jperez999:
This fix updates the dataloader to use the dataset again by default, even when creating from a series of paths and not an nvt dataset. It also ensures that the cpu toggle forces complete logic off gpu for dataloader.

### DIFF
--- a/merlin/models/tf/dataset.py
+++ b/merlin/models/tf/dataset.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 import contextlib
+import logging
 import os
 
 import dask.dataframe as dd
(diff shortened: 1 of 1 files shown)


merlin/models/tf/dataset.py
```diff
@@ -76,6 +82,16 @@ def _validate_dataset(paths_or_dataset, batch_size, buffer_size, engine, reader_
...
         # default engine is parquet
         engine = "parquet"
+
+    cpu = device and "cpu" in device
+
+    if nvt_dataset_class:
+        return nvt_dataset_class(files, engine=engine, cpu=cpu)
+    else:
+        LOG.warning(
+            "NVTabular Dataset class not detected, reverting to Dask Dataframe."
```
karlhigley, line 92:
```suggestion
            "Merlin Dataset class not detected, reverting to Dask Dataframe."
```

karlhigley (approved):

marcromeyn (approved):

PR #226, opened 2022-03-08
Refactor and optimize model evaluation 

### LINKED ISSUE #202: [FEA] Optimize `torch_gather_like` operator used in top-k extraction
# 🚀 Feature request

## Motivation
- Create a vectorized version of  torch_gather_like function implemented [here](https://github.com/NVIDIA-Merlin/models/blob/c0df6db16180d696400e7fab9747e907eb394cf4/merlin/models/tf/utils/tf_utils.py#L108). This function is used to gather top-k indices from each row of one-hot vector of labels. 

## Your contribution

Maybe later check if we could come up with a more efficient implementation of `gather_torch_like` which does not using `TensorArray`, but rather uses only `tf.gather_nd()` with a 2D index (rows, cols), like the one I implemented here [here](https://github.com/NVIDIA-Merlin/NVTabular/blob/dc9c04d126b8ad842a32a3618e8f4495afbe0678/nvtabular/loader/tensorflow.py#L404). 
Not a priority for now, but maybe it might improve the metrics performance

_Originally posted by @gabrielspmoreira in https://github.com/NVIDIA-Merlin/models/pull/189#discussion_r813052963_

### LINKED ISSUE #210: [FEA] Computing Training/Eval metrics after each n steps
# 🚀 Feature request

Create a custom Callback for computing metrics after each n-steps instead of computing them for each batch.  This callback can be then provided to the `fit` method.

## Motivation

- Evaluating the RetrievalModels trained with negative sampling strategies requires comparing the positive scores against a set of implicit negative ones using ranking metrics. One required step to compute those metrics is the extraction of the top-k negatives.  This process can be time-consuming (when the catalog of items get large) and the goal of this feature is to gather positive scores and labels for n steps before calling the top-k extraction and ranking metrics calculation.

### LINKED ISSUE #225: [BUG] Pre-sorting of predictions for metrics optimization causes incorrect metric calculation
Refactoring metrics pre-sorting computation to avoid incorrect calculation because pre-sorting and extracting top-k changes the number of relevant count for each row, which is not correct.

gabrielspmoreira:
This PR refactors model evaluation and also optimizes eval performance:

- [x] Renaming `pre_metrics` (e.g. BruteForceTopK) to `pre_eval_topk` (for both `ItemRetrievalTask` and `NextItemPredictionTask`), to make it clear it is only called during evaluation/prediction, and not when evaluating during training
- [x] Converting `tf_utils.gather_torch_like()` method into a vectorized version to retrieve the top-k labels. Closes #202 
- [x] Created the LabelToOneHot transformation block to centralized that logic, which was done in many places. Now it is expected that when ranking metrics are used the labels must be necessarily one-hot encoded, as that is required by `extract_topk`.
- [x] Encapsulated internal properties with setter methods for better maintainability (e.g. `self.loss_block.retrieval_scorer.cache_query` and self.loss_block.pre_eval_topk)
- [x] Refactored metrics pre-sorting computation to avoid incorrect calculation (because pre-sorting and extracting top-k changes the number of relevant count for each row, which is not correct). Closes #225 
- [x] Refactored ranking metrics to have a single instance for each k (rather one metric with many `top_ks`), which simplifies naming and collecting results 
- [ ] Refactored ranking metrics to accumulate only the sum of batch metrics and the batches count, to aggregate results the same way default Keras metrics does (simple average over the epoch by dividing the sum by the count) **(in progress)**
- [ ] Making train metrics to be computed each `train_metrics_steps`. If `train_metrics_steps=0` no metrics are computed during training. Closes #210  **(In progress)**

### DIFF
--- a/merlin/models/tf/blocks/core/transformations.py
+++ b/merlin/models/tf/blocks/core/transformations.py
@@ -25,6 +25,7 @@ from merlin.schema import Schema, Tags
 from ....utils.schema import categorical_cardinalities
 from ...core import Block, PredictionOutput, TabularBlock
 from ...typing import TabularData, TensorOrTabularData
+from ...utils.tf_utils import transform_label_to_onehot
 
 
 @Block.registry.register("as-sparse")
@@ -400,3 +401,23 @@ class ItemsPredictionWeightTying(Block):
         logits = tf.nn.bias_add(logits, self.bias)
 
         return logits
+
+
+@Block.registry.register_with_multiple_names("label_to_onehot")
+@tf.keras.utils.register_keras_serializable(package="merlin_models")
+class LabelToOneHot(Block):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def compute_output_shape(self, input_shape):
+        return input_shape
+
+    def call_outputs(
+        self, outputs: PredictionOutput, training=True, **kwargs
+    ) -> "PredictionOutput":
+        targets, predictions = outputs.targets, outputs.predictions
+
+        num_classes = tf.shape(predictions)[-1]
+        targets = transform_label_to_onehot(targets, num_classes)
+
+        return PredictionOutput(predictions, targets)
--- a/merlin/models/tf/core.py
+++ b/merlin/models/tf/core.py
@@ -17,6 +17,7 @@
 import abc
 import copy
 import sys
+from collections import Sequence as SequenceCollection
 from collections import defaultdict
 from dataclasses import dataclass
 from functools import reduce
@@ -50,6 +51,7 @@ from merlin.models.utils.schema import (
 )
 from merlin.schema import Schema, Tags
 
+from .metrics.ranking import RankingMetric2
 from .typing import TabularData, TensorOrTabularData
 from .utils.mixins import LossMixin, MetricsMixin, ModelLikeBlock
 from .utils.tf_utils import (
@@ -721,7 +723,11 @@ class SequentialBlock(Block):
             if i == len(self.layers) - 1:
                 filtered_kwargs = filter_kwargs(kwargs, layer, filter_positional_or_keyword=False)
                 filtered_kwargs.update(
-                    filter_kwargs(kwargs, layer.call, filter_positional_or_keyword=False)
+                    filter_kwargs(
+                        {**dict(training=training), **kwargs},
+                        layer.call,
+                        filter_positional_or_keyword=False,
+                    )
                 )
             else:
                 filtered_kwargs = filter_kwargs(
@@ -1736,12 +1742,11 @@ class PredictionTask(Layer, LossMixin, MetricsMixin, ContextMixin):
         task_name: Optional[str] = None,
         metrics: Optional[List[MetricOrMetricClass]] = None,
         pre: Optional[Block] = None,
-        pre_metrics: Optional[Block] = None,
+        pre_eval_topk: Optional[Block] = None,
         task_block: Optional[Layer] = None,
         prediction_metrics: Optional[List[tf.keras.metrics.Metric]] = None,
         label_metrics: Optional[List[tf.keras.metrics.Metric]] = None,
         loss_metrics: Optional[List[tf.keras.metrics.Metric]] = None,
-        compute_train_metrics: Optional[bool] = True,
         name: Optional[Text] = None,
         **kwargs,
     ) -> None:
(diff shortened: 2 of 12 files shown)


merlin/models/tf/blocks/core/transformations.py
```diff
@@ -400,3 +401,23 @@ def call(self, inputs, training=False, **kwargs) -> tf.Tensor:
...
+
+
+@Block.registry.register_with_multiple_names("label_to_onehot")
+@tf.keras.utils.register_keras_serializable(package="merlin_models")
+class LabelToOneHot(Block):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def compute_output_shape(self, input_shape):
+        return input_shape
```
marcromeyn, line 413: Doesn’t the shape change here?

merlin/models/tf/core.py
```diff
@@ -1788,14 +1795,22 @@ def __init__(
...
         self._task_name = task_name
         self.pre = pre
-        self.pre_metrics = pre_metrics
+        self.pre_eval_topk = pre_eval_topk
+
+        if metrics is not None:
+            if isinstance(metrics, SequenceCollection):
+                metrics = list(metrics)
+            else:
+                metrics = [metrics]
```
marcromeyn, line 1804: Seems like this allows you to pass in a single metric, if that’s the case we should update the type-hint in the class.

merlin/models/tf/core.py
```diff
@@ -1788,14 +1795,22 @@ def __init__(
...
 
         create_metrics = self._create_metrics
         self.eval_metrics = create_metrics(metrics) if metrics else []
         self.prediction_metrics = create_metrics(prediction_metrics) if prediction_metrics else []
         self.label_metrics = create_metrics(label_metrics) if label_metrics else []
         self.loss_metrics = create_metrics(loss_metrics) if loss_metrics else []
-        self.compute_train_metrics = compute_train_metrics
+
+    def set_pre_eval_topk(self, block):
```
marcromeyn, line 1812: Why do we want a setter for this? If it’s for safety we might want to make the variable private and expose it as a property.

merlin/models/tf/core.py
```diff
@@ -2274,6 +2367,33 @@ def get_config(self):
         return {"block": tf.keras.utils.serialize_keras_object(self.block)}
 
 
+class MetricsComputeCallback(tf.keras.callbacks.Callback):
+    def __init__(self, train_metrics_steps=1, **kwargs):
```
marcromeyn, line 2371: Maybe we should log how often we update the metrics here.

merlin/models/tf/core.py
```diff
@@ -2491,6 +2627,12 @@ def fit(
 
             x = Dataset(x, batch_size=batch_size, **kwargs)
 
+        if callbacks is None:
+            callbacks = []
+
+        # Adding a callback to control metrics computation
+        callbacks.append(MetricsComputeCallback(train_metrics_steps))
```
marcromeyn, line 2634: We probably should check whether `MetricsComputeCallback` is already part of the callbacks.

merlin/models/tf/core.py
```diff
@@ -2683,8 +2823,8 @@ def load_topk_evaluation(self, data: merlin.io.Dataset, k: int, **kwargs):
             self.retrieval_block.item_block(), data=data, k=k - 1, context=self.context, **kwargs
         )
         # set cache_query to True in the ItemRetrievalScorer
-        self.loss_block.retrieval_scorer.cache_query = True
-        self.loss_block.pre_metrics = topk_index
+        self.loss_block.set_retrieval_cache_query(True)
```
marcromeyn, line 2826: I think we should we check whether these functions are defined in the loss_block first. [This](https://github.com/NVIDIA-Merlin/models/blob/65e78feaac37fcc831de1fc7d043e4eb1f4751f9/merlin/models/tf/utils/mixins.py#L98) is the type, so it could not be there.

merlin/models/tf/blocks/core/transformations.py
gabrielspmoreira: I debugged it and input_shape comes with the correct dims `TensorShape(batch size, # items)`. In this transform, we only apply one-hot encoding to the `labels`, which in fact have there shape changed to be equal to `predictions`.

merlin/models/tf/core.py
gabrielspmoreira: Created a new `MetricOrMetrics = Union[Sequence[MetricOrMetricClass], MetricOrMetricClass]` that is now used in tasks metrics arguments

merlin/models/tf/core.py
gabrielspmoreira: Yes, it is to protect the internal variable. I did like you suggested

merlin/models/tf/core.py
gabrielspmoreira: Now I am checking if `self.loss_block` is defined and if it is an instance of `ItemRetrievalTask`

merlin/models/tf/core.py
gabrielspmoreira: When we fit() with keras, we can observe that metrics are only updated each N steps, keeping the last result for the other steps. That way we can verify that the metrics are really being computed each N steps.

PR #229, opened 2022-03-08
chore: add PR doc previews

mikemckiernan:
* I applied these files to my fork and it seemed to go well.  I pushed the workflow files to `main` in my own fork and then made a fake PR at <https://github.com/mikemckiernan/models/pull/3>.  The workflow seemed to run as expected.

* I still have fear that the workflow file updates will not run when I make this PR.  My suspicion is that contributing workflow files so they are run as part of the contribution has to happen with something other than a PR.  I'm happy to learn otherwise.

* At random, I picked the tensorflow build to update.  During the 08MAR Merlin CI meeting, the suggestion is that the torch tests can be merged with tensorflow into a single workflow file.

### DIFF
--- a/.github/workflows/docs-preview-pr.yaml
+++ b/.github/workflows/docs-preview-pr.yaml
@@ -1,0 +1,134 @@
+name: docs-preview-pr
+
+on:
+  workflow_run:
+    workflows: [tensorflow]
+    types: [completed]
+
+env:
+  WF_ID: ${{ github.event.workflow_run.id }}
+
+jobs:
+
+  # Always determine if GitHub Pages are configured for this repo.
+  get-gh-pages-url:
+    if: github.event.workflow_run.event == 'pull_request' && github.event.workflow_run.conclusion == 'success'
+    runs-on: ubuntu-latest
+    outputs:
+      url: ${{ steps.api-resp.outputs.html_url }}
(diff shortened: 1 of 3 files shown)


mikemckiernan:
New data point, but maybe only to me.  The `tensorflow` workflow from my PR ran, but the new workflow file, `docs-preview-pr`, did not.

I know the `tensorflow` workflow ran because the run includes the new artifacts: https://github.com/NVIDIA-Merlin/models/actions/runs/1953711188

I'm open to suggestion for how to best review and merge the proposed `docs-preview-pr` file.

marcromeyn:
This is awesome! I am wondering whether we should trigger this with a command, similar to how we have it for [blossom-ci](https://github.com/NVIDIA-Merlin/models/blob/main/.github/workflows/blossom-ci.yml#L20-L31)?

mikemckiernan:
> This is awesome! I am wondering whether we should trigger this with a command, similar to how we have it for [blossom-ci](https://github.com/NVIDIA-Merlin/models/blob/main/.github/workflows/blossom-ci.yml#L20-L31)?

Hmm.  I was scared away from the idea of adding a slash command because I feared having to deal with untrusted input in the comment.  Googling leads me to think dealing with untrusted input is a solved problem.  I'd like to put `/rebuild-doc` (or similar) on the table for v0.2.0.

PR #230, opened 2022-03-08
Adding NCFModel as benchmark

### LINKED ISSUE #221: [FEA] create NCFModel abstraction to build an NCF model easily
# 🌟 New model addition

Currently we build NCF model as follows:
```
mlp_path = mm.MatrixFactorizationBlock(schema, dim=64).connect(mm.MLPBlock([64, 32]))
mf_path = mm.MatrixFactorizationBlock(schema, dim=64, aggregation=mm.blocks.aggregation.ElementWiseMultiply())
ncf = mm.ParallelBlock({"mf": mf_path, "mlp": mlp_path}, aggregation='concat')
model = ncf.connect(mm.BinaryClassificationTask(target_column, metrics=[tf.keras.metrics.AUC()]))
```
It'd be nice to have an abstraction for building NCFModel, as we do for DLRMModel.

sararb:
Fixes #221

### DIFF
--- a/merlin/models/tf/blocks/retrieval/matrix_factorization.py
+++ b/merlin/models/tf/blocks/retrieval/matrix_factorization.py
@@ -60,7 +60,7 @@ def QueryItemIdsEmbeddingsBlock(
         post = rename_features
 
     embeddings_blocks = EmbeddingFeatures.from_schema(
-        query_item_schema, post=post, options=embedding_options, **kwargs
+        query_item_schema, post=post, embedding_options=embedding_options, **kwargs
     )
 
     return embeddings_blocks


merlin/models/tf/__init__.py
```diff
@@ -94,6 +94,7 @@
 )
 from .losses import LossType
 from .metrics.ranking import AvgPrecisionAt, NDCGAt, RecallAt, ranking_metrics
+from .models.benchmark import NCFModel
```
marcromeyn, line 97: I think we shouldn’t expose this directly in our public API. I think just importing benchmark is enough here, so users can use it by doing: `ml.benchmark.NCFModel(...)`

merlin/models/tf/models/benchmark.py
```diff
@@ -0,0 +1,73 @@
...
+def NCFModel(
+    schema: Schema,
+    embedding_dim: int,
+    mlp_block: MLPBlock,
+    prediction_tasks: Optional[
+        Union[PredictionTask, List[PredictionTask], ParallelPredictionBlock]
+    ] = None,
+    **kwargs
+) -> Model:
+    """NCF-model architecture.
```
marcromeyn, line 36: Should we maybe add a comment here where we link to the paper that showed that properly initialized MF can outperform NCF? cc @karlhigley

marcromeyn (approved):

merlin/models/tf/models/benchmark.py
```diff
@@ -0,0 +1,83 @@
...
+    prediction_tasks: optional
+        The prediction tasks to be used, by default this will be inferred from the Schema.
+
+    Returns
+    -------
+    Model
+
+    """
+
+    mlp_branch = MatrixFactorizationBlock(schema, dim=embedding_dim).connect(mlp_block)
```
gabrielspmoreira, line 74: Using the `MatrixFactorizationBlock` for the MLP tower of NCF made me confused and I had to check it is code to understand that it just extracts and embeds user and item ids. What turns that into MF is using `aggregation=ElementWiseMultiply()`, like below.
I think we could maybe rename the current `MatrixFactorizationBlock` to something that better describes what it does (e.g. `UserItemIdsEmbeddingsBlock`) and then define a new `MatrixFactorizationBlock` that just calls `UserItemIdsEmbeddingsBlock` with `aggregation=ElementWiseMultiply()`

gabrielspmoreira (approved):
Simplicity is the ultimate sophistication! :)

It is great to see how easy is now building new models on top of existing building blocks!
And how flexible the building blocks are, allowing to define different dimensions, aggregations and the final prediction task (could be regression for example if needed by the user).

Just left a suggestion.

merlin/models/tf/models/benchmark.py
sararb: you're right! I have just added `QueryItemIdsEmbeddingsBlock` that returns the parallel block of query and user embeddings. Then I exposed the `aggregation` argument in the `MatrixFactorizationBlock` and set it to `ElementWiseMultiply()` by default.

gabrielspmoreira (approved):
Perfect. Sounds good to me

PR #237, opened 2022-03-11
Improvements in two-tower model and DCN

gabrielspmoreira:
- [x] Ensuring Two-tower ouputs are L2-normalized
- [x] Added option to MLP block to have the last layer with no activation. This is recommended for the MLP block used for query and item tower. For example

```python
tower_block = mm.MLPBlock(
    [128, 64],    
    activation="relu",
    no_activation_last_layer=True,
)

model = mm.TwoTowerModel(
    schema,
    tower_block,
    loss="categorical_crossentropy",  
    samplers=[mm.InBatchSampler()]
)
```
- [x] Added `embedding_options` to `TwoTowerBlock`, `TwoTowerModel`, `DCNModel` to give more flexibility on the embeddings dimension and initialization

### DIFF
--- a/merlin/models/tf/blocks/core/inputs.py
+++ b/merlin/models/tf/blocks/core/inputs.py
@@ -190,7 +190,7 @@ def InputBlock(
             emb_kwargs["max_seq_length"] = max_seq_length
 
         branches["categorical"] = emb_cls.from_schema(  # type: ignore
-            schema, tags=categorical_tags, options=embedding_options, **emb_kwargs
+            schema, tags=categorical_tags, embedding_options=embedding_options, **emb_kwargs
         )
 
     if continuous_projection:
--- a/merlin/models/tf/blocks/core/transformations.py
+++ b/merlin/models/tf/blocks/core/transformations.py
@@ -386,14 +386,15 @@ class SamplingBiasCorrection(Block):
 
 
 @tf.keras.utils.register_keras_serializable(package="merlin_models")
-class PredictionsScaler(Block):
-    def __init__(self, scale_factor: float, **kwargs):
-        super(PredictionsScaler, self).__init__(**kwargs)
-        self.scale_factor = scale_factor
+class LogitsTemperatureScaler(Block):
+    def __init__(self, temperature: float, **kwargs):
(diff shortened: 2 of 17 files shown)


sararb (approved):
This looks good to me! 

Maybe we can just set normalize=False in the [ItemRetrievalTask](https://github.com/NVIDIA-Merlin/models/blob/bc2500a1872c6cfc4c7f2ab90d6583759c0ccb8a/merlin/models/tf/prediction_tasks/retrieval.py#L86) to avoid performing normalization twice.

gabrielspmoreira:
Thanks for your review Sara. 
I think we can't set `normalize=False` by default in ItemRetrievalTask because it can be also used with MatrixFactorizationBlock or other block where inputs are not normalized. So i think it is better to ensure by default that tensors are normalize in ItemRetrievalTask. It is a cheap op and applying L2 norm more than once to the tensor always return the same values.

sararb:
> Thanks for your review Sara. I think we can't set `normalize=False` by default in ItemRetrievalTask because it can be also used with MatrixFactorizationBlock or other block where inputs are not normalized. So i think it is better to ensure by default that tensors are normalize in ItemRetrievalTask. It is a cheap op and applying L2 norm more than once to the tensor always return the same values.

You're right! So maybe we can disable it in the TwoTowerModel ([here](https://github.com/NVIDIA-Merlin/models/blob/bc2500a1872c6cfc4c7f2ab90d6583759c0ccb8a/merlin/models/tf/models/retrieval.py#L162)) ?

gabrielspmoreira:
> > Thanks for your review Sara. I think we can't set `normalize=False` by default in ItemRetrievalTask because it can be also used with MatrixFactorizationBlock or other block where inputs are not normalized. So i think it is better to ensure by default that tensors are normalize in ItemRetrievalTask. It is a cheap op and applying L2 norm more than once to the tensor always return the same values.
> 
> You're right! So maybe we can disable it in the TwoTowerModel ([here](https://github.com/NVIDIA-Merlin/models/blob/bc2500a1872c6cfc4c7f2ab90d6583759c0ccb8a/merlin/models/tf/models/retrieval.py#L162)) ?

Makes sense Sara. Just did that

PR #238, opened 2022-03-11
Automatic setting of retrieval model top-k evaluation

### LINKED ISSUE #233: [BUG] When using model.fit(... validation_data=...) explicit calls to model.evaluate() fails after (in graph mode)
### Bug description
The the TF model is compiled in graph mode, abd  `model.evaluate()` is called by `model.fit()` when the `validation_data` is set for fit(), running validation at the end of each epoch.
But when `model.evaluate()` is explicitly called by the user after `model.fit(validation_data=...)` an error is raised.

1. If this [PR](https://github.com/NVIDIA-Merlin/models/pull/226) is not merged yet, use its branch `metrics_opt3` to reproduce the issue
2. Uncomment [these lines](https://github.com/NVIDIA-Merlin/models/blob/metrics_opt3/tests/tf/models/test_retrieval.py#L55-L56) on test_two_tower_retrieval_model_with_metrics and run it with `run_eagerly=True`
3. The explicitly call to `model.evaluate()`  in the end of the test will fail

### Expected behavior
It should compute both validation metrics during training and also when `model.evaluate()` is called

### Additional context
I believe the issue happens because when we call `model.fit()` the `ItemRetrievalTask._pre_eval_topk` is not defined yet. The `_pre_eval_topk` is only set after `model.fit()`, when the `model.model.load_topk_evaluation()` is called (like [here](https://github.com/NVIDIA-Merlin/models/blob/metrics_opt3/tests/tf/models/test_retrieval.py#L75-L78)) to use the items dataset and the trained item tower.

The issue is that when `model.fit(validation_data=...)` is called first (and internally `model.evaluation()`), as `_pre_eval_topk`  is frozen in the graph as `None` instead of being `TopKIndexBlock`, it is not included in the evaluation graph failing in the explicit `model.evaluate()`.

### Potential solution
A potential solution could be setting `_pre_eval_topk=TopKIndexBlock()` automatically for the `ItemRetrievalTask`, with dynamic variables `ids` and `values` with the first dim undefined, and than loading the data into the index after fit using something like `variable.assign()`

### LINKED ISSUE #236: [BUG] model.evaluate throws an error for retrieval model
### Bug description

When training a simple MatrixFactorizationModel on the movielens dataset, I get an error when calling 'model.evaluate' on the validation set.
`The max k for ranking metrics (Tensor(\"cond/cond/Max:0\", shape=(), dtype=int32)) is higher than the number of predictions in this batch`

Run this code snippet to train a basic model on the ml-25m dataset (takes about 3 minutes currently):
```python
# get the movielens dataset
from merlin.models.data.movielens import get_movielens
train, valid = get_movielens(variant="ml-25m")

# create a simple TF MatrixFactorizationModel on the movielens dataset, and fit it
import merlin.models.tf as mm
model = mm.MatrixFactorizationModel(train.schema, dim=64)
model.compile(optimizer="adam")
model.fit(train, batch_size=1024)

# the next line throws an error:
model.evaluate(valid, batch_size=1024)
```

Then running evaluation throws an error like:

```
2022-03-11 10:40:38.974068: W tensorflow/core/grappler/optimizers/loop_optimizer.cc:907] Skipping loop optimization for Merge node with control input: cond/then/_0/cond/cond/branch_executed/_114
  File "movielens.py", line 12, in <module>
    metrics = model.evaluate(valid, batch_size=1024)
  File "/home/ben/code/models/merlin/models/tf/models/base.py", line 401, in evaluate
    return super().evaluate(
  File "/home/ben/miniconda3/envs/nvtabular/lib/python3.8/site-packages/keras/engine/training.py", line 1501, in evaluate
    tmp_logs = self.test_function(iterator)
  File "/home/ben/miniconda3/envs/nvtabular/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 885, in __call__
    result = self._call(*args, **kwds)
  File "/home/ben/miniconda3/envs/nvtabular/lib/python3.8/site-packages/tensorflow/python/eager/def_function.py", line 956, in _call
    return self._concrete_stateful_fn._call_flat(
  File "/home/ben/miniconda3/envs/nvtabular/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 1963, in _call_flat
    return self._build_call_outputs(self._inference_function.call(
  File "/home/ben/miniconda3/envs/nvtabular/lib/python3.8/site-packages/tensorflow/python/eager/function.py", line 591, in call
    outputs = execute.execute(
  File "/home/ben/miniconda3/envs/nvtabular/lib/python3.8/site-packages/tensorflow/python/eager/execute.py", line 59, in quick_execute
    tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
tensorflow.python.framework.errors_impl.InvalidArgumentError: 2 root error(s) found.
  (0) Invalid argument:  assertion failed: [The max k for ranking metrics (Tensor(\"cond/cond/Max:0\", shape=(), dtype=int32)) is higher than the number of predictions in this batch] [Condition x >= y did not hold element-wise:] [x (cond/cond/strided_slice:0) = ] [1] [y (cond/cond/Max:0) = ] [10]
	 [[{{node cond/else/_1/cond/cond/then/_132/cond/cond/assert_greater_equal/Assert/Assert}}]]
	 [[cond/else/_1/cond/BPRLoss/assert_equal_1/Assert/AssertGuard/else/_123/cond/BPRLoss/assert_equal_1/Assert/AssertGuard/Assert/data_0/_173]]
  (1) Invalid argument:  assertion failed: [The max k for ranking metrics (Tensor(\"cond/cond/Max:0\", shape=(), dtype=int32)) is higher than the number of predictions in this batch] [Condition x >= y did not hold element-wise:] [x (cond/cond/strided_slice:0) = ] [1] [y (cond/cond/Max:0) = ] [10]
	 [[{{node cond/else/_1/cond/cond/then/_132/cond/cond/assert_greater_equal/Assert/Assert}}]]
0 successful operations.
0 derived errors ignored. [Op:__inference_test_function_1718670]

Function call stack:
test_function -> test_function
```

sararb:
Fixes #236
Fixes #233

### Goals :soccer:

- [x]  This PR implements a new logic of setting up the evaluation of retrieval models from a set of `evaluation_candidates`.
        - .evaluate() model will load the top-k evaluation indices from the provided implicit negatives `evaluation_candidates`, provided by the user.

- [x] It also fixes the calculation of eval ranking metrics that were always equal to 100%: 
         - Zeroes the score of accidental hits 
         - Additional pre-sorting step of the positive and negatives scores before applying the ranking metric 

- [x] Fixes batch_predict step so that the model is always returning the same embeddings of each user/item ids. 
         - Specify `shuffle=False` in the [data iterator](https://github.com/NVIDIA-Merlin/models/blob/9d887ebef954cddda568657ed1e3629aecb0debe/merlin/models/tf/utils/batch_utils.py#L171)

- [x] Make sure positive and negative scores are comparable:  Apply the `pre-call` processing step applied to positives in the ItemRetrievalTask to the top negative scores in the `pre_eval_topk` index block.


merlin/models/tf/models/base.py
```diff
@@ -470,13 +482,18 @@ def __init__(
         self,
         *blocks: Union[Block, ModelLikeBlock],
         context: Optional[BlockContext] = None,
+        evaluation_candidates=None,
```
gabrielspmoreira, line 485:
```suggestion
        evaluation_candidates: merlin.io.Dataset = None,
```

merlin/models/tf/models/base.py
```diff
@@ -470,13 +482,18 @@ def __init__(
...
+        evaluation_top_k=None,
```
gabrielspmoreira, line 486: Where is `evaluation_top_k` used?

merlin/models/tf/models/base.py
```diff
@@ -556,7 +573,21 @@ def item_embeddings(
...
+            and getattr(self.loss_block, "set_retrieval_cache_query", None)
+        ):
+            raise ValueError(
+                "Your retrieval model should contain an ItemRetrievalTask "
+                "in the end (loss_block)."
+            )
+
+    def set_evaluation_candidates(self, candidates: merlin.io.Dataset, k: int):
+        self.evaluation_candidates = candidates
+        self.k = 10
```
gabrielspmoreira, line 588: Should it be `self.k = k`  with arg `k: int = 10`?

merlin/models/tf/models/base.py
```diff
@@ -398,6 +397,19 @@ def evaluate(
...
+        # using evaluation_candidates as indices.
+        if isinstance(self, RetrievalModel):
+            self.check_for_retrieval_task()
+            if not self.loss_block.pre_eval_topk:
+                if not self.evaluation_candidates:
+                    raise ValueError(
+                        "You need to specify the set of negatives to use for evaluation "
+                        "via `evaluation_candidates` argument"
+                    )
+                self = self._load_topk_evaluation(self.evaluation_candidates, k=self.k)
```
gabrielspmoreira, line 411: We could maybe have no arguments in `_load_topk_evaluation()` as the arguments are already object (self) attributes.

merlin/models/tf/utils/mixins.py
```diff
@@ -50,23 +50,21 @@ class MetricsMixin(abc.ABC):
 
     def calculate_metrics(
         self,
-        inputs: Union[tf.Tensor, TabularData],
-        targets: Union[tf.Tensor, TabularData],
+        outputs,
```
gabrielspmoreira, line 53:
```suggestion
        outputs: PredictionOutput,
```

merlin/models/tf/utils/mixins.py
```diff
@@ -108,8 +106,7 @@ def compute_loss(
...
```
gabrielspmoreira, line 109:
```suggestion
        outputs: PredictionOutput,
```

tests/tf/models/test_retrieval.py
```diff
@@ -46,21 +46,28 @@ def test_two_tower_retrieval_model_with_metrics(ecommerce_data: SyntheticData, r
         loss="categorical_crossentropy",
     )
 
+    # Setting up evaluation
+    item_features = ecommerce_data.schema.select_by_tag(Tags.ITEM).column_names
+    item_dataset = ecommerce_data.dataframe[item_features].drop_duplicates()
+    item_dataset = item_dataset.sort_values("item_id")
```
gabrielspmoreira, line 52: Do we need to sort by "item_id"? If that is required, can maybe do that inside `model.set_evaluation_candidates()` ?

gabrielspmoreira (commented):
Just left some comments

tests/tf/models/test_retrieval.py
sararb: We don't need the sorting step, I've just added it for some internal tests but forgot to remove it.

benfred:
@sararb - I pulled this branch and still got the error I reported in #236 =(.

Does this snippet work for you ?

```python
# get the movielens dataset
from merlin.models.data.movielens import get_movielens
train, valid = get_movielens(variant="ml-25m")

# create a simple TF MatrixFactorizationModel on the movielens dataset, and fit it
import merlin.models.tf as ml
model = ml.MatrixFactorizationModel(train.schema, dim=64)
model.compile(optimizer="adam")
model.fit(train, batch_size=1024)

# try evaluating the model
metrics = model.evaluate(valid, batch_size=1024)
```

sararb:
> @sararb - I pulled this branch and still got the error I reported in #236 =(.
> 
> Does this snippet work for you?
> 
> ```python
> # get the movielens dataset
> from merlin.models.data.movielens import get_movielens
> train, valid = get_movielens(variant="ml-25m")
> 
> # create a simple TF MatrixFactorizationModel on the movielens dataset, and fit it
> import merlin.models.tf as ml
> model = ml.MatrixFactorizationModel(train.schema, dim=64)
> model.compile(optimizer="adam")
> model.fit(train, batch_size=1024)
> 
> # try evaluating the model
> metrics = model.evaluate(valid, batch_size=1024)
> ```

Thanks, @benfred, for the code snippet! I was able to reproduce the error, and there are two issues here: 

1. To evaluate a retrieval model using `brute force top-k evaluation`, the user needs to specify additional steps after the model definition and which are : 
```
    # Setting up evaluation
    item_dataset = model._ensure_unique(train, tag=Tags.ITEM, id_tag=Tags.ITEM_ID) # build dataset of unique item features
    model.set_evaluation_candidates(item_dataset, k=5) # set the index for brute-force top-k
 ``` 
 
2. While testing your code, I noticed that `MatrixFactorizationModel` is returning the base Model class instead of the RetrievalModel [one](https://github.com/NVIDIA-Merlin/models/blob/6d341459802cbda96a6d6643b8d1c87a124ed1f0/merlin/models/tf/models/base.py#L466), which includes additional functions needed for evaluating/exporting the Retrieval-like models.

So would you mind testing with a `TwoTowerModel` and the code snippet of setting up the evaluation? I will open an issue related to `MatrixFactorizatiomModel`.

gabrielspmoreira (approved):
I have been checking this PR and discussing the solution with @sararb along the way. Have just tested the changes in my local training scripts and it seems metrics provided by model.evaluate() are consistent now. Just pushed minor fixes to the PR, and it is good to be merged

PR #257, opened 2022-03-15
Set up automated docstring coverage checks

karlhigley:
This sets up a docstring coverage tool in the pre-commit hooks and checks for 25% coverage as a starting point.

(Hoping to get that much higher in all our libraries, but also don't want to block every PR on docstrings. We can bump the threshold as coverage improves to check for regressions.)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,6 +17,11 @@ repos:
     rev: v0.3.1
     hooks:
       - id: absolufy-imports
+ - repo: https://github.com/econchick/interrogate
+        rev: 1.5.0
+        hooks:
+        - id: interrogate
+          args: [-viImMpPsSCn, -f, "25.0", merlin]
   # yaml formatting
   - repo: https://github.com/pre-commit/mirrors-prettier
     rev: v2.3.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,6 +22,7 @@ repos:
     hooks:
       - id: interrogate
         args: [-viImMpPsSCn, -f, "25.0", merlin]
+        exclude: tests
   # yaml formatting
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -17,6 +17,11 @@ repos:
     rev: v0.3.1
     hooks:
       - id: absolufy-imports
+  - repo: https://github.com/econchick/interrogate
+    rev: 1.5.0
+    hooks:
+      - id: interrogate
+        args: [-viImMpPsSCn, -f, "25.0", merlin]
```
benfred, line 24: When I was push out https://github.com/NVIDIA-Merlin/core/pull/44 - I noticed that the pre-commit hook with interrogate was failing the change - and I had to disable the pre-commit hook to commit the change.  

It appears that the interrogate pre-commit hook was including files from the tests/ directory (which you can verify by running `pre-commit run --all-files` on merlin-core) . Changing the pyproject.toml file didn't seem to work, nor did passing the '--exclude=tests' to the comamndline here.

This does seem to work though (like tell pre-commit to ignore tests rather than tell interrogate to ignore tests)

```suggestion
        args: [-viImMpPsSCn, -f, "25.0", merlin]
        exclude: tests
```

karlhigley: Huh, that's interesting. My impression was that it only checks the supplied directory (`merlin`.)

benfred: It sorta looked like pre-commit was passing a list of all the changed files to interrogate - which was overriding the other settings. Like it was running interrogate on the merlin directory, and the changed tests (rather than just the merlin directory) - but not on the tests that I didn't touch

karlhigley: Ahhhh, that makes sense

benfred (approved):

Issue #261, opened 2022-03-16
[BUG] Adding regularization to MLPBlock leads to serialization error on model.save()


gabrielspmoreira:
### Bug description
When we provide to `MLPBlock` regularizers (e.g. `ml.MLPBlock([64], kernel_regularizer=regularizers.l2(1e-1), bias_regularizer=regularizers.l2(1e-1))`) to build a model, then `model.save()` raises the following error.

```
TypeError: Exception encountered when calling layer "sequential_block_4" (type SequentialBlock).

Tensor is unhashable. Instead, use tensor.ref() as the key.

Call arguments received:
[...]

PR #264, opened 2022-03-17
Fixing MLPBlock saving with regularizers

marcromeyn:
Fixes 
- #261

### Goals :soccer:
Making it possible to save a model that contains a `MLPBlock` with regularizers.

Bug was in the `SequentialBlock` since a tf.Tensor is not hashable.

### Testing Details :mag:
Added an extra test to ensure the fix works correctly.


gabrielspmoreira (approved):
Perfect! Thanks @marcromeyn

PR #266, opened 2022-03-17
Fix mypy issues & make MF return a RetrievalModel

### LINKED ISSUE #250: [BUG] `MatrixFactorizationModel` does not return `RetrievalModel` object
### Bug description

`MatrixFactorizationModel` is set by default to use ItemRetrievalTask but it returns a `Model` object instead of `RetrievalModel`. The latter class includes additional methods specific to evaluating/exporting retrieval models, such as: _ensure_unique

```AttributeError: 'Model' object has no attribute '_ensure_unique'```

You can reproduce the bug using the following code: 
```
def matrix_factorization_bug():
    from merlin.models.data.synthetic import SyntheticData, _read_data
    music_streaming_data = SyntheticData("music_streaming", num_rows=100, read_data_fn=_read_data)
    music_streaming_data._schema = music_streaming_data.schema.remove_by_tag(Tags.TARGET)
    
    import merlin.models.tf as ml
    model = ml.MatrixFactorizationModel(music_streaming_data.schema, dim=64)
    # Set evaluation
    item_dataset = model._ensure_unique(music_streaming_data.dataset, tag=Tags.ITEM, id_tag=Tags.ITEM_ID)
    model.set_evaluation_candidates(item_dataset, k=10)
```
### Expected behavior

MatrixFactorizationModel() with an ItemRetrievalTask prediction should return an `RetrievalModel` object. 

### Additional context
 This [line](https://github.com/NVIDIA-Merlin/models/blob/bc2500a1872c6cfc4c7f2ab90d6583759c0ccb8a/merlin/models/tf/blocks/core/base.py#L381) checks if the block contains `item_block` and `query_block` to convert it to a `RetrievalModel`. So maybe we could redefine [QueryItemIdsEmbeddingsBlock](https://github.com/NVIDIA-Merlin/models/blob/6d341459802cbda96a6d6643b8d1c87a124ed1f0/merlin/models/tf/blocks/retrieval/matrix_factorization.py#L27) to return a class with  `item_block` and `query_block` methods. And add an additional check [here](https://github.com/NVIDIA-Merlin/models/blob/bc2500a1872c6cfc4c7f2ab90d6583759c0ccb8a/merlin/models/tf/blocks/core/base.py#L382) that makes sure the Prediction Task is `ItemRetrievalTak`

marcromeyn:
Fixes #250

### Goals :soccer:
Fix MF issue + fix all mypy issues


sararb (approved):
Thanks for the fix, it looks good to me!

PR #267, opened 2022-03-17
Moving RetrievalModel._ensure_unique to utils

### LINKED ISSUE #263: [FEA] _ensure_unique() should be part for Dataset object
# 🚀 Feature request

currently we get unique item and user features with the following command:

```
model._ensure_unique(Dataset('/workspace/retrieval/processed/train/*.parquet', schema=schema), tag=Tags.ITEM, id_tag=Tags.ITEM_ID).comp

```

This method should be function of Dataset object.

marcromeyn:
Fixes #263


merlin/models/utils/dataset.py
```diff
@@ -49,6 +50,18 @@ def dataset_to_coo(dataset: Dataset):
...
 
 
+def unique_by_tag(dataset: Dataset, tag: Union[str, Tags], id_tag: Union[str, Tags]):
+    # Check if merlin-dataset is passed
+    ddf = dataset.to_ddf() if hasattr(dataset, "to_ddf") else dataset
+
+    columns = dataset.schema.select_by_tag(tag).column_names
+    if columns:
+        id_col = dataset.schema.select_by_tag(id_tag).first.name
+        ddf = ddf[columns].groupby(id_col).agg("first").reset_index()
```
gabrielspmoreira, line 60: This line causes a dask_cudf error in my dev environment: `RecursionError: maximum recursion depth exceeded while calling a Python object`. I replaced by `drop_duplicates`, whose result is equivalent and then the error does not occur. It might be an issue with the package versions in my environment, but I think it would be safer using drop_duplicates rather than groupby in any case.
```suggestion
        ddf = ddf[columns].drop_duplicates(id_col, keep="first")
```

merlin/models/utils/dataset.py
```diff
@@ -49,6 +50,18 @@ def dataset_to_coo(dataset: Dataset):
     return coo_matrix((targets.astype("float32"), (userids, itemids)))
 
 
+def unique_by_tag(dataset: Dataset, tag: Union[str, Tags], id_tag: Union[str, Tags]):
```
gabrielspmoreira, line 53: What about renaming to something more intuitive like `unique_rows_by_features(dataset, features_tag, grouping_tag)` ?

gabrielspmoreira (approved):
@marcromeyn Are we not removing RetrievalModel._ensure_unique() now too?

marcromeyn:
@gabrielspmoreira good catch, I removed it now

PR #271, opened 2022-03-17
Implementation of DeepFM model

### LINKED ISSUE #235: DeepFM Model Implementation
https://github.com/ChenglongChen/tensorflow-DeepFM

### LINKED ISSUE #270: [FEA] Categorical one-hot encoding
# 🚀 Feature request

Create a Tabular Transformation Block that converts a dict of encoded categorical ids to a one-hot vector representation. 

## Motivation

- The main motivation is that the DeepFM model expects categorical features as a binary tensor to compute the first-order interactions between raw features (continuous + categorical). 

- This Transformation can be also used to transform targets to a one-hot format. As this format is required by the ranking metrics for example.

sararb:
Fixes #235 
Fixes #270 

### Goals :soccer:
- This PR implements DeepFM architecture as a ranking model.
- It also introduces two blocks: 
       - `FMPairwiseInteraction` to learn pairwise interactions from a set of features embeddings  (Categorical and Continuous)
       - `CategoricalOneHot` to transform categorical inputs into one-hot vector representation needed for computing the first-order interactions of the FM model.


merlin/models/tf/models/ranking.py
```diff
@@ -152,3 +160,68 @@ def DCNModel(
...
+    Returns
+    -------
+    Model
+
+    """
+    num_features = len(schema.column_names)
+    cardinalities = schema_utils.categorical_cardinalities(schema)
+
+    input_block = InputBlock(
+        schema, embedding_options=EmbeddingOptions(embedding_dim_default=embedding_dim), **kwargs
```
marcromeyn, line 207: I’ve seen this pattern in a few different places lately, not exactly sure if it’s the best way of doing this. This because we are only exposing one argument of `EmbeddingOptions`. Should we maybe add an extra argument (maybe like `embedding_options_kwargs`) to `DeepFMModel`?

merlin/models/tf/models/ranking.py
```diff
@@ -152,3 +160,68 @@ def DCNModel(
...
+    )
+
+    pairwise_block = TabularBlock(StackFeatures(axis=-1)).connect(FMPairwiseInteraction())
```
marcromeyn, line 210:
```suggestion
    pairwise_block = FMPairwiseInteraction().prepare(aggregation=StackFeatures(axis=-1))
```

merlin/models/tf/models/ranking.py
```diff
@@ -152,3 +160,68 @@ def DCNModel(
...
+    deep_block = TabularBlock(ConcatFeatures()).connect(deep_block)
```
marcromeyn, line 211:
```suggestion
    deep_block = deep_block.prepare(aggregation=ConcatFeatures())
```

merlin/models/tf/models/ranking.py
```diff
@@ -152,3 +160,68 @@ def DCNModel(
...
+    first_order_block = CategoricalOneHot(cardinalities).connect(
+        [
```
marcromeyn, line 213: `connect` can take in multiple arguments ([source](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/blocks/core/base.py#L353)), so `[` and `]` can be removed here

merlin/models/tf/models/ranking.py
```diff
@@ -152,3 +160,68 @@ def DCNModel(
...
+
+def DeepFMModel(
+    schema: Schema,
+    embedding_dim: int,
+    deep_block: Optional[Block] = MLPBlock([64, 128]),
+    prediction_tasks: Optional[
+        Union[PredictionTask, List[PredictionTask], ParallelPredictionBlock]
+    ] = None,
+    **kwargs
+):
```
marcromeyn, line 173:
```suggestion
) -> Model:
```

tests/tf/blocks/core/test_transformations.py
```diff
@@ -148,3 +148,24 @@ def test_expand_dims_axis_as_dict():
     assert list(expanded_inputs["cont_feat1"].shape) == [NUM_ROWS]
     assert list(expanded_inputs["cont_feat2"].shape) == [1, NUM_ROWS]
     assert list(expanded_inputs["multi_hot_categ_feat"].shape) == [NUM_ROWS, 1, 4]
+
+
+def test_categorical_one_hot_encoding():
+
```
marcromeyn, line 154:
```suggestion

```

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -436,6 +436,51 @@ def call(self, inputs, training=False, **kwargs) -> tf.Tensor:
         return logits
 
 
+@Block.registry.register_with_multiple_names("categorical_to_onehot")
```
marcromeyn, line 439: This won’t work to expose it as a string since this class in it’s current form requires an input argument (`cardinalities`). It would be better to refactor it a little bit and infer the cardinalities from the Schema. There’s a way to have this class be passed the schema automatically. See [this](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/blocks/core/aggregation.py#L237) transformation on how to do that.

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -436,6 +436,51 @@ def call(self, inputs, training=False, **kwargs) -> tf.Tensor:
...
+
+    def call(self, inputs: TabularData, **kwargs) -> TabularData:
+        outputs = {}
+        for name, val in inputs.items():
+            outputs[name] = tf.squeeze(tf.one_hot(val, self.cardinalities[name]))
+        return outputs
+
+    def compute_output_shape(self, input_shape):
+        outputs = {}
+        for key, val in input_shape.items():
```
gabrielspmoreira, line 471: Could we have logic issues here with 3D sequential features?

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -436,6 +436,51 @@ def call(self, inputs, training=False, **kwargs) -> tf.Tensor:
...
+            if len(val) == 3:
+                outputs[key] = tf.TensorShape((val[0], val[1], self.cardinalities[key]))
+            else:
+                outputs[key] = tf.TensorShape((val[0], self.cardinalities[key]))
+        return outputs
+
+    def get_config(self):
+        config = super().get_config()
+        config["caridinalities"] = self.cardinalities
```
gabrielspmoreira, line 480: There is a typo in "caridinalities"

merlin/models/tf/blocks/core/transformations.py
sararb: Right, thanks for the example! I updated the class accordingly.

merlin/models/tf/blocks/core/transformations.py
sararb: Meaning we’ll have a 4D tensor as input right? 
If so, I believe say merlin-models currently does not support such inputs as all the building blocks are defined based on 2-D and/or 3-D tensors. So I think it would be a future feature request where we’ll update the merlin-models blocks in order to support such inputs.

merlin/models/tf/models/ranking.py
sararb: Yes make sense, the user could have more flexibility on how to set the embeddings. I've added `embedding_options_kwargs`.

PR #272, opened 2022-03-18
Fix small bug when shuffle is passed in .fit

marcromeyn:
### Goals :soccer:
Right now when shuffle is passed in in kwargs in fit + validation_data we would get an error since the shuffle argument is passed in twice in BatchedDataset. This PR fixes this.


gabrielspmoreira (approved):
Sounds good

PR #275, opened 2022-03-18
Fix bug in tower saving where all features were required

### LINKED ISSUE #273: [BUG] Saved TwoTowerModel Query Block requires both item and user features as inputs
### Bug description
After training a two tower model, when you try to extract the user block (query tower) the model requires the both item and user features. 

https://github.com/NVIDIA-Merlin/models/blob/poc/examples/POC/ItemRetrieval-TwoTower.ipynb 
Run notebook through Save User Tower section and then check the query_tower object created with 
`query_tower.__dict__`
You will see the extra item features in the model spec.

### Expected behavior
Expect query block to only require the user features.

### Additional context
This is blocking the end to end Ensemble API example.

marcromeyn:
Fixes #273

### DIFF
--- a/merlin/models/tf/blocks/retrieval/base.py
+++ b/merlin/models/tf/blocks/retrieval/base.py
@@ -70,9 +70,10 @@ class DualEncoderBlock(ParallelBlock):
     ):
         self._query_block = TowerBlock(query_block)
         self._item_block = TowerBlock(item_block)
+
         branches = {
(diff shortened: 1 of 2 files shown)


karlhigley:
Have to admit that the fix is a bit mystifying to me, but if the test passes, looks like it should solve our issue. 👍🏻

karlhigley (approved):

PR #294, opened 2022-03-23
fix evaluation loss equal to zero and add docstrings

### LINKED ISSUE #291: [BUG] getting the loss value as `0.00` from model.evaluate() for Two-Tower retrieval model
### Bug description
When I perform the `model.evaluate` step after training a TT model. I get the following metrics printed out.

`model.evaluate(valid, return_dict=True, batch_size=1024)`
```
{'recall_at_10': 2.1537676730076782e-05,
 'ndcg_10': 9.275297088606749e-06,
 'loss': 0.0,
 'regularization_loss': 0.0,
 'total_loss': 0.0}
```

Please run the Two-Tower retrieval notebook [here](https://github.com/NVIDIA-Merlin/models/blob/retrieval/examples/04-Retrieval-Model.ipynb) to reproduce.

### Expected behavior
Given ndcg@10 value,  Loss value cannot be 0.00. This seems off.

### Environment details
- Merlin version: 
- Platform:
- Python version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?):

using `nvcr.io/nvidia/merlin/merlin-tensorflow-training:22.03` docker image with the latest main branch pulled.

sararb:
Fixes #291 

### Goals :soccer:

- This PR fixes the issue related to evaluation loss of retrieval models equal to 0, reported by bug #291. This is due to the fact that ItemRetrievalTask is returning only positive scores during the `eval` mode. However, to compute the loss we will need negative ones. 

- It also updates the docstrings to meet the 25% coverage requirement.

- This is a quick fix where the top-k index is called before computing the loss to get the negatives scores. 

- A long-term solution, would be to externalize the calculation logic of loss and metrics of the retrieval tasks so as to have access to all needed tensors in one place and compute the right metrics based on train/eval modes.


gabrielspmoreira (approved):
This solution seems good to me. I tested the branch locally and it doesn't change the ranking metrics results, only adds the loss

PR #301, opened 2022-03-28
first implementation of wide&deep model

### LINKED ISSUE #154: Add Wide&Deep model
For this we first need:
#153

sararb:
Fixes #154 

### Goals :soccer:
- This PR includes an experimental implementation of Wide&Deep model where both components are jointly trained using a single optimizer. 
- A cross-feature layer is defined to add high-order interaction features to the wide component.

### Testing Details :mag:
- Unit test to check the training/evaluation of the Wide&Deep model. 
- Unit test to catch the `ValueError` related to wrong keys names being provided to CrossFeatures.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,49 @@
+# Merlin Models Example Notebooks
+
+The example notebooks demonstrate how to use Merlin Models with TensorFlow on a variety of datasets.
+
+## Inventory
+
+1. **[Getting Started](01-Getting-started.ipynb)**: It provides a getting started with Merlin Models by training [Facebook's DLRM](https://arxiv.org/pdf/1906.00091.pdf) architecture with only 3 commands.
+
+2. **[Merlin Models and NVTabular: Applying To your own dataset](02-Merlin-Models-and-NVTabular-applying-to-your-own-dataset.ipynb)**: It demonstrates how the `schema` connects the ETL and training step. It will help to apply Merlin Models to your own dataset structures.
+
+3. **[Exploring different Ranking Models](03-Exploring-different-models.ipynb)**: The example will explore different ranking model architectures, such as [Neural Collaborative Filtering (NCF)](https://arxiv.org/pdf/1708.05031.pdf), MLP, [DRLM](https://arxiv.org/abs/1906.00091), and [Deep & Cross Network (DCN)](https://arxiv.org/pdf/1708.05123.pdf).
+
+4. **[Retrieval Model](04-Retrieval-Model.ipynb)**: It trains a [Two Tower architecture](https://dl.acm.org/doi/abs/10.1145/3366423.3380130) as a item retrieval task with Merlin Models.
+
+## Running the Example Notebooks
+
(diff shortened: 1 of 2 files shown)


EvenOldridge:
Can we merge this and flag the optimizer issue.

marcromeyn:
This PR uses the feature-column API for feature-crosses. This is not something we want to merge now. From talking to @benfred it seems like we actually support feature-crosses in NVT, although sounded like it’s a bit hidden. I think we need to create an example how to do it, and then incorporate it in the wide&deep model. It might also make sense to create a tag for a feature-cross that NVT could output.

karlhigley:
From reviewing the Wide and Deep paper, they perform feature crosses as part of the model architecture, so I don't think we want to move the responsibility for feature crossing to NVT. Seems like what would be required to move this PR forward would be an implementation of feature crossing in Merlin Models that doesn't rely on the feature column API. What would be involved in creating one?

marcromeyn:
I agree with you that the ideal case is to make feature-crosses part of Merlin Models but this will require extra work. Since we already seem to be supporting it in NVT, I was thinking that that would be the minimum scope to have a wide&deep model in Merlin Models.

We could create a separate ticket to add feature-crosses to Merlin Models. Not sure as of now, what the priority should be to work on that ticket though. 

Curiuous to your thoughts (cc @EvenOldridge @benfred)

marcromeyn:
I don’t think we ever made a proper definition of done, so not sure it’s competely fair to say that we have comitted to a W&D model that contains feature-crosses (as opposed to doing the feature-crossing in NVT).

But if you have time to help, it seems totally doable to add it though. I was under the impression that the string hashing-ops were CPU-only in TF, but it seems like that’s no longer the case ([PR](https://github.com/tensorflow/tensorflow/pull/47936)). This would mean that we could concat the ids of the columns you want to cross: `IDfeature1_IDfeature2_IDfeature3` and hash this. We could infer the cardinality of this by multiplying the cardinalities of each of the features. The user should be able to overwrite this though (to enable less buckets where embeddings are shared).

We have to figure out how to enable a nice API for the user to provide the feature-crosses.

karlhigley:
Until we either have Wide&Deep merged or we've explicitly decided not to add it, whatever is required to make it happen seems to me like part of the scope we've committed ourselves to. 🤷🏻 

Do we specifically need to use hashing, or can we exhaustively compute the pairwise feature crosses?

marcromeyn:
The ideal end state of the wide&deep model:
- One line to create the model
- Different optimizers for wide & deep part
- Feature crossing as part of the modeling code

We could either commit to do all these things in one go or to add things gradually. Up until this point we have been informally talking about the gradual approach, but totally fine to me if we want to do it all in go.

sararb:
I am closing this PR, as a new [PR](https://github.com/NVIDIA-Merlin/models/pull/502/files) was opened by @timmy00. This PR solves the first point: `One line to create the model`.  So I also added the three points needed for completing the wide&Deep model in this[ RMP ticket ](https://github.com/NVIDIA-Merlin/Merlin/issues/278)

gabrielspmoreira:
I agree feature crossing should be done in the modeling side, so that a Data Scientist would not have to preprocess multiple versions of the dataset to try out different feature interactions.

PR #313, opened 2022-03-31
Merlin models examples fixes

gabrielspmoreira:
- Adjustments in example notebooks 1 and 2 from the bug bash, like text editing and small improvements (e.g. returning metrics dict instead of list)
- Improving the text in general
- Adding instructions on how to use synthetic data (so that the user is aware)
- Renaming the example 2 notebook from "Applying to your own dataset with Merlin Models and NVTabular" to "From ETL to Training RecSys models - NVTabular and Merlin Models integrated example"

### DIFF
--- a/examples/01-Getting-started.ipynb
+++ b/examples/01-Getting-started.ipynb
@@ -136,7 +136,7 @@
    "id": "688b89c7",
    "metadata": {},
    "source": [
-    "We define the DLRM model, whose prediction task is a binary classification. From the `schema`, the categorical features are identified (and embedded) and the target column is also automatically inferred, because of the schema tags. We talk more about the schema in the example notebook [02-Merlin-Models-and-NVTabular-applying-to-your-own-dataset](02-Merlin-Models-and-NVTabular-applying-to-your-own-dataset),"
+    "We define the DLRM model, whose prediction task is a binary classification. From the `schema`, the categorical features are identified (and embedded) and the target column is also automatically inferred, because of the schema tags. We talk more about the schema in the next [example notebook (02)](02-Merlin-Models-and-NVTabular-integration.ipynb),"
    ]
   },
   {
--- a/examples/01-Getting-started.ipynb
+++ b/examples/01-Getting-started.ipynb
@@ -7,7 +7,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "# Copyright 2021 NVIDIA Corporation. All Rights Reserved.\n",
+    "# Copyright 2022 NVIDIA Corporation. All Rights Reserved.\n",
     "#\n",
     "# Licensed under the Apache License, Version 2.0 (the \"License\");\n",
     "# you may not use this file except in compliance with the License.\n",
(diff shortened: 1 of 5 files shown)


examples/02-Merlin-Models-and-NVTabular-applying-to-your-own-dataset.ipynb
```diff
@@ -7,7 +7,7 @@
    "metadata": {},
```
bschifferer, line 7: <p>I think "Applying to your own dataset" has a higher motivation factor for users to read the section than "integration example"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

bschifferer, line 7: <p>I think "Applying to your own dataset" has a higher motivation factor for users to read the section than "Integratd pipeline with"</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

bschifferer, line 7: <p>Why do we add "name+"_02"" "_02" to the path?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='36'/>

examples/01-Getting-started.ipynb
```diff
@@ -7,7 +7,7 @@
    "metadata": {},
```
marcromeyn, line 7: <p>We could add a bit of description here around how to use synthetic data instead </p><pre class="ql-syntax" spellcheck="false">data = generate_data(“movielens-1m", num_rows=10000)
</pre>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='4'/>

examples/02-Merlin-Models-and-NVTabular-applying-to-your-own-dataset.ipynb
marcromeyn, line 7: <p>Mention synthetic data option here as well.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/313/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='3'/>

gabrielspmoreira: <p>I agree "Applying to your own dataset" caches more the attention of the user. I just suggested the change bcs as I user, from that title I would hope a notebook with placeholders to provide my data path and some config option to run easily ETL on my own dataset, like in this [example notebook (section Config)](https://github.com/rapidsai/recsys/blob/main/recsys_dataset_profiles/template_recsys_dataset_profiler.ipynb).&nbsp;</p><p> </p><p>I think in this notebook we use movielens to demo the integration between NVTabular and Models using schema which is great, but the notebook does make it straightforward applying it to your own dataset, IMHO.</p><p>It is up to you if my point makes sense (and updating it to Integration example). If not I can return the previous title, just let me know your thoughts on that.</p>

gabrielspmoreira: <p>See my comment to the previous message</p>

gabrielspmoreira: <p>The reason was that the 1st example notebook was also saving to the same output path, potentially causing some. For example, in example notebook 2 I accidentally skip the step that runs the NVT workflow, and then in the models part of the notebook I end up using the dataset preprocessed on notebook example 1 (saved in the same path), which has different features than the ones expected.</p>

bschifferer: <p>I think we can assume that the user runs the notebooks end-to-end and do not skip code cells. If that is the case, we do not need to capture the edge case? I think there could be many breaking points, if users change code / do not run all cells</p>

bschifferer: <p>I understand the point - but the title does not communicate why this example is important for a user. A user do not know what Merlin / NVTabular really is and why the integration is important?</p>

examples/01-Getting-started.ipynb
bschifferer: I would not add synthetic data in the Getting Started notebook. This should be a really easy example. MovieLens 1M takes <1min to download and extract.

bschifferer:
If we change the 2nd notebook name, can you update the README for examples, as well @gabrielspmoreira ?

examples/02-Merlin-Models-and-NVTabular-applying-to-your-own-dataset.ipynb
gabrielspmoreira: I have changed the example title to "From ETL to Training RecSys models - NVTabular and Merlin Models integrated example" to highlight this example is about understanding the full preprocessing and training pipeline

gabrielspmoreira: I just think users can potentially be skipping some cell when having multiple example notebooks open when getting to know the library. That mistake happened to me and took a few minutes for me to figure it out that I was using cached results from other notebook. We'd better avoiding that with this simple change (outputing preprocessed files in different folders for different examples).

examples/01-Getting-started.ipynb
gabrielspmoreira: I added @marcromeyn suggestion to the 2nd example notebook, and kept Getting started as it is to keep it clean. In particular bcs generate_data also accepts schema instead of dataset name, and that concept is only introduced in the 2nd example.

examples/02-Merlin-Models-and-NVTabular-applying-to-your-own-dataset.ipynb
gabrielspmoreira: I included a markdown explaining what is our synthetic data generation engine and a code snippet to do so.

examples/01-Getting-started.ipynb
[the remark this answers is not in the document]
gabrielspmoreira: Fixed the links to example notebook 02 from the example notebooks 01 and 04, and also from the readme

sararb (approved):

PR #317, opened 2022-03-31
Pairwise losses fixes

### LINKED ISSUE #252: [BUG] For large batch sizes pairwise losses are leading to NaN loss
I have observed in experiments with large batch sizes (for Ali-CCP and H&M datasets) the pairwise losses sometimes lead to NaN loss.
I am investigating this issue

### LINKED ISSUE #316: [BUG] Pairwise losses fail on model.evaluate() of retrieval models
### Bug description
The pairwise losses are raising an error when `model.evaluate()` is called for an item retrieval model.

gabrielspmoreira:
Closes #316 , closes #252

This PR fixes issues related to pairwise loss:

- [x] Refactory of the pairwise losses to centralize common logic into the parent `PairwiseLoss` class and to make them receive the `valid_negatives_mask` for ignoring the loss of false negatives
- [x] Now `PredictionOutput` named tuple has a method `copy_with_updates()` that creates a copy of the tuple but allowing to update only the necessary attributes. That prevents potential issues where some blocks only change for example the `predictions` or `targets` and forget to include the original `sample_weight` or `valid_negatives_mask` in the new tuple instance
- [x] Fixed error on pairwise losses error when calling `model.evaluate()`
- [x] Solves `NaN` loss when training using pairwise losses with larger batch sizes. The Nan issue was caused by computing log(0) for some predictions of false negatives. Now we add a small epsilon to 0 values before `tf.math.log()`

### DIFF
--- a/merlin/models/tf/blocks/core/base.py
+++ b/merlin/models/tf/blocks/core/base.py
@@ -52,8 +52,8 @@ class PredictionOutput(NamedTuple):
 
     def copy_with_updates(
         self,
-        predictions: Union[TabularData, tf.Tensor] = None,
-        targets: Union[TabularData, tf.Tensor] = None,
+        predictions: Optional[Union[TabularData, tf.Tensor]] = None,
+        targets: Optional[Union[TabularData, tf.Tensor]] = None,
         positive_item_ids: Optional[tf.Tensor] = None,
         label_relevant_counts: Optional[tf.Tensor] = None,
         valid_negatives_mask: Optional[tf.Tensor] = None,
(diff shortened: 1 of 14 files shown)


merlin/models/tf/blocks/core/base.py
```diff
@@ -48,6 +48,35 @@ class PredictionOutput(NamedTuple):
     targets: Union[TabularData, tf.Tensor]
     positive_item_ids: Optional[tf.Tensor] = None
     label_relevant_counts: Optional[tf.Tensor] = None
+    valid_negatives_mask: Optional[tf.Tensor] = None
+
+    def copy_with_updates(
+        self,
+        predictions: Union[TabularData, tf.Tensor] = None,
```
marcromeyn, line 55: Type-hints here should be wrapped in `Optional[...]`

marcromeyn (approved):

viswa-nvidia:
@marcromeyn @benfred , I see that there were 9 skipped tests. Did we check what tests were skipped ?

PR #327, opened 2022-04-04
Fixing synthetic data-generation for aliccp-raw

marcromeyn:
### Goals :soccer:
- Fixing issues in end-to-end POC where we want to apply a nvt-transformation on synthetic data generated through `aliccp-raw`.
- Improving imports by moving `models/data` to top-level `datasets`.

In `generate_user_item_interactions` we were using `int64` for all features and therefore not always respecting the dtype from the schema. This PR fixes that.

### Testing Details :mag:
Updated `test_synthetic_aliccp_raw_data` to ensure that the data that gets generated can be used in a nvt-transform.


sararb:
rerun tests

merlin/datasets/ecommerce/aliccp/raw/schema.pbtxt
```diff
@@ -0,0 +1,294 @@
...
+    extra_metadata {
+      type_url: "type.googleapis.com/google.protobuf.Struct"
+      value: "\n\034\n\017dtype_item_size\022\t\021\000\000\000\000\000\000@@\nG\n\017embedding_sizes\0224*2\n\026\n\tdimension\022\t\021\000\000\000\000\000`j@\n\030\n\013cardinality\022\t\021\000\000\000\000\000\306\267@\n\025\n\010max_size\022\t\021\000\000\000\000\000\000\000\000\n\030\n\013start_index\022\t\021\000\000\000\000\000\000\000\000\n\021\n\013num_buckets\022\002\010\000\n\r\n\007is_list\022\002 \000\n:\n\010cat_path\022.\032,.//categories/unique.user_categories.parquet\n\017\n\tis_ragged\022\002 \000\n\033\n\016freq_threshold\022\t\021\000\000\000\000\000\000\000\000"
+    }
+  }
+}
+feature {
+  name: "click"
+  type: INT
+  annotation {
```
sararb, line 278: We need the tag BINARY_CLASSIFICATION  [here](https://github.com/NVIDIA-Merlin/models/blob/833b84c1575ce76e583450272f6ae1ee9fef4b34/merlin/models/data/synthetic.py#L234) to generate the corresponding vector with zeros and ones. 

I understand we need to remove the tags to avoid  tagging the target encoding variables we got from nvt workflow  with `Target` tag. So maybe we can remove the `TARGET` tag and keep the binary classification tag?

sararb (approved):
Look good to me, thanks! I've just left a comment related to how we generate synthetic binary targets.

viswa-nvidia:
👍

PR #329, opened 2022-04-04
docs: Add redirect page

mikemckiernan:
Automatically create a redirect page that
redirects to the index.html page for the default
branch of the repo.

Automatically check and create a dot-no-jekyll file.

### DIFF
--- a/.github/workflows/docs-sched-rebuild.yaml
+++ b/.github/workflows/docs-sched-rebuild.yaml
@@ -71,6 +71,46 @@ jobs:
             echo "Git adding ${i}"
             git add "${i}"
           done
+      - name: Check or create dot-no-jekyll file
+        run: |
+          if [ -f ".nojekyll" ]; then
+            echo "The dot-no-jekyll file already exists."
+            exit 0
+          fi
+          touch .nojekyll
+          git add .nojekyll
+      - name: Check or create redirect page
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/docs-sched-rebuild.yaml
```diff
@@ -71,6 +71,46 @@ jobs:
...
             git add "${i}"
           done
+      - name: Check or create dot-no-jekyll file
+        run: |
+          if [ -f ".nojekyll" ]; then
+            echo "The dot-no-jekyll file already exists."
+            exit 0
+          fi
+          touch .nojekyll
+          git add .nojekyll
```
benfred, line 81: doesn't the .nojekyll file get added by the `sphinx.ext.githubpages` https://github.com/NVIDIA-Merlin/models/blob/1abae0e2b60010c2bcd8cdb308b09cd98fdc3e75/docs/source/conf.py#L65  - or is this not playing well with sphinxmultiversion?

benfred (approved):
I like that we can automatically redirect!

.github/workflows/docs-sched-rebuild.yaml
mikemckiernan: Ah, look at that.  I didn't realize that would attempt to add the `.nojekyll` file.  Regardless of my ignorance, the file is added in the directory for each version, so `main/.nojekyll` and `v0.2.0/.nojekyll` and not in the root of the repo.  It was simple enough to add.

PR #352, opened 2022-04-08
Refactoring the evaulation of RetrievalModel

marcromeyn:
Fixes # (issue)


merlin/models/tf/models/base.py
```diff
@@ -13,8 +15,13 @@
...
+if TYPE_CHECKING:
+    from merlin.models.tf.blocks.core.index import TopKIndexBlock
+
 
 class MetricsComputeCallback(tf.keras.callbacks.Callback):
+    """Callback that handles when to compute metrics."""
+
     def __init__(self, train_metrics_steps=1, **kwargs):
         self.train_metrics_steps = train_metrics_steps
         self._is_fitting = False
```
gabrielspmoreira, line 27: I believe after removing `model._should_compute_eval_metrics_as_in_training`, the `self._is_fitting` is now useless

merlin/models/tf/models/base.py
```diff
@@ -328,48 +378,37 @@ def fit(
...
-            validation_steps,
-            validation_batch_size,
-            validation_freq,
-            max_queue_size,
-            workers,
-            use_multiprocessing,
+    def compute_loss_metrics(
+        self, inputs, targets, training: bool = False, compute_metrics=True, **kwargs
+    ):
+        predictions = self(inputs, training=training, **kwargs)
```
gabrielspmoreira, line 398: When I run `model.fit()` setting `validation_data` argument, it will call `model.evaluate()` in the end of each epoch. Then `RetrievalModel.compute_loss_metrics()` is called and sets [here](https://github.com/NVIDIA-Merlin/models/blob/tf/retrieval-eval/merlin/models/tf/models/base.py#L596) `kwargs["eval_sampling"] = True`, so that in the `ItemRetrievalScorer.call()` the `eval_sampling` is `True`.
But the `eval_sampling`  is not being passed through for some reason, which causes an error in `ItemRetrievalScorer` when using `model.fit(..., validation_data=...)`

gabrielspmoreira:
@marcromeyn I have reproduced some experiments with this refactory of the Retrieval models evaluation.
The final metrics computed with `model.evaluate(dataset, item_corpus=dataset, ...)`match the eval metrics I have got with the previous version, which is great. Also noticed that the `model.fit()` is much faster now for the first batch (as the logic to prepare the  `TopKIndexBlock` is now only done in model.evaluate(). I also noticed the run times dropped in about 2 min (e.g. 12->10, 13->11) with your changes.

P.s. I could not check the validation metrics during training (when you provide `validation_data` to `model.fit()`), because there is a bug (reported by me in my previous comment).

PR #357, opened 2022-04-12
Added TensorInitializer that can be used to initialize variables/embeddings with pretrained weights

### LINKED ISSUE #356: [FEA] Possibility to initialize embeddings with pretrained weights
# 🚀 Feature request
The `EmbeddingFeatures.from_schema()` accepts setting for each feature embedding initializers by setting`EmbeddingOptions.embeddings_initializers` dict, where TF [initializer](https://www.tensorflow.org/api_docs/python/tf/keras/initializers) can be used. But there is no straightforward way to initialize embeddings with a pre-trained weights tensor.

## Motivation

Tensorflow provides a number of [initializer](https://www.tensorflow.org/api_docs/python/tf/keras/initializers), but no one that can be used to initialize a variable or embedding with a tensor (e.g. obtained from pre-training)

gabrielspmoreira:
Added `TensorInitializer` that can be used to initialize variables/embeddings with pretrained weights.
Closes #356


merlin/models/tf/utils/tf_utils.py
```diff
@@ -248,3 +248,39 @@ def add_epsilon_to_zeros(tensor: tf.Tensor, epsilon: float = 1e-24) -> tf.Tensor
...
+        tf.Tensor
+            Returns the tensor set in the constructor
+        """
+        tf.assert_equal(shape, self._weights.shape)
+
+        weights = self._weights
+        if dtype:
+            weights = tf.cast(self._weights, dtype)
+        return weights
+
```
marcromeyn, line 284: Would be good to add a class-method `from_dataset` where we convert a merlin.io.Dataset to the weights.

gabrielspmoreira: I have added `TensorInitializer.from_dataset()` as you suggested

gabrielspmoreira: @marcromeyn I have added `TensorInitializer.from_dataset()` as you suggested. 
But end up finding a bug #406, as I noticed that `cudf.from_dlpack()` don't work properly with DLpack encoded 2D TF tensors. The fix of that issue depends on the issue [#10754](https://github.com/rapidsai/cudf/issues/10754) I created on `cudf` repo.
But I think we can merge this PR as `TensorInitializer.from_dataset()` is about loading the embeddings from a dataset, an the issue occurs when exporting the embeddings.

rnyak:
rerun tests

PR #362, opened 2022-04-12
First design of popularity correction

### LINKED ISSUE #161: Add logits correction for ItemsPredictionSampled (sampled softmax)
The `ItemsPredictionSampled` (previous SampledSoftmax) is doing popularity-based sampling but is missing the logits correction according to the items popularity. This is a pending issue from the PR  https://github.com/NVIDIA-Merlin/models/pull/152

sararb:
First implementation of popularity bias correction. 

- [x] First design of popularity sampling correction that uses a fixed frequency table
- [x] Store label ids for positive and negatives in the ItemRetrievalScorer block for the frequency table lookup 
- [x] Link  popularity sampling block to the Item Retrieval task

Closes #161


merlin/models/tf/blocks/core/transformations.py
```diff
@@ -556,3 +556,76 @@ def call_outputs(
...
+
+    TODO: Use the schema to infer the path to the item frequency parquet table
+
+    Parameters:
+    ----------
+    item_frequency : Union[tf.Tensor]
+        The item frequency table
+    """
+
+    def __init__(self, item_frequency: Union[tf.Tensor, str], **kwargs):
```
gabrielspmoreira, line 576: I think `item_frequency` should always be a tf.Tensor (not str), don't?

gabrielspmoreira, line 576: Maybe we should also have the schema as an argument, so that we retrieve the item_id cardinality from the schema and check if it matches the shape of `item_frequency`, just for defensive programming.

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -556,3 +556,76 @@ def call_outputs(
...
+        super().__init__(**kwargs)
+        candidate_probabilities = tf.cast(
+            item_frequency / tf.reduce_sum(item_frequency), tf.float32
+        )
+        self.candidate_probabilities = tf.Variable(
```
gabrielspmoreira, line 581: I think we should also have an `update()` method, in case we need to update the item popularities when fine tuning a pre-trained model

gabrielspmoreira:
We need to think how we can access the `PopularityLogitsCorrection.from_parquet()` from the higher-level building blocks (e.g. from `ItemRetrievalTask`)

merlin/models/tf/blocks/core/transformations.py
sararb: yes, I was testing different designs and forgot to remove the `str` type. As this is handled by the `from_parquet` method in the current implementation draft.

gabrielspmoreira (approved):
Sounds good to me! Just pushed a commit with some naming and documentation improvements, and added some checks and tests

PR #365, opened 2022-04-13
Fixes and improvements in embeddings export for MatrixFactorizationModel and TwoTowerModel

gabrielspmoreira:
- [x] Created a method to export the embedding vectors as a Tensor, in particular for MatrixFactorization.
- [x] Ensuring that when using RetrievalModel.query_embeddings() and .item_embeddings() the batch size is provided, as it is required to build the BatchedDataset under-the-hood

### DIFF
--- a/merlin/models/tf/features/embedding.py
+++ b/merlin/models/tf/features/embedding.py
@@ -249,7 +249,7 @@ class EmbeddingFeatures(TabularBlock):
     def table_config(self, feature_name: str):
         return self.feature_config[feature_name].table
 
-    def get_embedding_table(self, table_name: Union[str, Tags], l2_norm: bool = False):
+    def get_embedding_table(self, table_name: Union[str, Tags], l2_normalization: bool = False):
         if isinstance(table_name, Tags):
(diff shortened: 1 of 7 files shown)


merlin/models/tf/features/embedding.py
```diff
@@ -249,15 +249,28 @@ def lookup_feature(self, name, val, output_sequence=False):
     def table_config(self, feature_name: str):
         return self.feature_config[feature_name].table
 
-    def embedding_table_df(
-        self, table_name: Union[str, Tags], l2_norm: bool = False, gpu: bool = True
-    ):
+    def embedding_vector(self, table_name: Union[str, Tags], l2_norm: bool = False):
```
marcromeyn, line 252: Wondering it we should rename this to something else. This method name might make people think it would output an individual embedding instead of the whole table.

gabrielspmoreira: Makes sense. Renamed it to `get_embedding_table()`

sararb (approved):

Issue #167, opened 2022-02-04
Evaluation - Popularity-bias/novelty metric


gabrielspmoreira:
Metrics that computes the popularity bias of the top-n recommended items.
A common choice for this metric is [novelty](https://microsoft-recommenders.readthedocs.io/en/latest/evaluation.html#recommenders.evaluation.python_evaluation.historical_item_novelty) as `-log(norm_prob_item)`, where `norm_prob_item = item_count / total_items_count`, or just the average normalized popularity of top-k items like in [here](https://github.com/jacopotagliabue/reclist/blob/main/reclist/metrics/standard_metrics.py#L106).

The `popularity-bias` metric is also computed based on item popularities. It is describe in the end of page 15 of this [paper](https://arxiv.org/pdf/1803.09587.pdf). That paper implementation of this metric is available [here](https://github.com/rn5l/session-rec/blob/master/evaluation/metrics/popularity.py)

rnyak:
solved by https://github.com/NVIDIA-Merlin/models/pull/371

PR #371, opened 2022-04-18
Evaluation metrics of a top-k recommender 

### LINKED ISSUE #175: Evaluation - Item Coverage metric
Implement the `ItemCoverage` metric, that would receive in the constructor the `num_unique_items` and a cut-off `k`.
It should keep a fixed sized int tensor with `shape = (num_unique_items)` holding the count of items frequencies over batches, but only for the top-k predicted scores. The counts are not really needed (only for debugging maybe), it could also be just a bool tensor.
The metric result would be the something like `tf.cast(item_frequencies > 0, tf.float32) / tf.shape(item_frequencies)[0]`, i.e., the percentage of the item catalog that was recommended at least once among top-k for any example.

sararb:
Fixes #175 #167 

### Goals :soccer:
- Implement evaluation metrics for top-k recommender models: PopularityBias, Novelty, and Item coverage. 

- Add `get_candidate_probs` as tf util function to transform item frequencies into a probability distribution 
- Add custom Keras metrics to compute: popularity bias, novelty, and coverage from top-k predictions returned by a recommendation model. 

### Testing Details :mag:
- Add unit tests that check the scores returned by the implemented metrics.
- Update `test_topk_index` with evaluation metrics calculation.

### DIFF
--- a/merlin/models/tf/metrics/evaluation.py
+++ b/merlin/models/tf/metrics/evaluation.py
@@ -159,7 +159,8 @@ class PopularityMetric(Mean):
         self.candidate_probs.assign(candidate_probs)
 
     def reset_state(self):
-        super.reset_state()
+        # reset all metrics variables except `candidate_probs`
+        backend.batch_set_value([(v, 0) for v in self.variables if "candidate_probs" not in v.name])
 
     def check_cast_inputs(self, labels, predictions):
         tf.assert_equal(
@@ -298,7 +299,7 @@ class ItemCoverageAt(tf.keras.metrics.Metric):
         return tf.reduce_sum(coverage)
 
     def reset_state(self):
-        self.predicted_items_count.assign(tf.zeros((self.num_unique_items,), dtype=tf.uint32))
+        self.predicted_items_count = tf.zeros((self.num_unique_items,), dtype=tf.uint32)
 
     def check_cast_inputs(self, labels, predictions):
         tf.assert_equal(
--- a/merlin/models/tf/metrics/evaluation.py
+++ b/merlin/models/tf/metrics/evaluation.py
@@ -70,7 +70,6 @@ class PopularityMetric(Mean):
     fn:
         The popularity metric function to wrap,
         with signature `fn(predicted_candidates_probs, k, **kwargs)`.
-
     item_freq_probs: Union[tf.Tensor, Sequence]
         A Tensor or list with item frequencies (if is_prob_distribution=False)
         or with item probabilities (if is_prob_distribution=True)
--- a/merlin/models/tf/metrics/evaluation.py
+++ b/merlin/models/tf/metrics/evaluation.py
@@ -142,15 +142,6 @@ class PopularityMetric(Mean):
         )
         return super().update_state(matches, sample_weight=sample_weight)
 
-    def check_cast_inputs(self, labels, predictions):
-        tf.assert_equal(
-            tf.rank(predictions), 2, f"predictions must be 2-D tensor (got {predictions.shape})"
-        )
-        if labels is not None:
-            tf.assert_equal(tf.rank(labels), 2, f"labels must be 2-D tensor (got {labels.shape})")
-            labels = tf.cast(labels, tf.int32)
-        return labels, tf.cast(predictions, tf.int32)
-
     def update_candidate_probs(
         self, item_freq_probs: Union[tf.Tensor, Sequence], is_prob_distribution: bool = False
     ):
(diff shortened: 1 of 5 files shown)


merlin/models/tf/metrics/evaluation.py
```diff
@@ -0,0 +1,292 @@
...
+def novelty_at(
+    predicted_candidates_probs: tf.Tensor,
+    k: int = 5,
+) -> tf.Tensor:
+    """
+    Computes novely@K metric
+    ----------
+    {METRIC_PARAMETERS_DOCSTRING}
+    """
+    return -tf.math.log(predicted_candidates_probs[:, :k] + MIN_FLOAT)
```
gabrielspmoreira, line 50: I think we should rename `MIN_FLOAT` to `EPSILON`, as there is another `MIN_FLOAT` constant in `merlin.models.utils.constants` used to downscore the false negatives

merlin/models/tf/metrics/evaluation.py
```diff
@@ -0,0 +1,292 @@
...
+
+        candidate_probs = get_candidate_probs(item_freq_probs, self.is_prob_distribution)
+        self.candidate_probs = tf.Variable(
+            candidate_probs,
+            name="candidate_probs",
+            trainable=False,
+            dtype=tf.float32,
+            validate_shape=False,
+            shape=tf.shape(candidate_probs),
+        )
```
gabrielspmoreira, line 110: I think we could have a method `PopularityMetric.update_candidate_probs()` like [here](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/blocks/core/transformations.py#L673), to allow for users being able to update the probs.
That could be useful for example if training on streaming data, where the items popularity could be adjusted over time

merlin/models/tf/metrics/evaluation.py
```diff
@@ -0,0 +1,292 @@
...
+def popularity_bias_at(
+    predicted_candidates_probs: tf.Tensor,
+    k: int = 5,
+) -> tf.Tensor:
+    """
+    Computes popularity bias@K metric
+    ----------
+    {METRIC_PARAMETERS_DOCSTRING}
+    """
+    return tf.reduce_sum(predicted_candidates_probs[:, :k], -1) / k
```
gabrielspmoreira, line 62: I think `Mean` base class expects metric results to be 1D (batch size) or 2D (batch size x (top-k) values per example).
You can see that in [recall_at()](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/metrics/ranking.py#L60-L66) we return a 1D tensor. That, way, the base class can multiply the metrics per example by `sample_weight` (from `update_state()`), before computing the (weighted) average.
So I think here we should just return either `tf.reduce_mean(predicted_candidates_probs[:, :k], -1)` or `predicted_candidates_probs[:, :k]`

merlin/models/tf/metrics/evaluation.py
```diff
@@ -0,0 +1,292 @@
...
+        The prediction coverage of the recommendations at the k-th position.
+    """
+
+    def __init__(self, num_unique_items: int, k=5, name: str = None, dtype=None, **kwargs):
+        if name is not None:
+            name = f"{name}_{k}"
+        super().__init__(name=name, dtype=dtype)
+
+        self.k = k
+        self.num_unique_items = num_unique_items
```
gabrielspmoreira, line 220: Maybe we could make `self.num_unique_items` a scalar non-trainable variable, and have an `update()` method to allow for users to update `num_unique_items` (with variable assign), as that might be useful when training models with streaming data, as the (number of )available items at a given point of time might change depending on the domain (e.g. on news recommendation)

tests/tf/metrics/test_metrics_popularity.py
```diff
@@ -0,0 +1,59 @@
...
+    predictions = tf.convert_to_tensor(
+        [[1, 9, 8, 7, 6], [1, 4, 3, 2, 5], [5, 9, 8, 7, 6]], tf.int64
+    )
+    item_freq_probs = [0.1, 0.05, 0.2, 0.05, 0.1, 0.1, 0.15, 0.05, 0.1, 0.1]
+    return labels, predictions, item_freq_probs
+
+
+def test_novelty_at_k(popularity_metrics_test_data):
+    labels, predictions, item_freq_probs = popularity_metrics_test_data
+    metric1 = NoveltyAt(item_freq_probs=item_freq_probs, is_prob_distribution=True, k=3)
```
gabrielspmoreira, line 35: Maybe we should have `is_prob_distribution` as a test argument and test both scenarios (True and False)

gabrielspmoreira (approved):
The metrics cannot currently be included in the `PredictionTask` because they require to receive in `update_state()` the predicted item ids instead of the prediction scores (y_pred) and labels (y_true), as usual.
I have tested the metrics locally following the example from the unit test and they do work properly when been called after `model.fit()` and after exporting to topk recommender. So I am merging this PR to be able to use them for the experiments.
I have created an issue #372 to have those metrics compatible with the other metrics, which are available for training and evaluation.

PR #377, opened 2022-04-19
Adding unttests for example notebooks

bschifferer:
Adding unittests with testbook for notebooks (currently 01, 02 and 06). Other notebooks will follow with next PR.

https://github.com/NVIDIA-Merlin/Merlin/issues/205


bschifferer:
rerun tests

merlin/datasets/entertainment/movielens/1m-raw/ratings/schema.pbtxt
```diff
@@ -0,0 +1,68 @@
...
+feature {
+  name: "rating"
+  type: INT
+  int_domain {
+    name: "rating"
+    max: 6
+    is_categorical: true
+  }
+  annotation {
+    tag: "categorical"
```
marcromeyn, line 46: Should we tag this as target?

marcromeyn (approved):

merlin/datasets/entertainment/movielens/1m-raw/ratings/schema.pbtxt
bschifferer: This schema represents the raw data for unittests. The raw data does not provide any tags. I wanted to use as few tags as possible. Actually, we should remove the tags from all columns, but then the data generator did not work.

PR #378, opened 2022-04-19
unittest examples v2 [direct push]

bschifferer:


bschifferer:
duplication of another branch

Issue #287, opened 2022-03-22
[BUG] model.predict() is providing random top-k recommendations for RetrievalModel


gabrielspmoreira:
### Bug description
The `model.predict()` method is providing random top-k recommendations for `RetrievalModel`.

### How to reproduce
1 - Download the [sample dataset](https://drive.google.com/file/d/10LqUW768FxhaVL4rx8lVJar_Mhbx-MvI/view?usp=sharing)
2 - Run the following code as it is. It will print `2 - EVALUATION METRICS` with `'recall_at_1000': 0.1835` and also print a very close RECALL@1000 for `4 - MANUAL TOP-K PREDICTION`, which is expected.
3 - Uncomment the `BLOCK-01` in the code. The `2 - EVALUATION METRICS`  will be the same, but now you will now notice that for `4 - MANUAL TOP-K PREDICTION` the `RECALL@1000` will be close to random, as shuffling predictions gives a similar metric result (~ `0.019`). So it is weird that calling `topk_recommender_model.predict()` affects MANUAL TOP-K PREDICTION results (which does not use the `TopKIndexBlock`). This fact might help to isolate better what might be generating the issue.
4 - Uncomment the `BLOCK-02` in the code. Now metrics are computed also for `3 - TOP-K PREDICTION` using the Topk_recommender and `RECALL@1000` will be close to random predictions (~ `0.019`) for both `3 - TOP-K PREDICTION` and for `4 - MANUAL TOP-K PREDICTION` .

### Code to reproduce
```python
import tensorflow as tf

tf.keras.utils.set_random_seed(52)

import cupy
import merlin.models.tf as mm
import merlin.models.tf.dataset as tf_dataloader
import numpy as np
from merlin.io.dataset import Dataset
from merlin.models.tf.blocks.core.index import IndexBlock
from merlin.models.utils.dataset import unique_rows_by_features
from merlin.schema import Schema
from merlin.schema.io.tensorflow_metadata import TensorflowMetadata
from merlin.schema.tags import Tags
from nvtabular.ops import *
from tensorflow.keras import regularizers

data_path = "./data/processed"


schema = TensorflowMetadata.from_proto_text_file(
    data_path + "/train/"
).to_merlin_schema()


schema = schema.select_by_tag([Tags.ITEM_ID, Tags.USER_ID, Tags.ITEM, Tags.USER])

model = mm.TwoTowerModel(
    schema,
    query_tower=mm.MLPBlock(
        [128, 64],
[...]

PR #389, opened 2022-04-21
Fix top-k recommender to return correct top-k relevant items. 

### LINKED ISSUE #387: [BUG] model.top_k_recommender() is providing random recommendations
### Bug description
The `model.top_k_recommender()` returns a retrieval model that provides top-k recommended items from the user features, that are sent through the user tower.

The issue is that such topk model it is not providing the most relevant items after training, but rather random items.

1. Download this [dataset](https://drive.google.com/file/d/1Wqddy7Jrdt5MdAWC_YLZ53M6eFrpzf37/view?usp=sharing)
1. Run this [script](https://github.com/NVIDIA-Merlin/research/blob/main/retrieval_exp/scripts/retrieval_train_eval_v07.py) with the following command line. P.s. The argument `--eval_additional_metrics` is important, as it enables the manual computation of the recall (and coverage, novelty metrics). It should last about 5 min.

```bash
TF_GPU_ALLOCATOR=cuda_malloc_async python scripts/retrieval_train_eval_v07.py --dataset handm --wandb_exp_group test --model_type two_tower --loss categorical_crossentropy --two_tower_activation selu --neg_sampling inbatch --epochs 1 --lr_decay_steps 50 --output_path /results --data_path /data --log_to_wandb --optimizer adam --eval_batch_size 4096 --train_metrics_steps 100 --eval_on_train_set_steps 0 --xe_label_smoothing 0.1 --two_tower_mlp_layers 128 --two_tower_dropout 0.2 --two_tower_embedding_sizes_multiplier 3.0 --logits_temperature 1.1 --lr 0.0019999262749752942 --lr_decay_rate 0.96 --l2_reg 1.3844195004925041e-05 --train_batch_size 16384 --logq_correction_factor 1.0 --eval_additional_metrics
```

2. You will be able to see in the end that the `recall_at_500-final` metric (computed by model.evaluate()) is about `0.4043`, and the `recall_at_500-manual-final` we compute manually using the top-k predictions of Topk recommender is only `0.0674`.

### Expected behavior
Manual computation of recall metrics from topk recommender should match the metrics from model.evaluate()

### Notes
- This issue is preventing us to use the diversity and novelty metrics created in PR #371
- This issue might be related to the issue #287

sararb:
Fixes #387 #287

### Goals :soccer:
Using Gabriel's script, I tried to debug the issue causing the top-k recommender to return random top-k indices.

These are the main findings:  

- The item_embeddings returned by the item block are identical between the manual run ( Step 4 ) and the index values of the topk_recommender_model (Step 3)

- I ran topk_recommender_model twice on the same input (to check if there is any randomness): the outputs are identical

- The top-ids returned by manual iteration do not match the top ids returned by the topk_recommender_model. 

- Digging into the top-k recommender call, I noticed that the query embeddings from the manual calculation do not match the query embeddings returned by the topk_recommender_model.

- After comparing the query-block before and after calling `to_top_k_recommender()`, I noticed that the embeddings weights of the user tower change. 

- This change is due to calling the build method of the query tower after the model was converted. The build method is re-initializing the embeddings features [here](https://github.com/NVIDIA-Merlin/models/blob/d9853020d9e32854b3ad7d37ec79a8b33fa2c438/merlin/models/tf/blocks/core/base.py#L169)

- I set the `built` flag to True after converting the model to a `topk-recommender` to avoid calling the build method when calling the model. 

### Testing Details :mag:
- I added Gabriel's script for debugging as an integration test and added a check that ensures the output of the top-k recommender matches with the manually calculated top-ids. 

These are the scores results from the different tested ideas: 
image

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -717,8 +717,11 @@ class RetrievalModel(Model):
             )
         else:
             topk_index = item_corpus
-        recommender = ModelBlock(self.retrieval_block.query_block().connect(topk_index))
-
+        # Set the blocks for recommenders with built=True to keep pre-trained embeddings
+        recommender_block = self.retrieval_block.query_block().connect(topk_index)
+        recommender_block.built = True
+        recommender = ModelBlock(recommender_block)
+        recommender.built = True
         return recommender
 
 
--- a/tests/tf/blocks/core/test_index.py
+++ b/tests/tf/blocks/core/test_index.py
@@ -16,7 +16,7 @@
 
 import pytest
 
-import merlin.models.tf as ml
+import merlin.models.tf as mm
 from merlin.io.dataset import Dataset
 from merlin.schema import Tags
 
@@ -26,8 +26,8 @@ def test_topk_index(ecommerce_data: Dataset):
 
     from merlin.models.tf.metrics.evaluation import ItemCoverageAt, PopularityBiasAt
 
-    model: ml.RetrievalModel = ml.TwoTowerModel(
-        ecommerce_data.schema, query_tower=ml.MLPBlock([64, 128])
+    model: mm.RetrievalModel = mm.TwoTowerModel(
+        ecommerce_data.schema, query_tower=mm.MLPBlock([64, 128])
     )
     model.compile(run_eagerly=False, optimizer="adam")
     model.fit(ecommerce_data, epochs=1, batch_size=50)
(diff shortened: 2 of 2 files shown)


gabrielspmoreira (approved):
Just tried the test case for issue #389 with this branch and it fixes the problem, as now the manual recall using topk recommender matches the evaluation recall

PR #402, opened 2022-04-27
Fixes error with fit() with validation data

### LINKED ISSUE #364: [BUG] Model.fit() with validation_data arg is failing after the Retrieval models evaluation refactory
### Bug description
After this #352 the following error occurs.
When I run `model.fit()` setting `validation_data` argument, it will call `model.evaluate()` in the end of each epoch. Then `RetrievalModel.compute_loss_metrics()` is called and sets [here](https://github.com/NVIDIA-Merlin/models/blob/tf/retrieval-eval/merlin/models/tf/models/base.py#L596) `kwargs["eval_sampling"] = True`, so that in the `ItemRetrievalScorer.call()` the `eval_sampling` is `True`.
But the eval_sampling is not being passed through for some reason, which causes an error in `ItemRetrievalScorer` when using model.fit(..., validation_data=...)

gabrielspmoreira:
- Closes #364 
- Changes `filter_kwargs()` to support forwarding kwargs for functions that have **kwargs

### DIFF
--- a/merlin/models/tf/blocks/core/combinators.py
+++ b/merlin/models/tf/blocks/core/combinators.py
@@ -209,19 +209,19 @@ class SequentialBlock(Block):
 
         outputs = inputs
         for i, layer in enumerate(self.layers):
-            if i == len(self.layers) - 1:
-                filtered_kwargs = filter_kwargs(kwargs, layer, filter_positional_or_keyword=False)
-                filtered_kwargs.update(
-                    filter_kwargs(
-                        {**dict(training=training), **kwargs},
-                        layer.call,
-                        filter_positional_or_keyword=False,
-                    )
+            filtered_kwargs = filter_kwargs(
+                {**dict(training=training), **kwargs},
+                layer,
+                cascade_kwargs_if_possible=True,
+            )
+            filtered_kwargs.update(
+                filter_kwargs(
+                    {**dict(training=training), **kwargs},
+                    layer.call,
+                    cascade_kwargs_if_possible=True,
                 )
-            else:
-                filtered_kwargs = filter_kwargs(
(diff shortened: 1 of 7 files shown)


merlin/models/tf/blocks/core/combinators.py
```diff
@@ -209,19 +209,19 @@ def call(self, inputs, training=False, **kwargs):
...
-                        {**dict(training=training), **kwargs},
-                        layer.call,
-                        filter_positional_or_keyword=False,
-                    )
-                )
-            else:
-                filtered_kwargs = filter_kwargs(
-                    dict(training=training), layer, filter_positional_or_keyword=False
+            filtered_kwargs = filter_kwargs(
+                {**dict(training=training), **kwargs},
```
marcromeyn, line 213:
```suggestion
                {training=training, **kwargs},
```

merlin/models/tf/blocks/core/combinators.py
```diff
@@ -209,19 +209,19 @@ def call(self, inputs, training=False, **kwargs):
...
+                layer,
+                cascade_kwargs_if_possible=True,
+            )
+            filtered_kwargs.update(
+                filter_kwargs(
+                    {**dict(training=training), **kwargs},
```
marcromeyn, line 219:
```suggestion
                    {training=training, **kwargs},
```

marcromeyn (approved):
Looks good, just left some small comments

merlin/models/tf/blocks/core/combinators.py
gabrielspmoreira: Thanks, just added your suggestion in my latest commit

PR #368, opened 2022-04-14
Prototype of revised Merlin Models API

marcromeyn:
Fixes # (issue)

PR #403, opened 2022-04-27
Making evaluation of retrieval models work differently during fit(validation_data) and evaluate(item_corpus)

### LINKED ISSUE #373: [BUG] model.evaluate() gives much higher acc value for small batch size when we set `validation_data` int the model.fit()
### Bug description

When we train Two-tower model and run  model.evaluate()  afterwards, we observe very different  ndcg value wrt different batch size set in the `model.evaluate()`. if you remove `validation_data` from `model.fit()` we dont observe high ndcg value wrt different batch size. 

Please run the `05-Retrieval-Model` notebook and  add the following line after  model.fit() training step. Play with the batch_size. Set it to `4096` as well to see the difference in the `ndcg` value.

```
metrics = model.evaluate(valid, batch_size=128, item_corpus=train, return_dict=True)
```

### Expected behavior
We should not get a big gap in ndcg value when we change the batch size in model.evaluate().

### Environment details
- Merlin version:  
- Platform:
- Python version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?): Yes.

`merlin-tensorflow-training:22.04` container with the latest main branch pulled.

gabrielspmoreira:
This is a draft PR trying to fix #373 . The issue root cause is that  `model.evaluate()` is called by `model.fit()` when `validation_data` argument is passed, but can be also be called by the user explicitly. 
For `RetrievalModel`, the evaluation is done differently during `model.fit()` and `model.evaluate()` calls. During fit, we use the `ItemRetrievalScorer` to sample negative items for evaluation as we do for training (so that we can compare train and validation metrics, but during evaluation we set a `pre_eval_topk` with a `TopKIndexBlock` which scores the top-k items and return them and the candidate samples. 
The issue happens in graph mode, when calling `model.fit(validation_data=...)` and then `model.evaluate(item_corpus=...)`. The reason is that model.evaluate() is going to be called first by `model.fit()` when `pre_eval_topk` is not set yet. Then, when `model.evaluate(item_corpus=...)` is explicitly called by the user, it is going to use the sampled negatives instead of `TopKIndexBlock`. That is why in issue #373 the accuracy matrix vary according to the eval batch size (because when using in-batch negative sampling, the larger the batch sizes, the more negative samples are provided)

### DIFF
--- a/merlin/models/tf/blocks/core/combinators.py
+++ b/merlin/models/tf/blocks/core/combinators.py
@@ -209,19 +209,19 @@ class SequentialBlock(Block):
 
         outputs = inputs
         for i, layer in enumerate(self.layers):
-            if i == len(self.layers) - 1:
-                filtered_kwargs = filter_kwargs(kwargs, layer, filter_positional_or_keyword=False)
-                filtered_kwargs.update(
-                    filter_kwargs(
-                        {**dict(training=training), **kwargs},
-                        layer.call,
-                        filter_positional_or_keyword=False,
-                    )
+            filtered_kwargs = filter_kwargs(
+                {**dict(training=training), **kwargs},
+                layer,
+                cascade_kwargs_if_possible=True,
+            )
+            filtered_kwargs.update(
+                filter_kwargs(
+                    {**dict(training=training), **kwargs},
+                    layer.call,
+                    cascade_kwargs_if_possible=True,
                 )
-            else:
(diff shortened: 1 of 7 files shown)


gabrielspmoreira:
Closing this PR as it would be hacky, and we decided to wait for PR #368 to be merged to fix the issue. See my comments on issue #373

PR #430, opened 2022-05-12
Creating generic model_test

marcromeyn:
### Goals :soccer:
Right now there’s a lot of variation from test to test how models are being tested. This PR is a first step towards standardizing this. 

In the `model_test` we validate if the model can be trained, evaluated & re-trained.

Note that this PR also moves the model tests from `test_core` to `models/test_base`. This is something we forgot to do when we broke up core.py into many smaller files.


marcromeyn:
rerun tests

karlhigley (approved):

Issue #132, opened 2022-01-14
XGBoost Integration


viswa-nvidia:
https://docs.google.com/presentation/d/1jxj9zjeRRu1BJSf8tzaWoQVr5h7HZbOUsSh39Rcwv80/edit#slide=id.g10be2c57ddf\_7\_2

Aha! Link: https://nvaiinfa.aha.io/features/MERLIN-673

radekosmulski:
Hey! I am not sure if this is the right place to post this, @bschifferer asked me to work on an example for the `xgboost` integration and I left some thoughts in a PR #522. Not sure on the process side of things, where things should go or if what I share there can be helpful, if I am doing something wrong or these thoughts are not adding value, please let me know 🙂Thank you!

radekosmulski:
I am not sure if I should have linked my PR here in the first place, but the three of us (@bschifferer, @oliverholworthy, and me) just had a great chat, I also now know what would be the way to capture all this feedback (via feature requests in Merlin Models), so essentially all is well 🙂

PR #433, opened 2022-05-12
First sketch of XGBoost integration with Merlin Dataset

oliverholworthy:
First draft of XGBoost integration with a Merlin Dataset.

Relates to #132 (issue)

### Goals :soccer:

Make it easy to use a Merlin Dataset with XGBoost

Some details to figure out:

- [X] Should the target objective Tag (e.g. Tags.REGRESSION) be specified in constructor or automatically figured out from schema.
  - changed so that the XGBoost objective is passed in and we figure out the target tag from that.
- [X]  What should the default params be for XGBoost and how should customisation work with this wrapper?
   - Allowing all params supported by `train` function to be passed in to the constructor 
- [X] find ranking objectives examples and see if this can be extended to work with learning to rank style models.
  - added support for `rank:*` metrics. Defaults to grouping by the User ID column. Otherwise, the group column(s) can be passed into the fit method. The data will be sorted when the merlin dataset is converted to the DMatrix.
- [X] what do we do with list-like columns in dataset.
  - ignoring list columns and logging a warning when found
- [X] check that categorical features don't need any special treatment with XGBoost here.
  - [As of XGBoost 1.6, the feature is experimental and has limited features](https://xgboost.readthedocs.io/en/latest/tutorials/categorical.html). For the `category` dtype.

#### Example Usage

```python
from merlin.datasets.entertainment import get_movielens
train, valid = get_movielens(variant="ml-1m")

model = XGBoost(
    objective="binary:logistic",
    eval_metric=["logloss", "auc"],
    tree_method="gpu_hist",
)
model.fit(train)
model.evaluate(valid)
# => {'logloss': 0.56554, 'auc': 0.767453}
```


merlin/models/xgb/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+    df = dataset.to_ddf().compute(scheduler="synchronous")
+
+    all_target_columns = dataset.schema.select_by_tag(Tags.TARGET).column_names
+
+    # Ignore list-like columns from schema
+    list_column_names = [
+        col_name
+        for col_name, col_schema in dataset.schema.column_schemas.items()
+        if col_schema.is_list
+    ]
```
marcromeyn, line 115: Should we log a warning here that these columns are being ignored?

merlin/models/xgb/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+
+    X = _to_numpy(df.drop(all_target_columns + list_column_names, axis=1))
+    y = _to_numpy(df[target_columns])
+
+    return xgb.DMatrix(X, label=y)
+
+
+def normalize_regession_targets(dataset: Dataset) -> Dataset:
```
marcromeyn, line 123: What’s the reasoning behind this function? Shouldn’t GBDT models be able to handle non-normalized features/targets?

merlin/models/xgb/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+
+        # Train an XGBoost model
+        from merlin.schema import Tags
+        from merlin.models.xgb import XGBoost
+
+        model = XGBoost(Tags.REGRESSION)
+        model.fit(train)
+    """
+
+    def __init__(self, target_type: Tags, *args, **kwargs):
```
marcromeyn, line 30: I think it could be more generic if we would allow to user to pass in the objective here. A user could use the `get_objective` utililty to infer this from the target-type.

merlin/models/xgb/__init__.py
```diff
@@ -0,0 +1,133 @@
...
+        pass
+
+
+OBJECTIVES = {
+    Tags.BINARY_CLASSIFICATION: "binary:logistic",
+    Tags.REGRESSION: "reg:logistic",
+}
+
+
+def get_objective(target_type: Tags) -> str:
```
marcromeyn, line 75: `target_type` could be both a Tags type or a string.

merlin/models/xgb/__init__.py
```diff
@@ -0,0 +1,145 @@
...
+        constructor as the labels.  And all other non-list columns as
+        input features.
+        """
+        objective = self.objective
+        target_tag = get_target_tag(objective)
+        self.target_columns = get_targets(train, target_tag)
+
+        # if the target is a regression, normalize values
+        if objective == "reg:logistic":
+            train = normalize_regession_targets(train)
```
marcromeyn, line 53: Do we want to do this by default? This would also mean we would re-normalize if the targets are already between 0 and 1.

karlhigley (approved):

merlin/models/xgb/__init__.py
oliverholworthy: It works with non-normalised targets if you use the objective as `reg:logistic` XGBoost raises an error:

```
XGBoostError: ...: label must be in [0,1] for logistic regression
```

I think it's fair enough that the user should be required to transform their data if necessary. I've removed this function `normalize_regession_targets`.

merlin/models/xgb/__init__.py
oliverholworthy: warning now logged if list columns are ignored

merlin/models/xgb/__init__.py
oliverholworthy: XGBoost objective now in constructor instead of the tag

merlin/models/xgb/__init__.py
oliverholworthy: Removed

marcromeyn:
For the xgboost pytest marker, you should register it [here](https://github.com/NVIDIA-Merlin/models/blob/main/pyproject.toml#L60).

PR #435, opened 2022-05-12
Add common release-drafter configuration

mikemckiernan:
Introduce release-drafter consistently
across the organization repositories.
This change should enable us to produce
release notes.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,28 @@
+categories:
+- title: '⚠ Breaking Changes'
+  labels:
+  - 'breaking'
+- title: '🐜 Bug Fixes'
+  labels:
+  - 'bug'
+- title: '🚀 Features'
+  labels:
+  - 'feature'
+  - 'enhancement'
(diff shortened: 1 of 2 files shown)


benfred (approved):

benfred:
I see the pre-commit hook is complaining about the yaml formatting - I think we might need to tweak this file https://github.com/NVIDIA-Merlin/models/blob/d169d335f790abf23039e0c0dd6b22479000c666/.pre-commit-config.yaml#L32 (so instead of `.github/workflows` just have `.github`)

mikemckiernan:
> I see the pre-commit hook is complaining about the yaml formatting - I think we might need to tweak this file
> 
> https://github.com/NVIDIA-Merlin/models/blob/d169d335f790abf23039e0c0dd6b22479000c666/.pre-commit-config.yaml#L32
> 
> (so instead of `.github/workflows` just have `.github`)

I made the tweak though I'm stammering to explain how I made it past pre-commit in my clone.

PR #445, opened 2022-05-18
Move tests to tests/unit or tests/integration 

jperez999:
This moves tests to either tests/unit or tests/integration - to be consistent with other repo's in merlin like merlin-core, nvtabular and merlin-systems.


jperez999:
rerun tests

jperez999:
rerun tests

benfred (approved):

ci/test_integration.sh
```diff
@@ -17,15 +17,4 @@
...
-# Get latest models version
-cd /models/
-git pull origin main
-
-container=$1
-
-## Tensorflow training container
-if [ "$container" == "merlin-tensorflow-training" ]; then
-    ## TODO: Remove this when it would become part of the container
-    pip3 install testbook
```
benfred, line 29: do we have testbook and jupyterlab  on the container now ?

ci/test_integration.sh
```diff
@@ -17,15 +17,4 @@
 #!/bin/bash
 set -e
 
-# Get latest models version
-cd /models/
-git pull origin main
```
benfred, line 22: good catch on removing the 'git pull' - I thought we had got rid of all these

ci/test_integration.sh
jperez999: no they don't currently.

PR #452, opened 2022-05-18
Fixing the inconsistency when exporting embeddings to cudf DataFrame

### LINKED ISSUE #406: [BUG] Exporting embedding tables to parquet does not match the TF variable weights
### Bug description
When we export the embedding table of a categorical feature to a dataframe (`embedding_table_df()`), dataset (`embedding_table_dataset()`) or parquet file (`export_embedding_table()`) the values in the exported dataframe/parquet do not match the values of the original feature embedding weighs.

1. Disable skipping the test `test_embedding_features_exporting_and_loading_pretrained_initializer()` in `tests\tf\features\test_embeddings.py`
2. That test will fail because `EmbeddingFeatures.embedding_table_dataset()` uses under the hood `cudf.from_dlpack()`, which is not working as expected we reported in this issue [#10754](https://github.com/rapidsai/cudf/issues/10754) of `cudf` repo.

### Expected behavior
The values of the exported embedding table should match the embedding table weights.

gabrielspmoreira:
Closes #406 

## Context
I noticed that `cudf.from_dlpack()` is not compatible with TF `to_dlpack()` and opened this [issue](https://github.com/rapidsai/cudf/issues/10754) in RAPIDS cudf repo. They said that `cudf.from_dlpack()` expects Fortran-order (column-major) tensors, which are not compatible with TF tensors and then they placed a [PR](https://github.com/rapidsai/cudf/pull/10850) that adds some checks in `cudf.from_dlpack()`  to raise an exception if that is the case.

## Solution
So I changed the strategy and now use DLPack to convert between from TF to cuPy tensor, and then create a cuDF dataframe from the cuPy tensor

### DIFF
--- a/merlin/models/tf/features/embedding.py
+++ b/merlin/models/tf/features/embedding.py
@@ -311,8 +311,15 @@ class EmbeddingFeatures(TabularBlock):
         embeddings = self.get_embedding_table(table_name, l2_normalization)
         if gpu:
             import cudf
+            import cupy
 
-            df = cudf.from_dlpack(to_dlpack(tf.convert_to_tensor(embeddings)))
+            # Note: It is not possible to convert Tensorflow tensors to the cudf dataframe
+            # directly using dlPack (as the example commented below) because cudf.from_dlpack()
+            # expects the 2D tensor to be in Fortran order (column-major), which is not
+            # supported by TF (https://github.com/rapidsai/cudf/issues/10754).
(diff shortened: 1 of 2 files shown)


gabrielspmoreira:
rerun tests

gabrielspmoreira:
rerun tests

gabrielspmoreira:
rerun tests

karlhigley (approved):

PR #454, opened 2022-05-19
Tf/ragged tensors

marcromeyn:
### Goals :soccer:
When adding features to the `FeatureContext` we convert each list-feature to dense. This makes the memory footprint larger than is necesairy.

This PR adds all list-features to the `FeatureContext`as `tf.RaggedTensor`. Afterwards this can be converted to dense or sparse when needed.


marcromeyn:
rerun tests

karlhigley:
Question: If we returned `RaggedTensors` from the data loaders, would that help?

marcromeyn:
My reasoning to keep it as part of the model is for prediction. When the features come from a feature-store I thought it's easier to retrieve it as tuple and then do the conversion to ragged-tensor in the model.

karlhigley:
I think you're right that we might also want to be able to convert to `RaggedTensor` in the model for ease of use at serving time, but remember those changes to Tensorflow you and I worked on way back to make sequence-based dataloaders support `RaggedTensors`? Those got merged and pushed out in Tensorflow 2.8, but we've never gotten around to circling back to the dataloaders to make the corresponding updates that take advantage of that new functionality.

PR #464, opened 2022-05-26
Fixes and improves masking (for both Causal LM and Masked LM) and adds SequenceAggregation.MASKED_MEAN (used by YouTubeDNN)

gabrielspmoreira:
- Fixes a bug that was masking the inputs incorrectly. Instead of masking the input features at positions of the selected targets, it was doing the opposite, keeping only the input features for the target position and leading to overfitting (on YouTubeDNN for example).
- This PR separates the `label_mask` from the `input_mask` The `label_mask` is used to select the labels to be predicted and `input_mask` is used to replace input features for positions of the selected labels when necessary.
- Added `SequenceAggregation.MASKED_MEAN`, which allows aggregating a tensor considering only masked values. This is useful for example to average (3D) sequences of embeddings considering only non-padded positions (useful for YouTubeDNNModel for example)

### DIFF
--- a/merlin/models/tf/blocks/core/masking.py
+++ b/merlin/models/tf/blocks/core/masking.py
@@ -193,7 +193,13 @@ class CausalLanguageModeling(MaskingBlock):
             mask_inputs = mask_labels
 
         else:
-            mask_labels = item_ids != self.padding_idx
+            labels = item_ids[:, 1:]
+            # pad shifted sequence to original length
+            labels = tf.concat(
+                [labels, tf.zeros((tf.shape(item_ids)[0], 1), dtype=labels.dtype)],
+                axis=-1,
+            )
+            mask_labels = labels != self.padding_idx
             mask_inputs = None
 
         return mask_labels, mask_inputs
--- a/merlin/models/tf/blocks/core/masking.py
+++ b/merlin/models/tf/blocks/core/masking.py
@@ -88,14 +88,6 @@ class MaskingBlock(Block):
             dtype=tf.float32,
         )
 
-        self.label_seq_trg_eval = tf.Variable(
-            tf.zeros(shape=[1, input_shapes[1]], dtype=tf.int64),
-            name="target_labels",
-            dtype=tf.int64,
-            trainable=False,
-            shape=tf.TensorShape([None, input_shapes[1]]),
-        )
-
         super().build(input_shapes)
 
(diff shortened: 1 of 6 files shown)


merlin/models/tf/blocks/core/masking.py
```diff
@@ -328,18 +339,12 @@ def compute_feature_mask(self, item_ids: tf.Tensor, training: bool = False) -> t
...
-            labels = item_ids[:, 1:]
-            labels = tf.concat(
-                [
-                    labels,
-                    tf.zeros((tf.shape(labels)[0], 1), dtype=labels.dtype),
-                ],
-                axis=-1,
-            )
-            mask_labels = labels != self.padding_idx
+            mask_labels = item_ids != self.padding_idx
```
sararb, line 344: I guess we will still need to shift the sequence of item ids as the first element won't be used as `target` (we don't have any prior information to feed to the model and getting predictions)

sararb (commented):
Thank you for the catch! We had a similar observation in T4rec (TensorFlow API) where the model was overfitting (see open [issue](https://github.com/NVIDIA-Merlin/Transformers4Rec/issues/333)). 

I have just left one comment regarding the case where we evaluate the model for all items instead of last-item only.

merlin/models/tf/blocks/core/masking.py
gabrielspmoreira: Hey Sara. Yes, shifting labels for next-click prediction with CLM is still necessary. I just removed the [shifting](https://github.com/NVIDIA-Merlin/models/pull/464/files/9a73b2fbef8bbfcd5ac5e70f4498b745b6794cc3#diff-b5dafb33a21eb7e088c3576769b9bc2c184742a8f5096dbad501304786a22f1eL187-L193) here because it was not necessary, as this method returns the label mask and not the label values

sararb: Thanks for the answer! I guess it would impact the definition of targets [here](https://github.com/NVIDIA-Merlin/models/blob/6fc41b65e26c50a336053258ba602293d8a9e9d3/merlin/models/tf/blocks/core/masking.py#L379). As the current implementation of MaskingHead uses the masking tensor stored in the context to retrieve targets?

gabrielspmoreira:
rerun tests

gabrielspmoreira:
rerun tests

PR #466, opened 2022-05-27
XGBoost - Switch to dask API

oliverholworthy:
### Goals :soccer:

Since the Merlin Datasets use dask natively. Using the XGBoost Dask API allows us to pass the underlying dask dataframe without needing to call `.compute` on the dataframe.

- Refactored `fit` method. Moving the optional `target_columns` and `qid_column` to the constructor.
- Added new required argument `schema` to the constructor
- Added `booster` argument so that a class can be constructed from a loaded booster. `predict` can then be called without needing to call `fit` first.

### Testing Details :mag:

- Added a test for the case where a dask client is not initialised. We get a ValueError in this case.

### Example Usage

```python
from merlin.datasets.entertainment import get_movielens
from merlin.core.utils import Distributed

train, valid = get_movielens(variant="ml-1m")

with Distributed():
    model = XGBoost(train.schema, objective="binary:logistic")
    model.fit(train)
    metrics = model.evaluate(valid)
```

With a ranking objective

```python
from merlin.datasets.entertainment import get_movielens
from merlin.core.utils import Distributed

train, valid = get_movielens(variant="ml-1m")

with Distributed():
    model = XGBoost(train.schema, target_columns=["rating"], qid_column="userId", objective="rank:ndcg", eval_metric="ndcg")
    model.fit(train)
    metrics = model.evaluate(valid)
```


tests/conftest.py
```diff
@@ -60,6 +62,12 @@ def testing_data() -> Dataset:
     return data
 
 
+@pytest.fixture(scope="module")
+def dask_client() -> distributed.Client:
+    with Distributed(cluster_type="cpu") as dist:
```
oliverholworthy, line 67: we can remove the cluster_type here if the the [gpu detection logic (currently relying on a numba import) is updated to something else](https://github.com/NVIDIA-Merlin/core/blob/8a23dc89b6b6b195670e1fea4162f506847c056b/merlin/core/utils.py#L33-L36)

benfred: Whats the issue with the current GPU detection logic? Can we use the HAS_GPU flag from merlin.core.dispatch ?

https://github.com/NVIDIA-Merlin/core/blob/8a23dc89b6b6b195670e1fea4162f506847c056b/merlin/core/dispatch.py#L43-L45

oliverholworthy: I've created a PR on core to try to make the automatic GPU detection logic inside this `merlin.core.utils.Distributeed` class more reliable here: https://github.com/NVIDIA-Merlin/core/pull/98

rnyak:
rerun tests

merlin/models/xgb/__init__.py
```diff
@@ -164,28 +193,29 @@ def get_target_tag(objective: str) -> Tags:
...
-def dataset_to_dmatrix(
-    dataset: Dataset, target_columns: Union[str, list], qid_column: Optional[str]
-) -> xgb.DMatrix:
+def dataset_to_xy(
+    dataset: Dataset,
+    target_columns: Union[str, list],
+    qid_column: Optional[str],
+):
     """Convert Merlin Dataset to XGBoost DMatrix"""
-    df = dataset.to_ddf().compute()
```
oliverholworthy, line 188: Switching to use dask, avoids the need to call `compute` on the dataset here

PR #472, opened 2022-05-31
Use standard Keras embedding layers inside `EmbeddingFeatures` block

karlhigley:
### Goals :soccer:
This PR aims to incorporate the standard Keras embedding layer into the `EmbeddingFeatures` class, since other embedding implementations (e.g. `distributed-embeddings`) adhere to the Keras embedding layer API

Instead of having the `ModelContext` own the embedding tables, we've kept ownership of the embedding tables in the `EmbeddingFeatures` class and added a reference to the embedding tables to the `ModelContext`. (It was the other way previously, which was kind of confusing.)

### Testing Details :mag:
Covered by existing tests.


rnyak:
@bschifferer can you take charge of validating the perf with/without this change?

marcromeyn (approved):
I got some time to properly parse this PR. @karlhigley is completely right, this PR won’t introduce any performance regressions. This because the keras `Embedding` layer is used to create the embedding-table, but we still use a custom function to do look-ups. Our custom code is in the lookup, not in the table-creation.

We might want to standardize to also use the keras `Embedding`-layer to do lookups, but since this can actually introduce performance regressions we need to first have a way to track these. We can make this part of future work around optimizing embedding-lookups.

marcromeyn:
rerun tests

gabrielspmoreira:
I have tested the throughtput (average examples / sec) of this branch (just rebased to main branch) compared to main branch by using the [retrieval experiments script](https://github.com/NVIDIA-Merlin/research/blob/main/retrieval_exp/scripts/retrieval_train_eval.py) on NGC.

### Results
- There was no difference in the loss. 
- It seems the avg throughtput reduced just by ~1% (from 41869 -> 41480)

image

fyi @karlhigley @marcromeyn @EvenOldridge @benfred

rnyak:
rerun tests

karlhigley:
This is the first step in a longer path than I'm prepared to walk. If someone else wants to revive it, feel free, but I'm going to set this aside for now.

karlhigley:
I'm not even fundamentally convinced this change is a good idea. It _seemed_ like a good idea while we were trying to make YouTubeDNN into a `RetrievalModel`, but it turns out that change can be accomplished without this. I really think we'd be better off starting fresh on whatever rework y'all decide to do for the input layer, but make whatever decision works for y'all here. As it stands, this seems like a performance regression without a purpose to me at this point.

gabrielspmoreira:
I see @karlhigley . The small 1% drop in performance might also be due to some variation in the instance hardware made available by NGC for the test. @marcromeyn as you are working on a flexible design for users defining different embedding tables, I will defer to you the decision on whether this PR is aligned with the design you have in mind

PR #473, opened 2022-05-31
Make `YoutubeDNNRetrieval` inherit from `RetrievalModel`

karlhigley:
### Goals :soccer:
YoutubeDNNRetrieval is the only retrieval model that doesn't inherit from `RetrievalModel` and therefore doesn't have embedding export functionality. This PR reworks the model so it does inherit from `RetrevalModel`.

We construct `YoutubeDNNRetrieval` as a two-tower model where the item tower is just an `InputBlock` for the item embedding table. We subsequently ignore the output of the item tower (as before) and use only the output from the user tower. Item embeddings are still shared with the sampled softmax at the end through the `ModelContext`. 

Since we now have two prediction outputs (one from each tower), we've updated `RemovePad3D` and `ItemRetrievalScorer` to accept either a `tf.Tensor` (as before) or a dictionary of `tf.Tensors`.

### Testing Details :mag:
Covered by existing tests.

### DIFF
--- a/merlin/models/tf/blocks/core/transformations.py
+++ b/merlin/models/tf/blocks/core/transformations.py
@@ -387,16 +387,26 @@ class RemovePad3D(Block):
         non_pad_mask = targets != self.padding_idx
         targets = tf.boolean_mask(targets, non_pad_mask)
 
-        assert isinstance(predictions, tf.Tensor), "Predictions must be a tensor"
+        if isinstance(predictions, dict):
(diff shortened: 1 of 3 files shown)


gabrielspmoreira:
@marcromeyn you have mentioned you would analyse the PR to check if it matches with the refactory you have been making on retrieval models. How should we proceed with this PR?

karlhigley:
Closing since this now conflicts with more recent changes in MM

PR #474, opened 2022-05-31
fix make file to run unit tests

jperez999:


karlhigley (approved):

marcromeyn:
I am a bit confused about the need of this PR. I was under the impression that: `"tensorflow and not (integration or example)”` would take care of ignoring the integration-tests? We automatically mark the integration tests as `integration` using pytest-markers [here](https://github.com/NVIDIA-Merlin/models/blob/main/tests/conftest.py#L85-L86). Is this not working?

karlhigley:
Yeah, the integration tests were running on all PRs for a bit. Not sure if that's still the case though.

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

oliverholworthy:
Looks like the markers are not correctly configured for the integration tests.

I've created a change here to simplify the marker creation and fix the integration test marking. #510

benfred:
@jperez999 I don't think this is necessary after @oliverholworthy's fix in #510

PR #478, opened 2022-06-01
Propagate schema to branches if needed in ParallelBlock

marcromeyn:
### Goals :soccer:
This PR fixes a bug where if you create a `ParallelBlock` with a schema, the schema needs to be propagated to all branches.


merlin/models/tf/blocks/core/combinators.py
```diff
@@ -379,6 +379,11 @@ def __init__(
                 f"dictionaries of layer. got: {inputs}"
             )
 
+        if schema:
+            for branch in self.parallel_values:
+                if not getattr(branch, "has_schema", True):
```
oliverholworthy, line 384: If it's safe to assume the `branch` here is a `Block` (which has  `set_schema`, and `has_schema` from the SchemaMixin). Can we call `.has_schema` without `getattr` here?

marcromeyn: Yeah I thought about this as well. My reasoning was that this code is more generic (using duck-typing) without being that much less readable. Curious to your thoughts!

oliverholworthy (approved):

PR #479, opened 2022-06-01
Quick fix to simplify DLRM when there are no continuous features

marcromeyn:
### Goals :soccer:
This PR adds a quick-fix to our `DLRMBlock`, when there are no continuous-features we don’t need the shortcut connection.


marcromeyn:
rerun tests

PR #481, opened 2022-06-01
Add tests for calling filter block when output returns empty

oliverholworthy:
### Goals :soccer:

Improve testing of Filter block. 

Add test for calling filter block when output returns empty dict.


tests/unit/tf/blocks/core/test_tabular.py
```diff
@@ -1,28 +1,42 @@
-import merlin.models.tf as ml
```
oliverholworthy, line 1: i understand that `mm` is now the preferred import alias, and when we're changing imports as we go.

tests/unit/tf/blocks/core/test_tabular.py
```diff
@@ -1,28 +1,42 @@
...
+    assert mm.Filter([])({}) == {}
+    assert mm.Filter([])(tf_con_features) == {}
+    assert mm.Filter(["unknown"])(tf_con_features) == {}
+
+    assert mm.Filter(Schema())({}) == {}
+    assert mm.Filter(Schema())(tf_con_features) == {}
+    assert mm.Filter(Schema(["unknown"]))(tf_con_features) == {}
+
+    assert mm.Filter(Tags.CONTINUOUS)({}) == {}
+    assert mm.Filter(Tags.CONTINUOUS).set_schema(Schema(["unknown"]))(tf_con_features) == {}
```
oliverholworthy, line 22: This tag version is a bit different to the others because passing a Tag to the constructor doesn't provide enough information to figure out the column names.

tests/unit/tf/blocks/core/test_tabular.py
rnyak: Yes, we go with `mm`.

gabrielspmoreira (approved):

PR #483, opened 2022-06-01
Remove single-use abstractions `RetrievalBlock` and `RetrievalMixin`

karlhigley:
### Goals :soccer:
It's possible we'll want to have a way to share code across retrieval models in the future, but we're currently only using these abstractions in one place each, which means that they're an unnecessary layer of indirection right now.

Since `DualEncoderBlock` and `RetrievalMixin` were the only classes that matched the `RetrievalProtocol` and `DualEncoderBlock` was the only class using `RetrievalMixin`, we can consolidate references to all three to use `DualEncoderBlock` directly and remove the other two.

### Testing Details :mag:
Covered by existing tests.

### DIFF
--- a/merlin/models/tf/blocks/retrieval/base.py
+++ b/merlin/models/tf/blocks/retrieval/base.py
@@ -50,16 +50,6 @@ class TowerBlock(ModelBlock):
     pass
 
 
-class RetrievalMixin:
-    def query_block(self) -> TowerBlock:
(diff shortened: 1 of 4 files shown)


karlhigley:
This causes circular import issues, since most everything in the `tf` sub-package is exported from `tf/__init__.py`. That might not be a great way to do it, but changing that is beyond the scope of this PR.

PR #488, opened 2022-06-02
Fixes errors on retrieval models evaluation in graph mode

gabrielspmoreira:
After the recent refactory, errors like below one were being raised during evaluation on retrieval models because the targets 
had shape undefined. That caused errors when using pairwise losses (e.g. bpr-max) and for custom metrics that use `tf.boolean_mask()` and require the mask tensor to have at most one dim undefined.
```
y_pred_valid_rows = tf.boolean_mask(y_pred, valid_rows_with_positive_mask)

    ValueError: Number of mask dimensions must be specified, even if some dimensions are None.  E.g. shape=[None] is ok, but shape=None is not.
```


gabrielspmoreira:
rerun tests

gabrielspmoreira:
rerun tests

benfred (approved):

PR #489, opened 2022-06-03
Add local installtion instructions to the README

radekosmulski:
I believe that for many of our potential users using docker might be problematic.

Many even very good data scientists don't have docker in their toolbelt. Other people (including very successful folks) are very opinionated about their setups. Plus they might want to integrate Merlin tools with other libraries of their choice that do not come preinstalled in our containers (or are not at the version they want to use).

Additionally, developing against docker containers is not a very enjoyable experience (for instance, I couldn't get `pip -e .[all]` with a mounted directory to work with Merlin Models, which made it harder to figure out how things work, in the end I decided to install it locally). 

Not to mention that if people will not know how to do this, they won't be able to run our code on Kaggle/collab/paperspace/easily on GCP&AWS (Kaggle is especially amazing for fostering adoption).

Just wanted to open sharing directions for local installation up for discussion 🙂  (this might seem a no brainer that you can figure out how to install stuff locally via looking at dockerfiles but most people wouldn't make the connection due to lack of experience with docker plus they might not know where to find our dockerfiles)

(BTW have just gotten `nvtabular` to work on kaggle 🥳 just had to add the below)
```
sudo apt-get update
sudo apt-get install -y libopenmpi-dev
```

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,14 +58,9 @@ pip install merlin-models
 
 ```
 sudo apt update -y --fix-missing
-sudo apt install -y --no-install-recommends libexpat1-dev libsasl2-2 libssl-dev graphviz openssl protobuf-compiler software-properties-common
-sudo apt autoremove -y
-sudo apt clean
-sudo rm -rf /var/lib/apt/lists/*
+sudo apt install -y --no-install-recommends libexpat1-dev libsasl2-2 libssl-dev graphviz openssl software-properties-common
 
 pip install betterproto graphviz pybind11 pydot pytest mpi4py transformers==4.12
-pip install --upgrade notebook
-pip install --upgrade ipython
 pip install nvidia-pyindex
 pip install tritonclient[all] grpcio-channelz
 pip install numba==0.55.1
--- a/README.md
+++ b/README.md
@@ -57,7 +57,6 @@ pip install merlin-models
 > The following script can be used for installing dependencies locally, however due to the intricacies of individual setups, we are only able to help you troubleshoot any issues you might encounter when running Merlin Models inside one of the provided docker containers.
 
 ```
-sudo apt update -y --fix-missing
 sudo apt install -y --no-install-recommends libexpat1-dev libsasl2-2 libssl-dev graphviz openssl software-properties-common
 
 pip install betterproto graphviz pybind11 pydot pytest mpi4py transformers==4.12
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@ pip install merlin-models
 
 > Installing Merlin Models with `pip` does not install some additional GPU dependencies, such as the CUDA Toolkit.
 > It is best to use the highly optimized Docker containers with dependencies installed.
-> The following script can be used for installing dependencies locally, however due to the intricacies of individual setups, we are only able to help you troubleshoot any issues you might encounter when running merlin-models inside one of the provided docker containers.
+> The following script can be used for installing dependencies locally, however due to the intricacies of individual setups, we are only able to help you troubleshoot any issues you might encounter when running Merlin Models inside one of the provided docker containers.
 
 ```
 sudo apt update -y --fix-missing
@@ -73,7 +73,7 @@ pip install git+https://github.com/rapidsai/asvdb.git@main
 pip install merlin-models
 ```
 
-You can consult the following [docker file](https://github.com/NVIDIA-Merlin/Merlin/tree/main/docker) for further details on the most up to date steps on installing dependencies.
+Please consult the Dockerfiles [here](https://github.com/NVIDIA-Merlin/Merlin/tree/main/docker) for further details on the most up to date steps for installing dependencies.
 
 #### Docker Containers that include Merlin Models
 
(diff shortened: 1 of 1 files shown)


benfred:
rerun tests

README.md
```diff
@@ -53,7 +53,27 @@ pip install merlin-models
 ```
 
 > Installing Merlin Models with `pip` does not install some additional GPU dependencies, such as the CUDA Toolkit.
-> When you run Merlin Models in one of our Docker containers, the dependencies are already installed.
+> It is best to use the highly optimized Docker containers with dependencies installed.
+> The following script can be used for installing dependencies locally, however due to the intricacies of individual setups, we are only able to help you troubleshoot any issues you might encounter when running Merlin Models inside one of the provided docker containers.
+
+```
+sudo apt update -y --fix-missing
+sudo apt install -y --no-install-recommends libexpat1-dev libsasl2-2 libssl-dev graphviz openssl protobuf-compiler software-properties-common
```
oliverholworthy, line 61: For simplicity we might wish to keep this as minimal as possible. I think most of this script is not a dependency of merlin models or required for the installation or use of the library. 

The main one I can see is the `protobuf-compiler` which is required for the nvtabular installation.

And nvtabular is only installed currently if you request it with `pip install merlin-models[nvtabular]`.

rnyak:
@radekosmulski did you get a chance to test if Merlin Models run on CPU as well (particularly, dataloaders)?

radekosmulski:
@rnyak didn't have a chance to! could I please ask what is the "proper" way of trying to run merlin models on a CPU? I guess I can just hide my GPU using CUDA_VISIBLE_DEVICES 🙂 would this be the way to go about this?

Will give it a shot and report here how it worked

rnyak:
> @rnyak didn't have a chance to! could I please ask what is the "proper" way of trying to run merlin models on a CPU? I guess I can just hide my GPU using CUDA_VISIBLE_DEVICES 🙂 would this be the way to go about this?
> 
> Will give it a shot and report here how it worked

@radekosmulski you need to set device to `cpu` in the dataloader, AFAIK. Pls  check `BatchedDataset` class [arguments](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/dataset.py#L262)

README.md
```diff
@@ -53,7 +53,21 @@ pip install merlin-models
...
 
 > Installing Merlin Models with `pip` does not install some additional GPU dependencies, such as the CUDA Toolkit.
-> When you run Merlin Models in one of our Docker containers, the dependencies are already installed.
+> It is best to use the highly optimized Docker containers with dependencies installed.
+> The following script can be used for installing dependencies locally, however due to the intricacies of individual setups, we are only able to help you troubleshoot any issues you might encounter when running Merlin Models inside one of the provided docker containers.
+
+```
+sudo apt install -y --no-install-recommends libexpat1-dev libsasl2-2 libssl-dev graphviz openssl software-properties-common
+
+pip install betterproto graphviz pybind11 pydot pytest mpi4py transformers==4.12
```
rnyak, line 62: looks like this installs graphviz two times. in the above line you install it as well.

radekosmulski:
I think my thinking was sound in hoping to provide a way for users to install merlin models (and ideally all our other libraries) locally, but this pr doesn't make sense...

I experimented with this a bit more now, to understand better what could be pruned from what I was installing there originally. Apparently, these installation instructions were allowing me to install merlin models in a way where it ran on the GPU given what I already had in my environment...

I played with this a bit more in a fresh ubuntu docker container and I don't think I can figure out what dependencies are needed. Unfortunately, upon taking a closer look at our Dockerfiles, this PR is not very useful.

I think I now am starting to understand a bit better the struggles associated with creating containers for releases.

My current understanding is that I wouldn't be able to collect the list of dependencies here and that in practice all this will be very dependent on what the user already has on their machine.

I would like to somehow point the user in the right direction, that they can review our dockerfiles and look for hints on what they need to install if they want to run merlin models locally on the GPU, but that is not very elegant either and I don't think it is very approachable to people who can't or are unwilling to run docker in the first place.

Let me please kill this PR and think on what else -- if anything -- could be done here.

oliverholworthy:
One option might be to suggest using a  conda environment to get the required RAPIDS dependencies setup.

The following is a Dockerfile I've tried that works with the getting started example (Tensorflow + DLRM):

```Dockerfile
FROM nvidia/cuda:11.7.0-base-ubuntu22.04

# Conda + RAPIDS
RUN apt-get update
RUN apt-get install -y wget
RUN wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh
RUN bash Miniconda3-latest-Linux-x86_64.sh -b
ENV PATH=/root/miniconda3/bin:${PATH}
RUN conda install -c rapidsai -c nvidia -c conda-forge \
    rapids=22.04 python=3.9 cudatoolkit=11.7 cudnn

# Merlin Models
RUN apt-get install -y build-essential protobuf-compiler  # required for nvtabular pip install
RUN pip install merlin-models[nvtabular,tensorflow]
```

PR #492, opened 2022-06-03
Fix dataloader docstring

benfred:
Docstrings were referring to 'sparse_dense' instead of 'sparse_as_dense'
which is the actual parameter name. Fix and also fix an issue with
formatting with the tf dataloader class docstring.


jperez999 (approved):

benfred:
rerun tests

PR #493, opened 2022-06-04
Fix to pass shuffle argument to dataloader

benfred:
The 'shuffle' argument wasn't being passed to the dataloader in _maybe_convert_merlin_dataset,
and was instead just silently being ignored. This meant that it wasn't possible to
set 'shuffle=False' . Fix this, and also change the behaviour for evaluate/predict to
not shuffle the dataset - since it shouldn't ever be necessary, and in the case of predict
could return confusing results


marcromeyn:
rerun tests

benfred:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

oliverholworthy (approved):

PR #502, opened 2022-06-08
Add Wide and Deep model into benchmark

timmy00:
### Goals :soccer:
Replaced by #623 

Add Wide & Deep Model into benchmark

Arguments of wide_schema and deep_schema are used to specify features for wide model and for deep model, for now it is only suitable for binary classification, which is similar to [tf.estimator.DNNLinearCombinedClassifier](https://www.tensorflow.org/api_docs/python/tf/estimator/DNNLinearCombinedClassifier)

### Testing Details :mag:
Passed the unit test defined in test_benchmark.py


merlin/models/tf/models/benchmark.py
```diff
@@ -94,3 +98,107 @@ def NCFModel(
...
+    prediction_tasks = parse_prediction_tasks(schema, prediction_tasks)
+
+    if schema is None:
+        raise ValueError("The schema is required by Wide and Deep Model")
+
+    if embedding_dim is None:
+        raise ValueError("The embedding_dim is required")
+
+    if not wide_schema:
+        warnings.warn("If not specify wide_schema, all features would be sent to wide " "model")
```
timmy00, line 163: I am not sure if it is correct to add a warning like this.

merlin/models/tf/models/benchmark.py
```diff
@@ -94,3 +98,107 @@ def NCFModel(
...
+            "The deep_block is required by Deep & Wide Model when "
+            "features are available in the deep_schema"
+        )
+
+    if len(deep_schema) > 0:
+        if not deep_input_block:
+            deep_input_block = InputBlock(
+                deep_schema,
+                embedding_options=EmbeddingOptions(
+                    embedding_dim_default=embedding_dim, **embedding_option_kwargs
```
sararb, line 179: Fixing the embedding dimension here will define the same dimension for all input features. However, some high-cardinality features (like item-ids )would require a high embedding dimension while low-cardinality (in the order of 5 or 10 possible values) would require a smaller dimension. 

I wonder if it would be more efficient to infer the dimension of the embeddings from the schema based on the feature cardinality and then project them into the final `embedding_dim` using an MLP layer? The motivation would be to reduce the model's parameters while ensuring the same performance level. So if it makes sense to you, this would require to run some training experiments to compare between the two configurations.

merlin/models/tf/models/benchmark.py
```diff
@@ -94,3 +98,107 @@ def NCFModel(
...
+                ),
+                **kwargs,
+            )
+        deep_body = deep_input_block.connect(deep_block).connect(
```
sararb, line 183: The `connect` method accepts a series of blocks and links them sequentially, so you could rewrite the deep_body as: 

`deep_input_block.connect(deep_block, MLPBlock(dimensions=[1], no_activation_last_layer=True))`

merlin/models/tf/models/benchmark.py
```diff
@@ -94,3 +98,107 @@ def NCFModel(
...
+
+    if embedding_dim is None:
+        raise ValueError("The embedding_dim is required")
+
+    if not wide_schema:
+        warnings.warn("If not specify wide_schema, all features would be sent to wide " "model")
+        wide_schema = schema
+
+    if not deep_schema:
+        deep_schema = schema
```
sararb, line 167: Would it make sense to allow users to define and train only the Wide model?

merlin/models/tf/models/benchmark.py
```diff
@@ -94,3 +98,107 @@ def NCFModel(
...
+    Model
+
+    """
+
+    prediction_tasks = parse_prediction_tasks(schema, prediction_tasks)
+
+    if schema is None:
+        raise ValueError("The schema is required by Wide and Deep Model")
+
+    if embedding_dim is None:
```
sararb, line 159: The `embedding_dim` is a required argument in your function so I think you could ignore this check as the `call` will fail if `embedding_dim` is not provided.

sararb (commented):
@timmy00, thank you for the PR! The overall definition of the model seems good to me. I've left a few comments for discussion, I'm curious to hear what you think? 

It would be great if you could also add unit tests to check for cases raising ValueError (like missing deep_block). Here is an [example](https://github.com/NVIDIA-Merlin/models/blob/87b5d07cabb097e0dde1613ae0e881dd7a84aa38/tests/unit/tf/blocks/test_dlrm.py#L55) of unit test that checks if the ValueError was correctly raised. 

I also wonder if we should keep the benchmark.py file that was originally created for including recsys architectures that we will not recommend for use due to their poor performance. We could also add them as part of the API (in ranking.py or retrieval.py) and have separate benchmark examples to demonstrate their underperformance.   What do you think, @EvenOldridge @benfred?

merlin/models/tf/models/benchmark.py
timmy00: Sure, thank you for pointing out

merlin/models/tf/models/benchmark.py
timmy00: I think if users only want to train a wide model, they would simply choose a dense layer or a linear layer, no need to use wide and deep model.

merlin/models/tf/models/benchmark.py
timmy00: sure, will do. Thank you

timmy00:
> It would be great if you could also add unit tests to check for cases raising ValueError (like missing deep_block). Here is an [example](https://github.com/NVIDIA-Merlin/models/blob/87b5d07cabb097e0dde1613ae0e881dd7a84aa38/tests/unit/tf/blocks/test_dlrm.py#L55) of unit test that checks if the ValueError was correctly raised.

It seems that I do not have to check required arguments like this:
`if schema is None:
        raise ValueError("The schema is required by Wide and Deep Model")`

I make both schema and deep_block become required arguments, so if they are not specified, there would be original type error:
image

sararb:
> > It would be great if you could also add unit tests to check for cases raising ValueError (like missing deep_block). Here is an [example](https://github.com/NVIDIA-Merlin/models/blob/87b5d07cabb097e0dde1613ae0e881dd7a84aa38/tests/unit/tf/blocks/test_dlrm.py#L55) of unit test that checks if the ValueError was correctly raised.
> 
> It seems that I do not have to check required arguments like this:
> 
> ```
> if schema is None:
>        raise ValueError("The schema is required by Wide and Deep Model")
> ```
> 
> I make both schema and deep_block become required arguments, so if they are not specified, there would be original type error: image

You're right! This is not needed anymore, I was suggesting such a test when `deep_block` was not required.

merlin/models/tf/models/benchmark.py
timmy00: Thank you for your suggestion, I will add this into my implementation.

merlin/models/tf/models/benchmark.py
```diff
@@ -94,3 +99,125 @@ def NCFModel(
...
+            )
+        deep_body = deep_input_block.connect(
+            deep_block, MLPBlock(dimensions=[1], no_activation_last_layer=True)
+        )
+
+    if len(wide_schema) > 0:
+        if not wide_input_block:
+            wide_input_block = InputBlock(
+                wide_schema,
+                post=AsSparseFeatures(),
```
timmy00, line 212: No sure if using AsSparseFeatures is correct, here I want to transfer categorical features into sparse matrix, I refer from the source code of `class EmbeddingFeatures(TabularBlock)`, which uses this bock as pre block, but inside `AsSparseFeatures`, it calls `tf.RaggedTensor.from_row_lengths(values, row_lengths).to_sparse()`, so I am not sure if category feature column in dataset can be transformed by `tf.RaggedTensor.from_row_lengths(values, row_lengths).to_sparse()`.

merlin/models/tf/models/benchmark.py
gabrielspmoreira: I think feature interactions should be done in the model side (and not in preprocessing side), for being more flexible for a Data Scientist to experiment with different feature combinations without the need of preprocessing a new version of the dataset.  
I agree that the feature interactions of the wide part of W&D could be defined by using a `CrossFeature` block, so that it is reusable by other models which include a linear model with feature interaction.
That was the design choice of TF ([tf.feature_column.crossed_column](https://www.tensorflow.org/api_docs/python/tf/feature_column/crossed_column)).
I know Sara's PR #301 has a prototype of a `CrossFeatures` block which leverages `tf.feature_column.crossed_column` that can be useful.

merlin/models/tf/models/benchmark.py
timmy00: Got it, thank you Gabriel!

merlin/models/tf/models/benchmark.py
timmy00: Created [HashedCross ](https://github.com/NVIDIA-Merlin/models/pull/587)and [HashedCrossAll](https://github.com/NVIDIA-Merlin/models/pull/590)

gabrielspmoreira:
Hey @timmy00 . It seems your PR is including others commit, which makes it hard to review it.
Would you like to try rebasing your PR locally with `main` and then using `git push -f origin wide_deep_model_test` to see if it keeps only your PR commits here?

timmy00:
Refactor in #623 , close this one.

PR #508, opened 2022-06-13
add usecase with pretrained embeddings

### LINKED ISSUE #421: [Task] Example for using pre-trained embeddings
### Description
We want to add an example how to use pre-trained embeddings:

Multiple options
- pre-train item embeddings with word2vec/product2vec
- dataset with images and extract latent vector with pre-trained cnn model
- process text

**Additional context**
- related issues: https://github.com/NVIDIA-Merlin/Merlin/issues/211

radekosmulski:
Following the great advice from @rnyak (thank you!!!) I was able to create a use case around using pretrained embeddings.

I use `TensorInitializer` to provide the model with pretrained embeddings for the movies.
```
model = mm.DCNModel(
train.schema,
    depth=2,
    deep_block=mm.MLPBlock([64, 32]),
    prediction_tasks=mm.BinaryClassificationTask(target_column),
    embedding_options=mm.EmbeddingOptions(
        embeddings_initializers={
            "movieId": mm.TensorInitializer(pretrained_movie_embs),
        }
    )
)
```

But while this works, I am not sure that this is what this functionality should look like (I am using preexisting pieces that probably were developed for some other purpose).

The full-blown scenario IMHO should look a little bit more like this:

 - you still can train embeddings from scratch for each of your categorical variables
 - but you can also specify any number of pretrained embedding matrices for any of your categorical features

For instance, in an extreme scenario where you have 3 different shots for each product, you should be able to train a regular embedding for that product, and then feed in each of the 3 pretrained image embeddings for each item.

Something like this
```
    embedding_options=mm.EmbeddingOptions(
        embedding_dims={'product_id': 128}
        embeddings_initializers={
            "product_id_shot_A": mm.TensorInitializer(pretrained_product_id_embs_A),
            "product_id_shot_B": mm.TensorInitializer(pretrained_product_id_embs_B),
            "product_id_shot_C": mm.TensorInitializer(pretrained_product_id_embs_C),
        }
    )
```

where the model would still be able to tie each of the embeddings to the `product_id` (to know which row to fetch for a given example).

This is by all means not finished, just probably another step in the conversation.

resolves #421

### DIFF
--- a/examples/usecases/entertainment-with-pretrained-embeddings.ipynb
+++ b/examples/usecases/entertainment-with-pretrained-embeddings.ipynb
@@ -1,17 +1,59 @@
 {
  "cells": [
   {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "92d580b6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Copyright 2022 NVIDIA Corporation. All Rights Reserved.\n",
+    "#\n",
+    "# Licensed under the Apache License, Version 2.0 (the \"License\");\n",
+    "# you may not use this file except in compliance with the License.\n",
+    "# You may obtain a copy of the License at\n",
+    "#\n",
+    "#     http://www.apache.org/licenses/LICENSE-2.0\n",
+    "#\n",
+    "# Unless required by applicable law or agreed to in writing, software\n",
+    "# distributed under the License is distributed on an \"AS IS\" BASIS,\n",
+    "# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.\n",
+    "# See the License for the specific language governing permissions anda\n",
+    "# limitations under the License.\n",
+    "# =============================================================================="
+   ]
+  },
+  {
    "cell_type": "markdown",
    "id": "697d1452",
    "metadata": {},
    "source": [
+    "<img src=\"http://developer.download.nvidia.com/compute/machine-learning/frameworks/nvidia_logo.png\" style=\"width: 90px; float: right;\">\n",
+    "\n",
+    "# Training with pretrained embeddings\n",
+    "\n",
+    "## Overview\n",
+    "\n",
     "In this use case we will consider how we might train with pretrained embeddings.\n",
     "\n",
     "Pretrained embeddings can allow our model to include information from additional modalities (for instance, we might want to grab CNN descriptors of product images). They can also come from other models that we train on our data. For example, we might train a word2vec model on the sequence of purchased items by a customer and want to include this information in our retrieval or ranking model.\n",
     "\n",
     "The use cases are many, but this particular example will focus on the technical aspects of working with pretrained embeddings.\n",
     "\n",
-    "We will use the MovieLens 100k dataset and emulate a scenario where we would have a pretrained embedding for each of the movies in the train dataset."
+    "We will use the MovieLens 100k dataset and emulate a scenario where we would have a pretrained embedding for each of the movies in the train dataset.\n",
+    "\n",
+    "### Learning objectives\n",
+    "\n",
+    "- Training with pretrained embeddings\n",
+    "- Understanding [the Schema file](https://github.com/NVIDIA-Merlin/core/blob/main/merlin/schema/schema.py)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "21edc9aa",
+   "metadata": {},
+   "source": [
+    "## Downloading and preparing the dataset"
    ]
   },
   {
@@ -508,6 +550,14 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "id": "c913b600",
+   "metadata": {},
+   "source": [
+    "## Building the model"
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": 8,
    "id": "22af1945",
@@ -533,6 +583,14 @@
     "        }\n",
     "    )\n",
     ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3b5e6631",
+   "metadata": {},
+   "source": [
+    "## Training"
    ]
   },
   {
--- a/examples/usecases/entertainment-with-pretrained-embeddings.ipynb
+++ b/examples/usecases/entertainment-with-pretrained-embeddings.ipynb
@@ -1,0 +1,408 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "3e97432c",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "/workspace/examples/usecases\r\n"
+     ]
+    }
+   ],
+   "source": [
+    "!pwd"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "028dc96f",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import os"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "120df1a2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "os.chdir('/workspace')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "697d1452",
+   "metadata": {},
+   "source": [
+    "In this use case we will consider how we might train with pretrained embeddings.\n",
+    "\n",
+    "Pretrained embeddings can allow our model to include information from other modalities (for instance, we might want to grab CNN descriptors of product images). They can also come from other models that we train on our data. For example, we might train a word2vec model on the sequence of purchased items by a customer and want to include this information in our retrieval or ranking model.\n",
+    "\n",
+    "The use cases are many, but this particular example will focus just on the technical aspects of working with pretrained embeddings.\n",
+    "\n",
+    "We will use the MovieLens 100k dataset and emulate a scenario where we would have a pretrained embedding for each of the movies in the train dataset."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 4,
+   "id": "e8c63b03",
+   "metadata": {},
+   "outputs": [
+    {
+     "name": "stderr",
+     "output_type": "stream",
+     "text": [
+      "2022-06-13 04:02:20.693184: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:20.693584: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:20.693744: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:20.715031: I tensorflow/core/platform/cpu_feature_guard.cc:152] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  SSE3 SSE4.1 SSE4.2 AVX\n",
+      "To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.\n",
+      "2022-06-13 04:02:20.715735: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:20.715924: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:20.716074: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:21.398757: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:21.398961: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
+      "2022-06-13 04:02:21.399109: I tensorflow/stream_executor/cuda/cuda_gpu_executor.cc:952] successful NUMA node read from SysFS had negative value (-1), but there must be at least one NUMA node, so returning NUMA node zero\n",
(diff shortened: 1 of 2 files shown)


rnyak:
@radekosmulski thanks for the PR. Arent we already capable of doing this?
```
    embedding_options=mm.EmbeddingOptions(
        embedding_dims={'product_id': 128}
        embeddings_initializers={
            "product_id_shot_A": mm.TensorInitializer(pretrained_product_id_embs_A),
            "product_id_shot_B": mm.TensorInitializer(pretrained_product_id_embs_B),
            "product_id_shot_C": mm.TensorInitializer(pretrained_product_id_embs_C),
        }
    )
```
dont we have such functionality right now? if not, what's missing can you please identify? thanks.

radekosmulski:
@rnyak I am not fully sure if I am understanding the functionality right but if I right now do the following:

```
model = mm.DCNModel(
train.schema,
    depth=2,
    deep_block=mm.MLPBlock([64, 32]),
    prediction_tasks=mm.BinaryClassificationTask(target_column),
    embedding_options=mm.EmbeddingOptions(
        embeddings_initializers={
            "movieId": mm.TensorInitializer(pretrained_movie_embs),
        }
    )
)
```

whatever embeddings would be automatically created for my `movieId` are replaced with the embedding matrix I create via `mm.TensorInitializer`. The `movieId` ids map to the indices in the newly created embedding matrix.

If I do the following:
```
    embedding_options=mm.EmbeddingOptions(
        embedding_dims={'product_id': 128}
        embeddings_initializers={
            "product_id_shot_A": mm.TensorInitializer(pretrained_product_id_embs_A),
            "product_id_shot_B": mm.TensorInitializer(pretrained_product_id_embs_B),
            "product_id_shot_C": mm.TensorInitializer(pretrained_product_id_embs_C),
        }
    )
```
there is no `product_id_shot_A` index in my dataset. I would like the embedding matrix to map to `movieId` index. But I don't think I can do the following either:

```
    embedding_options=mm.EmbeddingOptions(
        embedding_dims={'product_id': 128}
        embeddings_initializers={
            "movieId": mm.TensorInitializer(pretrained_product_id_embs_A),
            "movieId": mm.TensorInitializer(pretrained_product_id_embs_B),
            "movieId": mm.TensorInitializer(pretrained_product_id_embs_C),
        }
    )
    ```

or can I? 🤔

Will try that out right now 😄

**Comment by radekosmulski** [comment]
Just tried that

```
pretrained_movie_embs_A = np.random.random((1682, 64))
pretrained_movie_embs_B = np.random.random((1682, 64))

model = mm.DCNModel(
train.schema,
    depth=2,
    deep_block=mm.MLPBlock([64, 32]),
    prediction_tasks=mm.BinaryClassificationTask(target_column),
    embedding_options=mm.EmbeddingOptions(
        embeddings_initializers={
            "movieId": mm.TensorInitializer(pretrained_movie_embs_A),
            "movieId": mm.TensorInitializer(pretrained_movie_embs_B),
        }
    )
)
```
and I believe only a single embedding matrix can be associated with a `movieId`. We can either have the "default" embeddings for `movieId`, like for any other categorical variable, or we can have the pretrained embeddings that we provide, but we cannot have both or multiple pretrained embeddings.

This would be a super useful feature to have (and would allow us to train multimodal models).

I am thinking that this example might be good to go as is, it demonstrates the use of pretrained embeddings. There is also a workaround to this I believe, by creating duplicated `movieId` columns, but it would be really elegant if we could tie the same column ID to embeddings coming from different embedding tables

Maybe we could merge this use case and continue the conversation in NVIDIA-Merlin/Merlin#211?

**Comment by rnyak** [comment]
> Just tried that
> 
> ```
> pretrained_movie_embs_A = np.random.random((1682, 64))
> pretrained_movie_embs_B = np.random.random((1682, 64))
> 
> model = mm.DCNModel(
> train.schema,
>     depth=2,
>     deep_block=mm.MLPBlock([64, 32]),
>     prediction_tasks=mm.BinaryClassificationTask(target_column),
>     embedding_options=mm.EmbeddingOptions(
>         embeddings_initializers={
>             "movieId": mm.TensorInitializer(pretrained_movie_embs_A),
>             "movieId": mm.TensorInitializer(pretrained_movie_embs_B),
>         }
>     )
> )
> ```
> 
> and I believe only a single embedding matrix can be associated with a `movieId`. We can either have the "default" embeddings for `movieId`, like for any other categorical variable, or we can have the pretrained embeddings that we provide, but we cannot have both or multiple pretrained embeddings.
> 
> This would be a super useful feature to have (and would allow us to train multimodal models).
> 
> I am thinking that this example might be good to go as is, it demonstrates the use of pretrained embeddings. There is also a workaround to this I believe, by creating duplicated `movieId` columns, but it would be really elegant if we could tie the same column ID to embeddings coming from different embedding tables
> 
> Maybe we could merge this use case and continue the conversation in [NVIDIA-Merlin/Merlin#211](https://github.com/NVIDIA-Merlin/Merlin/issues/211)?

I think it is expected to have one embedding table per input column.  What I meant, if you have say image embeddings per movie_id, you can create an extra column in the input dataframe by copying the `movieId` column, and name it to something like `movieId_image` and then add that column name to TensorInitializer like below. Is that not working?

```

> pretrained_movie_embs_B = np.random.random((1682, 64))
> 
> model = mm.DCNModel(
> train.schema,
>     depth=2,
>     deep_block=mm.MLPBlock([64, 32]),
>     prediction_tasks=mm.BinaryClassificationTask(target_column),
>     embedding_options=mm.EmbeddingOptions(
>         embedding_dims={'movieId': 128},
           embeddings_initializers={
>             "movieId_image": mm.TensorInitializer(pretrained_movie_embs_B),
>         }
>     )
> )
> ```

radekosmulski:
It works but I was not sure if this is the solution we want? That was my entire point.

If we are happy with this approach, of duplicating the ID column, then that's perfect. We have everything we need here 🙂

And I am thinking you might be right -- this might be the preferred way to do this, as opposed to probably significantly increasing code complexity just to address this in a more elegant fashion (via being able to tie multiple embeddings to the same ID column)

rnyak:
@gabrielspmoreira can you provide your input pls? thanks.

rnyak:
> It works but I was not sure if this is the solution we want? That was my entire point.
> 
> If we are happy with this approach, of duplicating the ID column, then that's perfect. We have everything we need here 🙂
> 
> And I am thinking you might be right -- this might be the preferred way to do this, as opposed to probably significantly increasing code complexity just to address this in a more elegant fashion (via being able to tie multiple embeddings to the same ID column)

Let's loop Gabriel in and get his opinion as well.

gabrielspmoreira:
Yes @radekosmulski . Currently there is a many-to-one mapping between categorical features and embedding table. Meaning that one embedding table can be shared by multiple categorical features in the schema.pbtxt but not the opposite.
In this example snippet for a `schema.pbtxt` file you have the  categ features `clicked_item` and `previously_clicked_item` sharing the same embedding table `item_id`.

```
feature {
  name: "clicked_item"
  type: INT
  int_domain {
    name: "item_id"
    max: 3370
    is_categorical: true
  }
}
feature {
  name: "previously_clicked_item"
  type: INT
  int_domain {
    name: "item_id"
    max: 3370
    is_categorical: true
  }
}
```

Allowing many-to-many relationship between categ features and embedding tables would add complexity to the schema, requiring changes in both NVTabular and Merlin Models.

gabrielspmoreira:
Another possibility would be "cloning" features in the model side. For example, we could have a `CloneFeatures` block that clones a set of features with new names. For example:

```python
clone_features_block = CloneFeatures(cloning_dict={"item_id": ["Item_product_image", "item_product_test"]})
new_schema = clone_features_block.get_schema_with_cloned_features(schema)
```

Then you could define DLRM model using the low-level API like in this [example](https://github.com/NVIDIA-Merlin/models/blob/main/examples/06-Define-your-own-architecture-with-Merlin-Models.ipynb), so that you can create your own embedding_block like this, setting the `clone_features_block` as `pre` and providing the new schema which includes the cloned features.

```python
embedding_block = mm.EmbeddingFeatures.from_schema(new_schema, pre=clone_features_block)
```

Here is a prototype of how such `CloneFeatures` block could look like
```python
class CloneFeatures(TabularBlock):

    def __init__(
        self, cloning_dict, name=None, **kwargs
    ):
        self.cloning_dict = cloning_dict
        super().__init__(name=name, **kwargs)

    def call(self, inputs: TabularData, **kwargs) -> TabularData:
        outputs = inputs
        for feat_name in inputs:
            if feat_name in self.cloning_dict:
                new_feature_names = self.cloning_dict[feat_name]
                for new_feat_name in new_feature_names:
                    outputs[new_feat_name] = inputs[feat_name]

        return outputs

    def compute_call_output_shape(self, input_shape):
        output_shape = {k: v for k, v in input_shape.items()}

        for original_feat_name, new_feat_names in self.cloning_dict.items():
            for new_feat_name in new_feat_names:
                output_shape[new_feat_name] = output_shape[original_feat_name]

        return output_shape

    def get_config(self):
        config = super().get_config()
        config["cloning_dict"] = self.feature_names

        return config

    def get_schema_with_cloned_features(self, schema):
        new_column_schemas = []
        for column_schema in schema:
            if column_schema.name in self.cloning_dict:
                new_feature_names = self.cloning_dict[column_schema.name]
                for new_feat_name in new_feature_names:
                    new_col_schema = copy.deepcopy(column_schema)
                    new_col_schema.name = new_feat_name
                    new_column_schemas.append(new_col_schema)

        new_schema = Schema([schema] + new_column_schemas)
        return new_schema
```

Would that design make sense @marcromeyn @sararb ?

gabrielspmoreira:
In such cases, you might prefer setting the pre-trained embedding to `trainable=False`. Currently we don't have in the API an option to freeze weights of embedding tables. 
But I know that @marcromeyn is working in a refactory that will make it more flexible to set embedding table options. What are your thoughs on this @marcromeyn ?

rnyak:
> Yes @radekosmulski and @rnyak . Currently there is a many-to-one mapping between categorical features and embedding table. Meaning that one embedding table can be shared by multiple categorical features in the schema.pbtxt but not the opposite. In this example snippet for a `schema.pbtxt` file you have the categ features `clicked_item` and `previously_clicked_item` sharing the same embedding table `item_id`.
> 
> ```
> feature {
>   name: "clicked_item"
>   type: INT
>   int_domain {
>     name: "item_id"
>     max: 3370
>     is_categorical: true
>   }
> }
> feature {
>   name: "previously_clicked_item"
>   type: INT
>   int_domain {
>     name: "item_id"
>     max: 3370
>     is_categorical: true
>   }
> }
> ```
> 
> Allowing many-to-many relationship between categ features and embedding tables would add complexity to the schema, requiring changes in both NVTabular and Merlin Models.

that's a good point.

examples/usecases/entertainment-with-pretrained-embeddings.ipynb
```diff
@@ -0,0 +1,408 @@
+{
```
bschifferer, line 1: <p>I am not sure if we can assuem that workspace for every user. This will likely cause troubles. Can we avoid setting os.chdir?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/508/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='2'/>

bschifferer, line 1: <p>You should be able to read the information <code>1682</code> (how many movieIds are avalable) from the schema file.</p><p> </p><p><code>train.schema</code> </p><p> </p><p>Then we do not need to read the output of the NVTabular workflow. This could create issues when paths are not consistent with preprocessing and where the notebook is stored.</p><p> </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/508/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

gabrielspmoreira:
> > Yes @radekosmulski and @rnyak . Currently there is a many-to-one mapping between categorical features and embedding table. Meaning that one embedding table can be shared by multiple categorical features in the schema.pbtxt but not the opposite. In this example snippet for a `schema.pbtxt` file you have the categ features `clicked_item` and `previously_clicked_item` sharing the same embedding table `item_id`.
> > ```
> > feature {
> >   name: "clicked_item"
> >   type: INT
> >   int_domain {
> >     name: "item_id"
> >     max: 3370
> >     is_categorical: true
> >   }
> > }
> > feature {
> >   name: "previously_clicked_item"
> >   type: INT
> >   int_domain {
> >     name: "item_id"
> >     max: 3370
> >     is_categorical: true
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Allowing many-to-many relationship between categ features and embedding tables would add complexity to the schema, requiring changes in both NVTabular and Merlin Models.
> 
> that's a good point. why we need to tag `item_product_image` as `item_id` in the schema? isnt it an item feature like `brand` and `category`, so why not only tagging it as `item`. May be I am missing something :)

You are right @rnyak . Only the item id feature should be tagged with `item_id` Tag

examples/usecases/entertainment-with-pretrained-embeddings.ipynb
radekosmulski: <p>Good point on using paths!</p><p> </p><p>Here I don't seem to be able to find this information in the schema. It only contains info on cardinality but not the maximum possible value.</p><p> </p><p>I switched to grabbing the path from the schema, please let me know if that would be okay</p><p> </p><p><code></code><code>movieIds = cudf.read_parquet(train.schema.column_schemas['movieId'].properties['cat_path'])</code><code></code></p>

radekosmulski: good point, removed!

radekosmulski: saw your comment below, switched to synthetic data where this no longer is an issue 🙂

radekosmulski: yup, removed this

radekosmulski:
Seems the CI is not too happy about the label, I wonder what should we do here? I can create the `examples` label or would we rather align the CI with the scheme that we have there currently?

radekosmulski:
@bschifferer it is a test regarding a label on the PR that is failing:
image

I think it might not like the label `area/examples`, not sure what would be the best way to fix it (change the label to documentation or amend the test)

I also see that we have another test failing regarding code formatting, will fix this now

radekosmulski:
A really good day of meetings yesterday 🙂 I would like to thank @bschifferer and @oliverholworthy for their invaluable feedback! 🙂 Apart from working on this PR, I spent some time modifying the example for `xgboost` based on Oliver's comments. Seems talking to a person who implemented a particular functionality adds another dimension to what is happening in code and is super useful for me in creating examples 🙂 Might start bothering people more actively going forward for their perspective as I continue to work on these notebooks 😄 Thanks a lot for the help again!

I will push a new version of this example in a few minutes, just wanted to leave one comment before I do so.

Tried to add additional explanations in the areas identified by @bschifferer. One thing I was not able to do though that we discussed was showing how the embedding table changes after training. The blocker here was that the embedding table doesn't seem to exist before calling `fit` on the model (even after `compile`).

image

The embedding matrix should be accessible here:
`model.layers[0].layers[0].layers[0].parallel_layers['categorical'].embedding_tables`
and it indeed is after training.

BTW the fact that all this happens in `fit` is probably a useful consideration for designing how we can give users the ability to freeze these embedding layers.

I felt training for a single epoch and then showing the embedding table, and only then training for more epochs, is a little bit inelegant, hence opted for not including this comparison. But happy to add it if that would be the call on this one!

bschifferer (approved):
It looks good to me. I like the version as a first step.

We can update it with freezing the pretrained embeddings, when the feature is available.

rnyak:
@radekosmulski did you create a bug ticket for `The blocker here was that the embedding table doesn't seem to exist before calling fit on the model (even after compile).` issue? Thanks.

radekosmulski:
@rnyak apologies, didn't realize this would be a bug, thank you for catching this! Raised one now

PR #510, opened 2022-06-13
Fix pytest marker for integration tests.

oliverholworthy:
Follow-up to #474 . The integration test markers were not being set correctly. Looks like they were previously a subdirectory of the unit tests and was moved without updating the marker configuration in conftest.

- Fix markers for integration tests
- Replace `startswith` checks with string in checks to make the markers more robust to directory structure changes.
- Remove unused `tensorflow_example` marker from config in `pyproject.toml`


benfred:
rerun tests

benfred (approved):

PR #511, opened 2022-06-13
Fixes computing metrics each N steps during training and adding regularization_loss metric

### LINKED ISSUE #491: [BUG] Performance regression after the refactory that moves losses and metrics to model.compile()
### Bug description
We observed a strong performance (throughput) regression after the refactory that moved losses and metrics to `model.compile()`.
The *average examples / sec* metric dropped from ~105k to 42k. That can be reproduced using the [retrieval experiments scripts](https://github.com/NVIDIA-Merlin/research/blob/main/retrieval_exp/scripts/retrieval_train_eval.py) with the same arguments, only moving the loss and metrics out of the task to the `model.compile()`.

After using some commits to isolate the issue as suggested by @marcromeyn , it seems the problem is introduced by some commit after this [commit [A]](https://github.com/NVIDIA-Merlin/models/commit/168ad03f8089146647c23b5fbe5fc65b39779ecb) and before (including) this [commit [B]](https://github.com/NVIDIA-Merlin/models/commit/10a3b8132c89991e7d25b0101c8e811654f24458).

Isolating a bit the issue, the lower throughput seems to be related to the metrics computation during training, as I also tested the pipeline (1) after removing all ranking metrics eval during training (Recall, NDCG, MRR) and (2) removing all metrics, including custom ones where I aggregate statistics from the logits.

In the following plot, it is possible to see the throughput at commit A (before the change) in **brown**, the throughput at commit B in **purple**, the throughput after removing ranking metrics (1) in **blue**, and the throughput after removing all but one metric (2) in **green**, which is closer to the original throughput (brown) 

image

So I suspect that some mechanisms might not be working properly after the refactory:
1 - The `train_metrics_steps` arg for `model.fit()` that configures training to compute metric values only each n steps is not working properly after the refactory.
2 Maybe that mechanism that sort predictions and targets only once for all ranking metrics and different cut-offs is not working anymore

More details on these perf tests in this [Slack thread (Nvidia Merlin internal)](https://nvidia.slack.com/archives/C020P24SQJJ/p1654152215076019)

### LINKED ISSUE #495: [BUG] The regularization_loss and total_loss are not reported anymore after moving the loss to model.compile()
### Bug description
After the loss were moved to `model.compile()`, the `regularization_loss` and `total_loss` are not reported anymore.
From the experiments results it is also not clear if the regularization losses are being summed to the final loss or not

It can be seen in the plots below that the `regularization_loss` and `total_loss` were reported before (blue line) but are not being reported after the refactory (pink line) and the loss of the new `run` is worse than before. 

image

The results for the new run can be reproduced using the [retrieval experiments script](https://github.com/NVIDIA-Merlin/research/blob/main/retrieval_exp/scripts/retrieval_train_eval.py), with the following arguments and LastFM:

```bash
python scripts/retrieval_train_eval.py --dataset lastfmB --wandb_exp_group lastfmB_youtubednn_xe_sampledsofmax_v07.2 --model_type youtubednn --two_tower_activation selu --epochs 20 --lr_decay_steps 50 --output_path /results --data_path /data --log_to_wandb --optimizer adam --eval_batch_size 2048 --train_metrics_steps 100 --topk_metrics_cutoffs 100 --max_seq_length 20 --youtubednn_sampled_softmax True --fail_if_recall_at_100_higher_than 0.5 --xe_label_smoothing 0.0 --two_tower_mlp_layers None --two_tower_dropout 0.3 --two_tower_embedding_sizes_multiplier 5.0 --logits_temperature 0.8 --lr 0.02238982864512884 --lr_decay_rate 0.9400000000000001 --l2_reg 1.1472035643715902e-05 --embeddings_l2_reg 1.01774316277423e-06 --train_batch_size 4096 --item_id_emb_size 512 --youtubednn_sampled_softmax_n_candidates 500
```

gabrielspmoreira:
## Training performance improvement by computing metrics each N steps
Fixes #491
Restores the behavior in which metrics are computed during training each N steps (where N is defined by `model.fit(, ...train_metrics_steps=N)`. This leads to much better training performance when using metrics that are slower to compute (like ranking metrics over hundreds of thousand items).

In the following plot we can see the original throughput (avg examples per second) in **orange**, the throughtput after the refactory that moved metrics to `model.compile()` in **gray**, and the throughtput with this fix in **red** (computing metrics during each 100 steps, as did in the original run (orange)). 
image

There is still a gap to the original throughput (107k vs 97k), which I belives derives from the fact that we had before an optimization that sorted predictions only once for all ranking metrics, and now it is sorted for each ranking metric and cut-off. I will work in another PR to include that optimization back and check how much closer to the original throughput we get.

## Adding the `regularization_loss` metric
Fixes #495 
Included logging `regularization_loss` as a metric as before. A difference from the previous behavior was that we used a custom `compute_loss()` method which provided `loss`, `regularization_loss` and `total_loss` (sum of the previous losses) metrics. Now we use standard Keras `Model.compute_loss()` method,  which returns a single total loss whose metric name is `loss` (which corresponds to the previous `total_loss`). So the `loss` now includes the `regularization_loss`.
Also used the opportunity to update/fix the metrics names assertions in unit and integration tests

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -5,6 +5,7 @@ from collections.abc import Sequence as SequenceCollection
 from typing import TYPE_CHECKING, Dict, List, Optional, Protocol, Union, runtime_checkable
 
 import tensorflow as tf
+from keras.utils.losses_utils import cast_losses_to_common_dtype
 from tensorflow.python.keras.engine import data_adapter
 
 import merlin.io
@@ -464,7 +465,9 @@ class Model(tf.keras.Model):
         # Run backwards pass.
         self.optimizer.minimize(loss, self.trainable_variables, tape=tape)
 
-        metrics = self.compute_metrics(x, outputs.targets, outputs.predictions, sample_weight)
+        metrics = self.compute_metrics(
+            x, outputs.targets, outputs.predictions, sample_weight, training=True
+        )
 
         return metrics
 
@@ -480,9 +483,72 @@ class Model(tf.keras.Model):
             outputs = self.pre_eval_topk.call_outputs(outputs)
 
         self.compute_loss(x, outputs.targets, outputs.predictions, sample_weight)
-        metrics = self.compute_metrics(x, outputs.targets, outputs.predictions, sample_weight)
+        metrics = self.compute_metrics(
+            x, outputs.targets, outputs.predictions, sample_weight, training=False
+        )
 
         return metrics
+
+    @tf.function
+    def compute_metrics(
+        self,
+        x: tf.Tensor,
+        y: tf.Tensor,
+        y_pred: tf.Tensor,
+        sample_weight: tf.Tensor,
+        training: bool,
+    ) -> Dict[str, tf.Tensor]:
+        """Overrides default Keras Model.compute_metrics()
+           to compute metrics each N steps (and not for all steps)
+           during training
+
+        Parameters
+        ----------
+        x : tf.Tensor
+            Input features. Not used, and kept only to align with the signature
+            of super Model.compute_metrics()
+        y : tf.Tensor
+            Target values
+        y_pred : tf.Tensor
+            Predictions values
+        sample_weight : tf.Tensor
+            Sample weights for weighted metrics
+        training : bool
+            Flag that indicates if metrics are being computed during
+            training or evaluation
+
+        Returns
+        -------
+        Dict[str, tf.Tensor]
+            Dict with the metrics values
+        """
+        del x
+        # During training updates metric states each N steps
+        if self._should_compute_train_metrics_for_batch or not training:
+            self.compiled_metrics.update_state(y, y_pred, sample_weight)
+        # Returns the current value of metrics
+        metrics = self.metrics_results()
+        # Adding regularization loss
+        metrics["regularization_loss"] = tf.reduce_sum(cast_losses_to_common_dtype(self.losses))
+        return metrics
+
+    def metrics_results(self) -> Dict[str, tf.Tensor]:
+        """Logic to consolidate metrics results
+        extracted from standard Keras Model.compute_metrics()
+
+        Returns
+        -------
+        Dict[str, tf.Tensor]
+            Dict with the metrics values
+        """
+        return_metrics = {}
+        for metric in self.metrics:
+            result = metric.result()
+            if isinstance(result, dict):
+                return_metrics.update(result)
+            else:
+                return_metrics[metric.name] = result
+        return return_metrics
 
     def fit(
         self,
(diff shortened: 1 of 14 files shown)


merlin/models/tf/models/base.py
```diff
@@ -480,10 +485,73 @@ def test_step(self, data):
...
+        training : bool
+            Flag that indicates if metrics are being computed during
+            training or evaluation
+
+        Returns
+        -------
+        Dict[str, tf.Tensor]
+            Dict with the metrics values
+        """
+        del x
```
karlhigley, line 529: Does this actually need to be deleted? Totally possible that it does, but wondering how we know it's safe and necessary to mark it for garbage collection here.

karlhigley (commented):
LGTM so registering a vote for approval, but I'll defer to others on whether to merge it before the release

merlin/models/tf/models/base.py
gabrielspmoreira: Thanks for your review @karlhigley. We have been using standard Keras `Model.compute_metrics()` (after Marc's refactory), which had the same line [here](https://github.com/keras-team/keras/blob/v2.9.0/keras/engine/training.py#L986). I just override that method so that we conditionally update the metrics (each N steps during training or all steps during evaluation) but kept the same signature and inner code.

oliverholworthy (approved):

merlin/models/tf/models/base.py
```diff
@@ -719,15 +787,6 @@ def evaluate(
             **kwargs,
         )
 
-    def compute_loss_metrics(
```
oliverholworthy, line 722: looks like this is an unused function? and unrelated to the issue of throughput?

gabrielspmoreira: Exactly @oliverholworthy . This is dead code that was used before Marc's refactory.

tests/unit/tf/prediction_tasks/test_multi_task.py
```diff
@@ -40,10 +55,34 @@ def test_mmoe_head(music_streaming_data: Dataset):
...
     model.compile(optimizer="adam", run_eagerly=True)
 
-    step = model.train_step(ml.sample_batch(music_streaming_data, batch_size=50))
+    metrics = model.train_step(ml.sample_batch(music_streaming_data, batch_size=50))
 
-    assert step["loss"] >= 0
-    assert len(step) == 13
+    assert metrics["loss"] >= 0
+    assert len(metrics) == 14
+    # TODO: Investigate why metrics names change when multiple tests are run,
```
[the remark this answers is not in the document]
gabrielspmoreira: That is a good point @oliverholworthy. I also observed in the past other tests failing only when executed together with other tests. It seems that TF shares graph state among tests.
I tried your suggestion locally and it seems to work! I gonna place separate PR with your suggestion.

PR #513, opened 2022-06-14
Creating test fixture to reset state of TF Keras graph before each test

gabrielspmoreira:
I have observed that some unit tests that run successfully when executed individually were failing when multiple tests are run.
@oliverholworthy suggested [here](https://github.com/NVIDIA-Merlin/models/pull/511) using this Pytest fixture to clear the session before every test. I tested that approach and was able to uncomment some test asserts that check for metric names of default Keras metric (e.g. auc, precision, recall) which before were given different names (e.g. auc, auc_2, auc_3) depending on tests executed before.

```python
@pytest.fixture(autouse=True)
def tf_clear_session():
    tf.keras.backend.clear_session()
```
<!--

### DIFF
--- a/tests/unit/tf/_conftest.py
+++ b/tests/unit/tf/_conftest.py
@@ -25,6 +25,11 @@ MAX_CARDINALITY = 100
 ASSETS_DIR = pathlib.Path(__file__).parent.parent / "assets"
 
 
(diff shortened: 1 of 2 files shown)


benfred (approved):
looks good! nice suggestion @oliverholworthy

PR #514, opened 2022-06-15
Refactored top-k metrics and created TopKMetricsAggregator for optimized metrics computation

### LINKED ISSUE #491: [BUG] Performance regression after the refactory that moves losses and metrics to model.compile()

gabrielspmoreira:
### Goals :soccer:
This PR improves metrics computation performance (both during training and inference) by sorting predictions only once before computing Top-k metrics. It is an addition improvement on top of #511
It also renames ranking metrics to top-k metrics which is more general (as RecallAt and PrecisionAt are top-k metrics and not ranking metrics).

- The `TopKMetricsAggregator` metric is introduced to aggregate a list of `TopkMetric` and sort predictions only once before computing the top-k metrics.
- I changed the top-k metrics `update_state(self,y_true,y_pred,sample_weight, label_relevant_counts)` to follow the standard Keras signature, by removing the `label_relevant_counts` arg, which cannot be passed to the standard `model.compiled_metrics.update_state()` method we are now using. To be able to pass that info to `TopkMetric` and to `TopKMetricsAggregator` a property`label_relevant_counts` was created for those classes, so that `model.compute_metrics()` can set it for all metrics before calling `update_state()`

### Testing Details :mag:
- Unit tests were updated to reflect the changes
- I ran a performance test on NGC without and with `TopKMetricsAggregator` 
1. Using `model.compile(..., metrics=[RecallAt(100), RecallAt(300), RecallAt(500), MRRAt(100), MRRAt(300), MRRAt(500), NDCGAt(100), NDCGAt(300), NDCGAt(500)])` 
2. Using  `model.compile(..., metrics=[TopKMetricsAggregator([RecallAt(100), RecallAt(300), ...])]`. 

You can see in the following plot that the training throughtput (average examples per second) increases from 40k to 55k when using `TopKMetricsAggregator`. 

image

P.s. This test was performed setting `model.fit(..., train_metrics_steps=1)` which computes metrics for each training step. The throughput of this experiment is improved to 97k (see #511) when we set `train_metrics_steps=100`.

### DIFF
--- a/merlin/models/tf/metrics/topk.py
+++ b/merlin/models/tf/metrics/topk.py
@@ -15,13 +15,13 @@
 #
 
 # Adapted from source code: https://github.com/karlhigley/ranking-metrics-torch
-from typing import List, Optional, Sequence
+from typing import List, Optional, Sequence, Union
 
 import tensorflow as tf
 from keras.utils import losses_utils, metrics_utils
 from keras.utils.tf_utils import is_tensor_or_variable
 from tensorflow.keras import backend
-from tensorflow.keras.metrics import Mean
+from tensorflow.keras.metrics import Mean, Metric
 from tensorflow.keras.metrics import get as get_metric
 
 from merlin.models.tf.metrics import metrics_registry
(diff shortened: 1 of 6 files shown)


merlin/models/tf/metrics/topk.py
```diff
@@ -282,43 +297,126 @@ def from_config(cls, config):
...
-    metrics: List[RankingMetric] = []
+class TopKMetricsAggregator(Metric, TopkMetricWithLabelRelevantCountsMixin):
+    """Aggregator for top-k metrics (TopkMetric) that is optimized
+    to sort top-k predictions only once for all metrics.
+
+    topk_metrics : Sequence[TopkMetric]
+        Multiple arguments with TopkMetric instances
+    """
+
+    def __init__(self, topk_metrics: Sequence[TopkMetric]):
```
marcromeyn, line 343:
```suggestion
    def __init__(self, *topk_metrics: TopkMetric):
```

marcromeyn: This would simplify the API a bit. Since you could do: `TopKMetricsAggregator(RecallAt(5), MRRAt(5))`, instead of having to wrap it into a list.

gabrielspmoreira: Done

merlin/models/tf/metrics/topk.py
```diff
@@ -282,43 +297,126 @@ def from_config(cls, config):
...
+
+    def result(self):
+        outputs = {}
+        for metric in self.topk_metrics:
+            outputs[metric.name] = metric.result()
+
+        return outputs
+
+
+def topk_metrics_default(top_ks: Sequence[int], **kwargs) -> Sequence[TopkMetric]:
```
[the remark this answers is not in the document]
gabrielspmoreira: Makes sense. Moved to a `TopKMetricsAggregator.default_metrics()` classmethod as you have suggested

marcromeyn (approved):

PR #518, opened 2022-06-20
Breaking out BaseModel to support sub-classing

marcromeyn:
### Goals :soccer:
This PR breaks up `Model`, it introduces `BaseModel` that can be used in order to create a model with the sub-class style that Keras supports.

An example of a sub-classed style model can be found in [this](https://github.com/NVIDIA-Merlin/models/blob/tf/base-model/tests/unit/tf/models/test_base.py#L139-L167) test.

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -264,6 +264,15 @@ class BaseModel(tf.keras.Model):
             **kwargs: Arguments supported for backwards compatibility only.
         """
 
+        # Initializing model control flags controlled by MetricsComputeCallback()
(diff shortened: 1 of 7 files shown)


marcromeyn:
rerun tests

merlin/models/tf/models/base.py
```diff
@@ -582,6 +503,10 @@ def fit(
         **kwargs,
     ):
         x = _maybe_convert_merlin_dataset(x, batch_size, **kwargs)
+
+        if isinstance(x, BatchedDataset):
```
oliverholworthy, line 507: Can we add a comment explaining why this is required here.

oliverholworthy (approved):

PR #519, opened 2022-06-20
Add tests for train_metrics_steps ensuring we can control the frequency of computing metrics

oliverholworthy:
### Goals :soccer:

Improve the testing of  `train_metrics_steps` usage to reduce the risk of a regression to this functionality in future.

Followup to #511 

### Testing Details :mag:

Adding a test for different values of the `train_metrics_steps` parameter to `Model.fit` to check that we only compute the metrics every `train_metrics_steps`

### DIFF
--- a/tests/unit/tf/models/test_base.py
+++ b/tests/unit/tf/models/test_base.py
@@ -14,8 +14,10 @@
 # limitations under the License.
 #
 import pytest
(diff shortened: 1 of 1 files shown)


gabrielspmoreira:
@oliverholworthy thanks for putting together this test!

PR #523, opened 2022-06-22
Add hashed cross transformation block

timmy00:
### Goals :soccer:
Add a [HashedCross](https://www.tensorflow.org/api_docs/python/tf/keras/layers/experimental/preprocessing/HashedCrossing)  block as a transformation block

Inheriting from TabularBlock, quite similar to [keras implementation](https://github.com/keras-team/keras/blob/07e13740fd181fc3ddec7d9a594d8a08666645f6/keras/layers/preprocessing/hashed_crossing.py#L32-L198)

### Testing Details :mag:
7 unit tests are created in transformation.py, they are separately for different input and arguments, like scalar, 1d or 2d inputs, less hash bins and one_hot output.


merlin/models/tf/blocks/core/transformations.py
```diff
@@ -113,6 +115,142 @@ def get_config(self):
...
+            raise ValueError(
+                "All `HashedCrossing` inputs should have equal shape. "
+                f"Received: inputs={inputs_tensors}"
+            )
+        # TODO: Consider transfer tensors to dense tensors, do not require users do it by themself?
+        if any(isinstance(x, (tf.RaggedTensor, tf.SparseTensor)) for x in inputs_tensors):
+            raise ValueError(
+                "All `HashedCrossing` inputs should be dense tensors. "
+                f"Received: inputs={inputs_tensors}"
+            )
```
timmy00, line 249: Do we support call another transformation block inside this block? Or we should make `to_dense` as `pre` of a TabularBlock? And do we support it now?

merlin/models/tf/blocks/core/transformations.py
```diff
@@ -113,6 +115,142 @@ def get_config(self):
...
+                f"or `[batch_size, 1]`. Received: inputs shape={first_shape}"
+            )
+        if not all(x.shape.as_list() == first_shape for x in inputs_tensors):
+            raise ValueError(
+                "All `HashedCrossing` inputs should have equal shape. "
+                f"Received: inputs={inputs_tensors}"
+            )
+        # TODO: Consider transfer tensors to dense tensors, do not require users do it by themself?
+        if any(isinstance(x, (tf.RaggedTensor, tf.SparseTensor)) for x in inputs_tensors):
+            raise ValueError(
```
timmy00, line 246: Do we support call another transformation block inside this block? Or we should make `to_dense` as `pre` of a TabularBlock? And do we support it now?

marcromeyn (approved):

timmy00:
Close it and open another [PR](https://github.com/NVIDIA-Merlin/models/pull/587) to clean up history, since too many commit from main.

PR #528, opened 2022-06-23
XGBoost - Use DaskDeviceQuantileDMatrix with GPU Training

### LINKED ISSUE #525: [FEA] Provide the option of training xgboost models using `DeviceQuantileDMatrix`
# 🚀 Feature request

As outlined in this [article](https://medium.com/rapids-ai/new-features-and-optimizations-for-gpus-in-xgboost-1-1-fc153dc029ce), `DeviceQuantileDMatrix` offers significant advantages in terms of speed of execution and GPU memory footprint. It would be great if we could support it!

image

This can be very useful across a large number of scenarios (including enabling some workloads that might be terminating with OOM errors, though that is slightly hypothetically, maybe this aspect is already fully addressed through using a `dask` backend).

If an example of using `DeviceQuantileDMatrix` with `xgboost` could be helpful, please let me know, can provide this.

oliverholworthy:
Fixes #525 

### Goals :soccer:

Use more efficient datatype for training with GPU + XGBoost by default.

The DeviceQuantileDMatrix is recommended when using GPU to reduce memory usage and improve speed of training.

When `tree_method="gpu_hist"` is used. Automatically switch the datatype used for training from `DaskDMatrix` to `DaskDeviceQuantileDMatrix`.

### Testing Details :mag:

Test added for the default behaviour with gpu_hist - that the data used for training is of type `DaskDeviceQuantileDMatrix`. And checking that the optional argument added to the fit method `use_quantile=False` can be used to disable this.

### DIFF
--- a/merlin/models/xgb/__init__.py
+++ b/merlin/models/xgb/__init__.py
@@ -76,6 +76,8 @@ class XGBoost:
     def fit(
         self,
         train: Dataset,
+        *,
+        use_quantile=True,
         **train_kwargs,
     ) -> xgb.Booster:
         """Trains the XGBoost Model.
@@ -90,6 +92,11 @@ class XGBoost:
             The training dataset to use to fit the model.
             We will use the column(s) tagged with merlin.schema.Tags.TARGET that match the
             objective as the label(s).
+        use_quantile : bool
+            This param is only relevant when using GPU.  (with
+            tree_method="gpu_hist"). If set to False, will use a
+            `DaskDMatrix`, instead of the default
+            `DaskDeviceQuantileDMatrix`, which is preferred for GPU training.
         **train_kwargs
             Additional keyword arguments passed to the xgboost.train function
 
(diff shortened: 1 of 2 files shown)


radekosmulski:
Looks great! 🙌 🙂 Having this functionality will be very useful for running on GPU/large amounts of data!

One parameter that is very helpful when instantiating a `DaskDeviceQuantileDMatrix` is `max_bin`. It is the knob that allows you to turn down the precision of the calculation in order to limit your memory footprint (or conversely, to dial it up which can be quite useful in case of skewed data).

Your implementation is very elegant and I don't want to mess it up in any way. But I am wondering -- would there be a way to expose the `max_bin` parameter to the user?

radekosmulski:
@oliverholworthy I tried `max_bin` out and the default value of 256 seems to offer good performance and what is probably even more important, changing this value doesn't impact memory footprint to the extent I anticipated, there is an effect but smaller than I have imagined.

I am thinking that probably working towards exposing the `max_bin` parameter would be getting too much into the weeds and is likely not worth it at this point in time, given the other higher-level but generally much more useful functionality that we could have.

radekosmulski (approved):

PR #529, opened 2022-06-24
Migrate `FeatureCollection` to use the version from Merlin Core

karlhigley:
Depends https://github.com/NVIDIA-Merlin/core/pull/102

### Goals :soccer:
This PR aims to connect Merlin Models to standard abstractions for array-like data and dataframe-like objects that can be used throughout the Merlin eco-system.

We've moved `FeatureCollection` into Merlin Core (where it became `Features`) and connected it to cross-framework data transfer functionality from `MerlinArray` (which isn't super-useful inside Tensorflow models, but should be helpful for standard evaluation metrics.)

### Testing Details :mag:
 Covered by existing tests.


viswa-nvidia:
Pushed to arbitration - @karlhigley , how does this map to the roadmap ?

karlhigley:
@viswa-nvidia It stemmed from the `model.compile` work from a release or two ago, and was also part of the foundation of batch processing, but the PR in `core` it's based on needs some updates and has never been merged.

PR #537, opened 2022-06-28
Integration Tests for Retrieval models

### LINKED ISSUE #482: [FEA] Create a integration test for retrieval models based on the research experiments
# 🚀 Feature request

Create integration tests for retrieval models based on the research experiments.

## Motivation

We have created a script using advanced options of Merlin Models API to configure, build and train retrieval models (MF, Two Tower and YouTubeDNN). In addition, we performed a few hypertuning sessions to get best performance of those models for different datasets (H&M, LastFM).

This request is about creating integration tests for two reasons:
1. Fail when there are changes in our API related to Retrieval Models
2. Track over time if we have a drop in accuracy or performance (runtime) of those tests compared to the baselines (results from the research experiments).

gabrielspmoreira:
Fixes #482 

### Goals :soccer:
Create a set of integration tests for retrieval models to check for
- **Stability of our API** - by using advanced options of the APIs that might break if changed
- **Accuracy and Throughput** - Compute accuracy and throughput for training and evaluation, and log properly so that it is possible to track over time if we have a drop in those metrics (compared to the best hypertuned results from the retrival experiments).

- Uses [Fiddle](https://github.com/google/fiddle) for Python-first configuration management
- The retrieval training and evaluation must be available to learn in both PyTest (for CI) and also as a command line script (where hparams are provided as command line arguments)

### Testing Details :mag:
- In progress

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -24,3 +24,4 @@ myst-nb==0.13.2
 linkify-it-py==1.0.3
 sphinx-multiversion@git+https://github.com/mikemckiernan/sphinx-multiversion.git
 sphinxcontrib-copydirs@git+https://github.com/mikemckiernan/sphinxcontrib-copydirs.git
+fiddle
--- a/tests/integration/tf/retrieval/retrieval_config.py
+++ b/tests/integration/tf/retrieval/retrieval_config.py
@@ -1,0 +1,55 @@
+import fiddle as fdl
+
+from tests.integration.tf.retrieval.retrieval_utils import (
+    RetrievalTrainEvalRunner,
+    get_callbacks,
+    get_dataset,
+    get_dual_encoder_model,
+    get_item_frequencies,
+    get_loss,
+    get_metrics,
+    get_optimizer,
+    get_samplers,
+    get_schema,
+    get_youtube_dnn_model,
+)
+
+
+def config_retrieval_train_eval_runner(model_type, data_path):
+    def make_datasets(schema):
+        get_dataset_partial = fdl.Partial(get_dataset, schema)
+        train_ds = get_dataset_partial(dataset="train")
+        train_ds.data_path = schema.data_path
+        eval_ds = get_dataset_partial(dataset="valid")
+        eval_ds.data_path = schema.data_path
+        return train_ds, eval_ds
+
+    def make_model(schema, train_ds, model_type="two_tower"):
+        if model_type == "youtubednn":
+            model = fdl.Config(get_youtube_dnn_model, schema)
+        else:
+            samplers = fdl.Config(get_samplers, schema)
+            items_frequencies = fdl.Config(get_item_frequencies, schema, train_ds)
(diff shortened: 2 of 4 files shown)


gabrielspmoreira:
rerun tests

gabrielspmoreira:
rerun tests

gabrielspmoreira:
rerun tests

gabrielspmoreira:
rerun tests

tests/common/tf/retrieval/retrieval_utils.py
```diff
@@ -0,0 +1,575 @@
...
+from merlin.models.tf.blocks.core.transformations import PopularityLogitsCorrection
+from merlin.models.utils import schema_utils
+from merlin.schema.tags import Tags
+
+
+def filter_schema(schema, filter_dataset_features=""):
+    schema = schema.select_by_tag([Tags.ITEM_ID, Tags.USER_ID, Tags.ITEM, Tags.USER])
+
+    if filter_dataset_features:
+        schema = schema.select_by_name(filter_dataset_features.split(","))
```
marcromeyn, line 40: Now that we use Fiddle instead of a CLI, would it make sense to change this to a list instead of a string that needs to be splitted?

tests/common/tf/retrieval/retrieval_utils.py
```diff
@@ -0,0 +1,575 @@
...
+            clipnorm=opt_clip_norm,
+            clipvalue=opt_clip_value,
+        )
+    else:
+        raise ValueError("Invalid optimizer")
+
+    return opt
+
+
+class ExamplesPerSecondCallback(tf.keras.callbacks.Callback):
```
marcromeyn, line 364: Should we move this to merlin-models? Might be useful outside of these tests as well.

tests/common/tf/retrieval/retrieval_utils.py
gabrielspmoreira: Hey @marcromeyn . I think we can keep it as a string so that if we decide later to create a CLI based on Fiddle to run these unit tests with different arguments the string arg would be easier to parse that a list. what do you think?

tests/common/tf/retrieval/retrieval_utils.py
gabrielspmoreira: @marcromeyn  The `ExamplesPerSecondCallback` depends on `wandb` library. It is currently required only when running the integration tests. I have added the `wandb` only as dev requirement, so that CI and developers that want to check the metrics of the retrieval integration tests can do so. 
I think that maybe in the future we might having a logging module for Merlin Models, where me provide callbacks and utils for logging tools like Tensorboard, W&B, MLFlow. Then maybe then we could move this `ExamplesPerSecondCallback` to such logging module.
Does that make sense to you?

marcromeyn: I think we can remove that dependency but we can do that in a future PR

marcromeyn (approved):

PR #542, opened 2022-06-29
Adding pre & post to Model

marcromeyn:
### Goals :soccer:
This makes a `Model` more block-like by adding `pre` & `post`. This is a flexible way for a user to add transformations to the model in an easy way. Some of the use-cases would be negative-sampling (for ranking), swap noise & simple masking for youtube dnn.

Example of 
```python
model = ml.Model(
        ml.InputBlock(schema),
        ml.MLPBlock([64]),
        ml.BinaryClassificationTask("click"),
        pre=ml.StochasticSwapNoise(schema),
)
```

### DIFF
--- a/merlin/models/xgb/__init__.py
+++ b/merlin/models/xgb/__init__.py
@@ -67,7 +67,7 @@ class XGBoost:
         if objective.startswith("rank") and qid_column is None:
             qid_column = schema.select_by_tag(Tags.USER_ID).column_names[0]
         self.qid_column = qid_column
-
+        self.evals_result = {}
         self.booster = booster
 
     @property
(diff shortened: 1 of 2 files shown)


oliverholworthy:
Is the use of `pre/post` currently functionally equivalent to passing the block in along with the others in the right order (the first/last block/layer)?

```python
model = mm.Model(
    mm.InputBlock(schema), mm.MLPBlock([64]), mm.BinaryClassificationTask("click"), pre=mm.StochasticSwapNoise(schema)
)

# vs

model = mm.Model(
   mm.StochasticSwapNoise(schema), mm.InputBlock(schema), mm.MLPBlock([64]), mm.BinaryClassificationTask("click")
)
```

I'm trying to think of an example where the use of `pre/post` would make things possible or easier that are not currently, (If the above is not equivalent).

Is this motivated by the ability to run some parts of the model during training only, e.g.  StochasticSwapNoise seems like a concept that you might only want to apply during training and not inference. Or is that functionality enabled with a different mechanism?

PR #543, opened 2022-06-29
Move tf/blocks/core to tf/core

marcromeyn:
### Goals :soccer:
Making our core abstractions easier to navigate to. 

Moving tf/blocks/core to tf/core (same for tests)


oliverholworthy:
If this is about directory depth, another option could be to move one directory up

e.g. `merlin.models.tf.blocks.core.combinators` -> `merlin.models.tf.blocks.combinators` (in other words drop the `core`)

I think the way this PR is structured now might be confusing in the sense that we have blocks both inside and outside of the blocks directory. (We already do I suppose with the Input block that moved recently). Or is the end goal to remove the blocks directory, and move everything out into parent directories split by concepts there? And this PR is taking us partly toward that state?

Hopefully this structure choice shouldn't matter too much for users of the library if we expect people to generally import everything from `merlin.models.tf`

examples/06-Define-your-own-architecture-with-Merlin-Models.ipynb
```diff
@@ -1042,88 +1042,88 @@
   {
```
oliverholworthy, line 1042: <pre class="ql-syntax" spellcheck="false">Line #1.    from merlin.models.tf.core.transformations import LogitsTemperatureScaler
</pre><p>Should LogitsTemperatureScaler be available from merlin.models.tf ? It looks like other transformations are exposed there.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/543/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='51'/>

marcromeyn:
I was thinking that `tf.blocks` would contain all the blocks that users might want to use (for certain architectures). `tf.core` would then contain all base-classes.

PR #548, opened 2022-06-29
Add functionality to set activation as a list in MLPBlock

### LINKED ISSUE #467: [FEA]  Add functionality to be able to set activation as a list in MLPBlock
# 🚀 Feature request

Currently, one can set the activation func as `str` in [MLPBlock](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/blocks/mlp.py#L37), and this way they can use the same activation func for every layer. However, one might want to use different activation func for each Dense layer.  In order to make this happen, one should be able to set `activation` arg as a list as well.

rvk007:
Fixes #467 
- Added an option to provide activation as `str` or a `list of str` in [MLPBlock](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/blocks/mlp.py#L37).
- Performed basic tests with example [notebook](https://github.com/NVIDIA-Merlin/models/blob/main/examples/01-Getting-started.ipynb).

### DIFF
--- a/merlin/models/tf/blocks/mlp.py
+++ b/merlin/models/tf/blocks/mlp.py
@@ -34,7 +34,7 @@ RegularizerType = Union[str, tf.keras.regularizers.Regularizer]
 
 def MLPBlock(
     dimensions: List[int],
-    activation: str = "relu",
+    activation: Union[str, List[str]] = "relu",
     use_bias: bool = True,
     kernel_initializer: InitializerType = "glorot_uniform",
     bias_initializer: InitializerType = "zeros",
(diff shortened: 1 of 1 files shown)


merlin/models/tf/blocks/mlp.py
```diff
@@ -86,15 +86,19 @@ def MLPBlock(
         The name of the block.
     """
 
+    if isinstance(activation, list) and len(activation) != len(dimensions):
+        raise ValueError("Activation list should be of same length as Dimensions")
```
marcromeyn, line 90: Could be useful to add the lenghts of both lists to the error-message.

rvk007: Yes, correct. I have added that information.

PR #552, opened 2022-07-06
Add Cond Layer to tensorflow combinators

oliverholworthy:
Motivated by https://github.com/NVIDIA-Merlin/Merlin/issues/269

### Goals :soccer:

Enable conditional logic based layer inputs or attributes.

One example that motivated this layer was a way to conditionally run a layer for training or testing.

```python
class TrainOrTest(Layer):
    def call(self, inputs, training=False, testing=False):
        return training or testing

mm.Cond(TrainOrTest(), train_or_test_layer)
```

This Cond layer could also be used for dynamically changing other things about the model structure or inputs.

### Testing Details :mag:

Using the `tensorflow.keras.__internal__.utils.layer_test` helper to make sure this layer works correctly. However, the `__internal__` package doesn't feel like something we should be importing here.


merlin/models/tf/blocks/core/combinators.py
```diff
@@ -602,3 +603,69 @@ def __init__(
...
+    def compute_output_shape(self, input_shape):
+        """Computes the output shape of the layer."""
+        true_output_shape = self.true.compute_output_shape(input_shape)
+
+        if self.false:
+            false_output_shape = self.false.compute_output_shape(input_shape)
+        else:
+            false_output_shape = input_shape
+
+        if true_output_shape != false_output_shape:
```
marcromeyn, line 641: Should we add a flag in `init` to specify whether this should throw an exception?

marcromeyn (approved):

PR #554, opened 2022-07-07
Allow for blocks in the Model to transform the targets

marcromeyn:
### Goals :soccer:

- Enable in-batch negative sampling as part of the model layers.

- Replaces `FeatureContext` with `PredictionContext`
  PredictionContext is a NamedTuple containing {features, targets, mask, training, testing}. Replaces the use of the FeatureContext which was a python class. This may have some performance benefit.

- Using `keras.utils.tf_inspect.getfullargspec` instead of `inspect.getfullargspec` in the `call_layer` function

- Fixes a bug in the `compute_output_shape` method of `MLPBlock`
  Calling self.dense instead of super(_Dense, self) returns the correct output shape

- `UniformNegativeSampling` support for tensors > 2D
  replacing `tf.squeeze` with `tf.reduce_all` to get the batch-dimension mask correctly

- `UniformNegativeSampling` returns Prediction namedtuple
  This negative sampling layer can now be used in a model block to add negatives to the inputs and targets.

- Implementation of `Model` updated to support layers accepting and returning outputs, targets, and features
  First example of this is the  `UniformNegativeSampling` layer

### Testing Details :mag:

- Found a bug in the `ContinuousFeatures` compute_call_output_shape method in graph mode
  In one of the added tests the inputs have unknown shape


merlin/models/tf/models/base.py
```diff
@@ -666,6 +657,18 @@ def __init__(
...
 
+    def _maybe_build(self, inputs):
+        if isinstance(inputs, dict):
+            _ragged_inputs = AsRaggedFeatures()(inputs)
+            feature_shapes = {k: v.shape for k, v in _ragged_inputs.items()}
+            feature_dtypes = {k: v.dtype for k, v in _ragged_inputs.items()}
+
+            for block in self.blocks:
+                for child in block.submodules:
+                    child._feature_shapes = feature_shapes
```
oliverholworthy, line 670: 📝 this adds metadata to layers at build time, intended to be accessible in compute_output_shape for example.

merlin/models/tf/__init__.py
```diff
@@ -181,6 +180,7 @@
     "AvgPrecisionAt",
     "RecallAt",
     "BaseModel",
+    "PredictionContext",
```
oliverholworthy, line 183: If this is intended to be an internal implementation detail, then maybe we should remove from this top-level export?

oliverholworthy (approved):

PR #556, opened 2022-07-08
Adding AsRaggedTensors transformation

marcromeyn:
### Goals :soccer:
Add a transformation to convert list-columns to ragged-tensors.

This transformation is used in the to-dense & to-sparse.


merlin/models/tf/blocks/core/transformations.py
```diff
@@ -27,15 +27,16 @@
...
 from merlin.schema import Schema, Tags
 
 
-@Block.registry.register("as-sparse")
+@Block.registry.register("as-ragged")
 @tf.keras.utils.register_keras_serializable(package="merlin.models")
-class AsSparseFeatures(TabularBlock):
+class AsRaggedFeatures(TabularBlock):
     """
     Convert inputs to sparse tensors.
```
oliverholworthy, line 41: 💡 Updating the docstring here would help clarify that this is only applying to list-like columns. similar to the AsDenseFeatures block.

oliverholworthy (approved):

PR #560, opened 2022-07-11
Add In-Batch Negative Sampling Block for positive-only batches

oliverholworthy:
Fixes # (issue)

### Goals :soccer:

- Enable In-batch negative sampling with an existing block

Limited to batches with only positive examples and binary targets

### Testing Details :mag:

- [ ] Raise exception if target expectations are not met


merlin/models/tf/sampling/base.py
```diff
@@ -0,0 +1,91 @@
...
+        mask = tf.concat([tf.expand_dims(tf.repeat(True, batch_size), 1), mask], 0)
+
+        # 3. Loop through all features:
+        #   - For item-feature: append from item-collection
+        #   - For user-feature: repeat `n_per_positive` times
+        item_cols = self.schema.column_names
+        outputs = {}
+        for name, val in inputs.items():
+            if isinstance(val, tuple):
+                val = tf.RaggedTensor.from_row_lengths(val[0][:, 0], val[1][:, 0])
```
marcromeyn, line 75: Now that we have that helper function [list_col_to_ragged](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/utils/tf_utils.py#L366), we should probably use that here.

marcromeyn (approved):

PR #567, opened 2022-07-13
Support training with UniformNegativeSampling

oliverholworthy:
Follow-up to #560

### Goals :soccer:

- Enable training with UniformNegativeSampling in the model

- Implementing to/from config on UniformNegativeSampling layer
- If  `batch_size` is none on the inputs, we can't do anything, so returning early from call method.

### Testing Details :mag:

- Using model_test helper to check that the layer works when fitting the model.

### DIFF
--- a/merlin/models/tf/data_augmentation/negative_sampling.py
+++ b/merlin/models/tf/data_augmentation/negative_sampling.py
@@ -20,8 +20,8 @@ import tensorflow as tf
 from merlin.models.tf.core.prediction import Prediction
 from merlin.models.tf.typing import TabularData
 from merlin.models.tf.utils.tf_utils import list_col_to_ragged
-from merlin.schema import Schema, Tags
 from merlin.models.utils import schema_utils
+from merlin.schema import Schema, Tags
 
 
 @tf.keras.utils.register_keras_serializable(package="merlin.models")
--- a/merlin/models/tf/data_augmentation/negative_sampling.py
+++ b/merlin/models/tf/data_augmentation/negative_sampling.py
@@ -21,6 +21,7 @@ from merlin.models.tf.core.prediction import Prediction
 from merlin.models.tf.typing import TabularData
 from merlin.models.tf.utils.tf_utils import list_col_to_ragged
 from merlin.schema import Schema, Tags
+from merlin.models.utils import schema_utils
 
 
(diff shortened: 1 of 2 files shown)


tests/unit/tf/data_augmentation/test_negative_sampling.py
```diff
@@ -81,15 +82,28 @@ def test_calling(self, music_streaming_data: Dataset, to_dense: bool, tf_random_
...
+        dataset = music_streaming_data
+        schema = dataset.schema
+
+        @tf.keras.utils.register_keras_serializable(package="merlin.models")
         class Training(tf.keras.layers.Layer):
             def call(self, inputs, training=False):
                 return training
 
+        @tf.keras.utils.register_keras_serializable(package="merlin.models")
+        class Identity(tf.keras.layers.Layer):
```
oliverholworthy, line 94: @marcromeyn The Cond layer throws an exception without this, I think because the return value is different depending on whether or not targets is None or not.

merlin/models/tf/data_augmentation/negative_sampling.py
```diff
@@ -46,6 +47,11 @@ def call(self, inputs: TabularData, targets=None) -> Union[Prediction, TabularDa
             fist_input.shape[0] if not isinstance(fist_input, tuple) else fist_input[1].shape[0]
         )
 
+        if batch_size is None:
```
oliverholworthy, line 50: If batch dimension is `None`, there's not much we can do here

oliverholworthy: This was raising an exception when calling model.fit on a dataset at some stage in the calls made to the layer

tests/unit/tf/data_augmentation/test_negative_sampling.py
marcromeyn: Can’t we replace the `NoOp` inside the `Cond` with this? So that this is used by default and our users don’t have it call it explicitly?

oliverholworthy: I don't think we can because this logic is specific to the way the sampling layer works. sometimes an identity function and sometimes transforms into a tuple

marcromeyn (approved):

PR #572, opened 2022-07-14
Keep Dataloader lazy when using map operations

oliverholworthy:
This preserves the iterator nature of BatchedDataset when using a map operation.

### Goals :soccer:

Improve consistency of dataloader operation when adding map operations.

The BatchedDataset can be iterated through lazily after being constructed. However, if a map is applied, it runs on the whole dataset before returning the first batch,

#### Example

Create example merlin Dataset

```python
import numpy as np
import pandas as pd

dataset_size = 10_000
data_df = pd.DataFrame({"feature": np.random.randint(100, size=dataset_size)})

from merlin.io import Dataset
from merlin.schema import Schema, ColumnSchema, Tags

schema = Schema([ColumnSchema("feature", tags=[Tags.CATEGORICAL])])
dataset = Dataset(data_df, schema=schema)
```

Timing getting the first batch from the dataset with and without the `.map`

```python
import time
import timeit
from merlin.models.tf.dataset import BatchedDataset

# -----------------------------------------------------------------------------
# < 0.1 seconds - without .map

batched_dataset = BatchedDataset(dataset, batch_size=10)
timeit.timeit(lambda: next(batched_dataset), number=1)

# -----------------------------------------------------------------------------
# ~ 100 seconds - with a .map 

batched_dataset = BatchedDataset(dataset, batch_size=10)

def identity(x, y):
    time.sleep(0.1)
    return (x, y)

batched_dataset = batched_dataset.map(identity)
timeit.timeit(lambda: next(batched_dataset), number=1)
```

With this PR, both versions in the above example return in less than or approx. 0.1 seconds (in the case of the time.sleep)

Replacing the list comprehension in the Dataloader `make_tensors` method return a generator instead.

### Testing Details :mag:

Turned the above example into a unit test.

### DIFF
--- a/merlin/models/loader/backend.py
+++ b/merlin/models/loader/backend.py
@@ -492,7 +492,7 @@ class DataLoader:
                     c = (c, batch_lists)
 
                 batches[n].append(c)
-        return [self._handle_tensors(*batch) for batch in batches]
+        return (self._handle_tensors(*batch) for batch in batches)
 
     def _get_segment_lengths(self, num_samples):
         """
--- a/tests/unit/tf/test_dataset.py
+++ b/tests/unit/tf/test_dataset.py
@@ -14,6 +14,8 @@
 # limitations under the License.
 #
 import os
+import time
+import timeit
 
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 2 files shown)


benfred (approved):

benfred:
This looks good to me - Thanks for the fix! I will make sure this fix also gets into the new dataloader package @jperez999 and I are putting together

jperez999 (approved):
Nice move!

PR #574, opened 2022-07-14
Wrap Features in Class for compatibility with ItemRetrievalScorer

oliverholworthy:
Follow up to #554 

In #554 ,among other things, we changed the way features are stored and passed around to layers that need the context of the inputs. e.g. the `ItemRetrievalScorer`.

After #554 tests using the `RetrievalModel.evaluate` method started breaking with tensorflow version 2.8.0. (The tests pass with the newer release 2.9.1.)

The error message is:

```
>       metrics = model.evaluate(
            ecommerce_data, batch_size=10, item_corpus=ecommerce_data, return_dict=True
        )

tests/unit/tf/models/test_retrieval.py:244:

merlin/models/tf/models/base.py:878: in evaluate                                                                                                                                                                           
    self.pre_eval_topk = TopKIndexBlock.from_block(
merlin/models/tf/core/index.py:208: in from_block
    return super().from_block(block=block, data=data, id_column=id_column, k=k, **kwargs)  
merlin/models/tf/core/index.py:81: in from_block                                                                                                                                                                            
    embedding_df = cls.get_candidates_dataset(block, data, id_column)
merlin/models/tf/core/index.py:112: in get_candidates_dataset
    model_encode = TFModelEncode(model=block, output_concat_func=np.concatenate)
merlin/models/tf/utils/batch_utils.py:80: in __init__                                                                                                                                                                       
    model.save(save_path)

[...]

>         raise e.ag_error_metadata.to_exception(e)
E         TypeError: in user code:
E
E             File "/usr/local/lib/python3.8/dist-packages/keras/saving/saving_utils.py", line 138, in _wrapped_model  *
E                 outputs = model(*args, **kwargs)
E             File "/workspace/dev/merlin/models/merlin/models/config/schema.py", line 58, in __call__  *
E                 return super().__call__(*args, **kwargs)
E
E             TypeError: tf__call() takes 2 positional arguments but 3 were given
```

The issue is that the model object. In this case the item_block of a two tower model thinks it accepts 2 arguments when actually it only wants one. 

We can check this with the helper function  `model_call_inputs` and a breakpoint in the `RetrievalModel.evaluate` method.  which reports that the item_block has 2 arguments. (The error "_2 positional arguments but 3 were given_" is not 1 vs 2 because of the additional `self` argument.)

```python
from keras.saving.saving_utils import model_call_inputs

len(model_call_inputs(item_block)[0])  # => 2
```

By wrapping the features in a class, somehow this results in the model helper correctly returning 1 instead of 2.

```python
class Features:
    def __init__(self, values: Dict[str, TensorLike]):
        self.values = values

    def __getitem__(self, key: str) -> TensorLike:
        return self.values[key]
```

This is not necessary in tensorflow 2.9.0 and 2.9.1.

### Testing Details :mag:

Tested on 2.8.0 and 2.9.0 locally

PR #573 extends our testing matrix to check both these versions

### DIFF
--- a/merlin/models/tf/core/prediction.py
+++ b/merlin/models/tf/core/prediction.py
@@ -20,8 +20,16 @@ import tensorflow as tf
 TensorLike = Union[tf.Tensor, tf.SparseTensor, tf.RaggedTensor]
 
 
+class Features:
+    def __init__(self, values: Dict[str, TensorLike]):
+        self.values = values
+
+    def __getitem__(self, key: str) -> TensorLike:
+        return self.values[key]
+
+
 class PredictionContext(NamedTuple):
-    features: Dict[str, TensorLike]
+    features: Features
     targets: Optional[Union[tf.Tensor, Dict[str, tf.Tensor]]] = None
     mask: tf.Tensor = (None,)
     training: bool = False
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -13,7 +13,7 @@ from tensorflow.keras.utils import unpack_x_y_sample_weight
 import merlin.io
 from merlin.models.tf.core.base import Block, ModelContext, PredictionOutput, is_input_block
 from merlin.models.tf.core.combinators import SequentialBlock
-from merlin.models.tf.core.prediction import Prediction, PredictionContext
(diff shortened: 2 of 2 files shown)


oliverholworthy:
We have tracked down the underlying reason for this in the way the save model spec is computed. And have adopted a different solution in  #582

PR #581, opened 2022-07-15
Add multi optimizer

timmy00:
### Goals :soccer:
Add a multi-optimizer class, It allows different optimizers to be applied to different subsets of the model's variables. For example, it makes it possible to apply one optimizer to the blocks which contains the model's embeddings (sparse variables) and another optimizer to the rest of its variables (other blocks).

For example:
```
      user_tower = ml.InputBlock(schema.select_by_tag(Tags.USER), ml.MLPBlock([512, 256]))
      item_tower = ml.InputBlock(schema.select_by_tag(Tags.ITEM), ml.MLPBlock([512, 256]))
      two_tower = ml.ParallelBlock({"user": user_tower, "item": item_tower})
      model = ml.Model(two_tower, ml.ItemRetrievalTask())
      optimizer = MultiOptimizer(default_optimizer="adam",
        optimizers_and_blocks=[
          (tf.keras.optimizers.SGD(), user_tower),
          (tf.keras.optimizers.Adam(), item_tower),
        ])

      # The string identification of optimizer is also acceptable:
      optimizer = MultiOptimizer(default_optimizer="adam",
        optimizers_and_blocks=[
          ("sgd", user_tower),
          ("adam", item_tower),
        ])
```

[Note about get_trainable_variables_optimizer_dict() method ](https://docs.google.com/document/d/1J92fPLhH66O1EOObHZjALOIHg4b_U7skjixTclkz-A0/edit?usp=sharing)

Similar to [TensorFlow recommendation composite_optimizer](https://github.com/tensorflow/recommenders/blob/ed24b3e3ac1860001c9a920ed3d395c1201ab81a/tensorflow_recommenders/experimental/optimizers/composite_optimizer.py)

### Testing Details :mag:
6 unit tests:
1. test_optimizers: basic functional test to check if the multi-optimizers are working correctly
2. test_model_with_multi_optimizers: test if model with multi-optimizers works well with model_test
3. test_multi_optimizer_list_input: test the argument optimizers_and_blocks consisted of a list
4. test_multi_optimizer_add: test the MultiOptimizer.add() function
5. test_multi_optimizers_from_config: test get_config() and from_config()
6. test_multi_optimizers_from_config_list_input: test get_config() and from_config() for list input


merlin/models/tf/blocks/multi_optimizers.py
```diff
@@ -0,0 +1,214 @@
...
+          ("sgd", user_tower),
+          ("adam", item_tower),
+        ])
+    ```
+
+    """
+
+    def __init__(
+        self,
+        default_optimizer: tf.keras.optimizers.Optimizer,
```
marcromeyn, line 65: Shouldn’t this be optional, since it could be that the user provides a optimizer for all variables using the `optimizers_and_blocks` argument?

merlin/models/tf/blocks/multi_optimizers.py
```diff
@@ -0,0 +1,214 @@
...
+        optimizers_and_blocks: Sequence[
```
marcromeyn, line 66: I think it could make sense to make this optional, and add a method `add_optimizer` to the class. This allows for more flexibility.

timmy00: Hi Marc, I implemented the [add method ](https://github.com/NVIDIA-Merlin/models/blob/89b9763378e0f5d61562831230054989ceab7efd/merlin/models/tf/blocks/multi_optimizers.py#L148) in the class, and I am not sure if it is just as you want.

merlin/models/tf/blocks/multi_optimizers.py
timmy00: Yeah, make sense, I think we can make it optional, and if there is variable not given in optimizers_and_blocks by users, and also do not set this default_optimizer argument, we could warn that some variables would be assigned with the default default_optimizer.

merlin/models/tf/blocks/multi_optimizers.py
```diff
@@ -0,0 +1,225 @@
...
+          ("adam", item_tower),
+        ])
+    ```
+
+    """
+
+    def __init__(
+        self,
+        optimizers_and_blocks: Sequence[
+            Tuple[Union[str, tf.keras.optimizers.Optimizer], Sequence[Block]]
```
gabrielspmoreira, line 66: @timmy00 I see that the `optimizers_and_blocks` argument expects a sequence of tuples where first element is the optimizer and second element is the block (or list of blocks).
But the user might provide a tuple with more than length different than 2, or invert the order of elements.
What about changing it to accept instead a sequence of [data class](https://docs.python.org/3/library/dataclasses.html) (instead of the tuple), so that it is clear for the API user the structure of the expected input?

timmy00: Thank you Gabriel, I have fixed it.

gabrielspmoreira (approved):

PR #584, opened 2022-07-15
Add Keras layer_test function to testing_utils

oliverholworthy:
### Goals :soccer:

Improve testing of layers in our library

The  `layer_test` function in keras is helpful for ensuring our layers implement the required methods in a variety of supported uses.

We are copying it into the repo here because as far as we can tell it wasn't publicly exposed in the library prior to 2.9.0

### Testing Details :mag:

Replace usage of layer_test with the one we have copied in.


marcromeyn (approved):

mikemckiernan:
I'm stumped.  I can't rep the docs build failure.  I'm using Oliver's commit, `7d264494`, and I followed the same `pip install` commands that are in the `tensorflow.yml` file.  I'm baffled.

PR #586, opened 2022-07-15
Attempt to remove docs dependencies

mikemckiernan:
I don't understand how these dependencies could cause a build error, but we ought to learn something with this PR.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -54,7 +54,6 @@ extensions = [
     "myst_nb",
     "sphinx_multiversion",
     "sphinx_rtd_theme",
-    "sphinx_markdown_tables",
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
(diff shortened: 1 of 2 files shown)


mikemckiernan:
This change seems to work, though it seems a heavy-handed change.

@benfred has an alternative that pins the version of a dependency:

```
# needed to avoid bug in sphinx-markdown-tables
# https://github.com/ryanfox/sphinx-markdown-tables/issues/36
markdown==3.3.7
```

FYI, I made a small and simple Markdown table in one of the files after this change and the table still parsed to HTML.  I'm crediting MyST-NB for that, though I'm not 100% sure.

benfred (approved):
lgtm! I'd rather remove the dependency (and avoid all future problems) than fix the versions like I was suggesting =).

I scanned through the generated docs, and it looked good to me.

PR #588, opened 2022-07-18
Add signature checks for various models like two-tower & DLRM

marcromeyn:
**Fixes:**
- https://github.com/NVIDIA-Merlin/Merlin/issues/468

### Goals :soccer:
Add some signature-def checks to our model-test, to ensure that the model can be served by Merlin-systems.

Checks are based on [this[(https://github.com/NVIDIA-Merlin/systems/blob/main/merlin/systems/dag/ops/tensorflow.py#L66-L87) of `PredictTensorflow` inside `merlin.systems`.

This PR contains a fix where in the `Model` we check if a child contains any `ModelBlock`s and if so we don’t supply the features argument. The features argument will be added automatically inside the `ModelBlock`. This mechanism ensures that the features kwarg is not added to the model-signuture.

### DIFF
--- a/merlin/models/tf/utils/testing_utils.py
+++ b/merlin/models/tf/utils/testing_utils.py
@@ -13,9 +13,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+import pathlib
 import platform
 import tempfile
-from typing import Any, Tuple, Union
+from typing import Any, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import pytest
@@ -26,6 +27,7 @@ from tensorflow.python.framework.test_util import disable_cudnn_autotune
 import merlin.io
 from merlin.models.tf.dataset import BatchedDataset
 from merlin.models.tf.models.base import Model
+from merlin.schema import Schema
 
 
 def mark_run_eagerly_modes(*args, **kwargs):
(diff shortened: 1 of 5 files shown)


tests/unit/tf/models/test_retrieval.py
```diff
@@ -54,6 +55,16 @@ def test_two_tower_model(music_streaming_data: Dataset, run_eagerly, num_epochs=
     assert len(losses.epoch) == num_epochs
     assert all(measure >= 0 for metric in losses.history for measure in losses.history[metric])
 
+    query_features = testing_utils.get_model_inputs(
+        music_streaming_data.schema.select_by_tag(Tags.USER), ["user_genres"]
+    )
+    testing_utils.test_model_signature(model.first.query_block(), query_features, ["output_1"])
```
bschifferer, line 61: Why is the output for retrieval tasks `output_1`

bschifferer (approved):

merlin/models/tf/utils/testing_utils.py
```diff
@@ -107,6 +109,39 @@ def model_test(
...
 
 
+def get_model_inputs(schema: Schema, list_cols: Optional[Sequence[str]] = None):
+    list_cols = list_cols or []
+    features = schema.column_names
+
+    # Right now the model expects a tuple for each list-column
+    for list_col in list_cols:
+        features.pop(features.index(list_col))
+        for i in ["1", "2"]:
```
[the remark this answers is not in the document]
marcromeyn: Yeah, for these list-columns we now expect a tuple (since that’s how NVT outputs it).

tests/unit/tf/models/test_retrieval.py
marcromeyn: Good question, I actually just realized this today as well. I think if you don’t explicitly label the output, TF will call it `output_1`. I need to digg a bit deeper in order to figure out how to add custom naming.

PR #592, opened 2022-07-19
Add EmbeddingOptions arg in the DLRM block

rnyak:
This PR adds `EmbeddingOptions` arg to DLRMBlock so that one can train a DLRM model by setting  `embeddings_initializers` and  `  embeddings_l2_reg` hyper-params as in DCNModel.

### DIFF
--- a/merlin/models/tf/blocks/dlrm.py
+++ b/merlin/models/tf/blocks/dlrm.py
@@ -27,6 +27,7 @@ from merlin.schema import Schema, Tags
 def DLRMBlock(
     schema: Schema,
     embedding_dim: int,
+    embedding_options: EmbeddingOptions = None,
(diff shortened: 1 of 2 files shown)


sararb (approved):
Thank you for the PR ! 

I am just adding some context around the proposed changes: 
- The `DLRMBlock` will be refactored in the future to create a `DLRMInputBlock` using the simplified version of InputBlock captured in this [ticket](https://github.com/NVIDIA-Merlin/Merlin/issues/404).  In the meantime, this PR exposes the embedding options in DLRMBlock so as that we could run the HPO experiments (needed for the KDD tutorial) with `embeddings_initializers` and `embeddings_l2_reg` hyper-parameters.

gabrielspmoreira (approved):

Issue #498, opened 2022-06-07
[FEA] cannot properly save and load the  TF Retrieval model


rnyak:
### Bug description
 We want to be able to save the entire Two-Tower model  and load back to be able to do `model.evaluate()` and `model.predict()`. However, we get the following error when we load back the model.   To reproduce the following errors, please run the  [05-Retrieval-Model.ipynb](https://github.com/NVIDIA-Merlin/models/blob/main/examples/05-Retrieval-Model.ipynb) example. Then save and reload the model with the following scripts.

First save the model after `model.fit()` step:
`model.save('two_tower')`

Then when we load back the saved model we get the following error:

```
reloaded = tf.keras.models.load_model('two_tower')
```
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [17], in <cell line: 1>()
----> 1 reloaded = tf.keras.models.load_model('two_tower')

File /usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py:67, in filter_traceback.<locals>.error_handler(*args, **kwargs)
     65 except Exception as e:  # pylint: disable=broad-except
     66   filtered_tb = _process_traceback_frames(e.__traceback__)
---> 67   raise e.with_traceback(filtered_tb) from None
     68 finally:
     69   del filtered_tb

File /usr/local/lib/python3.8/dist-packages/keras/saving/saved_model/load.py:1000, in revive_custom_object(identifier, metadata)
    998   return revived_cls._init_from_metadata(metadata)  # pylint: disable=protected-access
    999 else:
-> 1000   raise ValueError(
   1001       f'Unable to restore custom object of type {identifier}. '
   1002       f'Please make sure that any custom layers are included in the '
   1003       f'`custom_objects` arg when calling `load_model()` and make sure that '
   1004       f'all layers implement `get_config` and `from_config`.')

ValueError: Unable to restore custom object of type _tf_keras_metric. Please make sure that any custom layers are included in the `custom_objects` arg when calling `load_model()` and make sure that all layers implement `get_config` and `from_config`.

```

### Environment details
- Merlin version: 
- Platform:
- Python version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?): TF 2.8.0

Using `merlin-tensorflow-training:22.05` docker image with the latest main branches pulled.

rnyak:
This is also required in this RMP https://github.com/NVIDIA-Merlin/Merlin/issues/271

[...]

PR #594, opened 2022-07-20
Tf/contrastive prediction

marcromeyn:
**Blocked by:**
- #589

### Goals :soccer:
- New implementation of the contrastive task used to train retrieval models
- Fix saving/loading two-tower model: #498

- Register metrics and makes sure TopKMetricsAggregator is serializable  
- Wrap the top-k metrics provided to the model in a  `TopKMetricsAggregator` instance to make sure top-k extraction and label relevant counts are computed once for all metrics. 
- Add a base-class `ContrastivePredictionBlock` . 
- Add `ContrastiveDotProduct` that replaces the ItemRetrievalScorer. This new implementation does not rely on the context layer and has direct access to the query embeddings. 
-  Add a new`DotProductCategoricalPrediction` for prediction tasks including negative sampling (will replace the `ItemRetrievalTask`)
---> (**Note**: This new implementation does not support the TopkIndex evaluation).
---> The motivation is to decouple the contrastive learning task of the RetrievalModel from the evaluation of the top-k recommender. 
---> So the supported evaluation strategy is to use negative sampling, similar to training. 

### Testing Details :mag:

- Add unit test of saving/loading RetrievalModel with the new `DotProductCategoricalPrediction` block. 
- Add unit tests for `ContrastivePredictionBlock`.
- Add a unit test for specifying the negative sampling strategy in the `compile` method.


viswa-nvidia:
Moved to arbitration. Not clear how this maps to the r[oadmap](https://github.com/orgs/NVIDIA-Merlin/projects/14/views/4) . @marcromeyn , please clarify.

merlin/models/tf/predictions/sampling/in_batch.py
```diff
@@ -0,0 +1,89 @@
...
+from typing import Optional
+
+import tensorflow as tf
+
+from merlin.models.tf.predictions.sampling.base import Items, ItemSampler
+
+
+@ItemSampler.registry.register("in-batch")
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class InBatchSampler(ItemSampler):
```
oliverholworthy, line 25: What is different about this compared to the `InBatchSampler` that exists in `tf.blocks.sampling.in_batch` ?

sararb: The objective is to move the sampling under the predictions folder as this is used by the new refactored prediction blocks. There are also some enhancements of the sampling blocks (base and InBatchSampler) to improve the API, which are: 

-  Add a registry for sampling strategies 
-  The class arguments are _last_batch_items_embeddings and _last_batch_items_metadata are wrapped in  a new NamedTuple class `Items`.  This new class will replace replace `EmbeddingWithMetadata` class. 
- Add `get_config` method to `InBatchSampler` for serialization. 

I think we could keep the old sampling strategies until we update the High-level API with the new prediction blocks and the old ItemRetrievalTask / ItemRetrievalScorer gets dropped.

merlin/models/tf/predictions/dot_product.py
```diff
@@ -0,0 +1,371 @@
...
+            post=post,
+            logits_temperature=logits_temperature,
+            **kwargs,
+        )
+
+    def compile(self, negative_sampling=None, downscore_false_negatives=False):
+        self.prediction_with_negatives.negative_sampling = negative_sampling
+        self.prediction_with_negatives.downscore_false_negatives = downscore_false_negatives
+
+    # TODO
```
oliverholworthy, line 118: If this TODO is still relevant, can you add more context of what it refers to (given that the method seems to have an implementation

merlin/models/tf/predictions/sampling/in_batch.py
oliverholworthy: We can follow-up on this in another PR. I'm wondering if the behavior of this class is the same from an interface/api perspective - will we be able to replace uses of the old one without any changes to the surrounding code.

merlin/models/tf/predictions/sampling/in_batch.py
```diff
@@ -0,0 +1,89 @@
...
+        return items
+
+    def sample(self) -> Items:
+        return self._last_batch
+
+    def get_config(self):
+        config = super().get_config()
+        config["batch_size"] = self._batch_size
+
+        # TODO: This is a side-effect, could this lead to problems?
```
oliverholworthy, line 86: can you explain why is this line necessary? something to do with resetting the state?

merlin/models/tf/predictions/dot_product.py
```diff
@@ -0,0 +1,371 @@
...
+from merlin.models.tf.predictions.base import ContrastivePredictionBlock
+from merlin.models.tf.predictions.sampling.base import Items, ItemSampler, ItemSamplersType
+from merlin.models.tf.typing import TabularData
+from merlin.models.tf.utils import tf_utils
+from merlin.models.tf.utils.tf_utils import call_layer, rescore_false_negatives
+from merlin.models.utils import schema_utils
+from merlin.models.utils.constants import MIN_FLOAT
+from merlin.schema import Schema, Tags
+
+LOG = logging.getLogger("merlin_models")
```
oliverholworthy, line 18: 🗒️  this doesn't need to be addressed in this PR. This is an observation that we should address the inconsistency we have in the name we used inside `getLogger` across the repo:

https://github.com/NVIDIA-Merlin/models/search?p=1&q=getLogger

merlin/models/tf/predictions/sampling/base.py
```diff
@@ -0,0 +1,114 @@
+import abc
```
oliverholworthy, line 1: missing license from the top of this file, + some of the other new files added

merlin/models/tf/predictions/dot_product.py
sararb: Thanks for the catch! I just opened a new issue for tracking

merlin/models/tf/predictions/sampling/base.py
sararb: I have just updated the files with the missing license

merlin/models/tf/predictions/sampling/in_batch.py
sararb: @marcromeyn added this code line, so I kept it until he returned to find out why it is needed.

merlin/models/tf/predictions/dot_product.py
sararb: Thanks for the catch!  I just removed the # TODO as the `add_sampler` is now implemented and tested.

oliverholworthy (approved):

merlin/models/tf/predictions/sampling/in_batch.py
sararb: Thanks, that's a good point! Once all tasks defined in this [ticket](https://github.com/NVIDIA-Merlin/Merlin/issues/464) are done, the next step is to do the benchmarking of the retrieval/ranking models and then we change the high-level API. The main constraint when implementing these tasks is to keep the user-facing API the same. 

But in order to simplify things, the item retrieval model will be converted into a Top-k recommender and Top-k evaluation will be implemented differently. The overall objective is to build a `ItemRecommenderModel` class that supports the base methods `predict()` for evaluation and `.save()`  for serialization. [Here](https://github.com/NVIDIA-Merlin/models/compare/main...tf/retrieval-models) is a draft code that summarizes what Marc and I discussed.

PR #595, opened 2022-07-20
Trying to make model unit-tests faster

marcromeyn:
### Goals :soccer:
Our unit-test has become slow (typically over 20 min). This PR tries to speed things up and get things well under 10 minutes. 

- Caching of pip-installs
- Reducing model + schema sizes in all tests
- Reducing number of epochs + steps.

### DIFF
--- a/.github/workflows/datasets.yml
+++ b/.github/workflows/datasets.yml
@@ -21,7 +21,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
           cache: 'pip'
-          cache-dependency-path: '**/**.txt'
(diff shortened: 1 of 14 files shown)


gabrielspmoreira:
Hey @marcromeyn . Not sure if it is related but we have added a few weeks ago this [auto fixture](https://github.com/NVIDIA-Merlin/models/blob/7f6170ca88848668ea5f54ca029e58f686163425/tests/unit/tf/_conftest.py#L29) to TF that clears the Keras backend before each test, as there was some shared state in TF graph among tests. Not sure if it affects tests performance, but it might be worth checking it.

gabrielspmoreira (approved):

PR #610, opened 2022-07-27
Use labels instead of predictions in XGBoost eval data

### LINKED ISSUE #598: [BUG] `XGBoost` evaluation is calculated against its predictions and not ground truth resulting in a perfect score
### Bug description

In `merlin/models/xgb/__init__.py` on line `188` we are creating a `DMatrix` to be used for evaluation, but are setting `label` to the predictions of the model instead of to ground truth.

Subsequently, we pass the `DMatrix` to `self.booster.eval(eval_data)` which predicts on the passed data and calculates metrics using predictions and `label` stored in the `DMatrix`.

image

image

```
from merlin.core.utils import Distributed
import sklearn.datasets
import cudf
import numpy as np
import nvtabular as nvt

rows = 200
num_features = 16
X, y = sklearn.datasets.make_regression(
    n_samples=rows,
    n_features=num_features,
    n_informative=num_features // 3,
    random_state=0,
)

feature_names = [str(i) for i in range(num_features)]
df = cudf.DataFrame(np.hstack((X, y.reshape(-1, 1))), columns=feature_names+['target'], dtype=np.float32)
train_ds = nvt.Dataset(df.iloc[:160])
valid_ds = nvt.Dataset(df.iloc[40:])

with Distributed():
    model = XGBoost(schema=train_ds.schema, target_columns='target')
    model.fit(
        train_ds,
        evals=[(valid_ds, 'validation_set'),]
    )

with Distributed():
    out = model.evaluate(valid_ds)

out
```

### Expected behavior
Metrics are calculated against ground truth and not model predictions

karlhigley:
Fixes #598

### DIFF
--- a/merlin/models/xgb/__init__.py
+++ b/merlin/models/xgb/__init__.py
@@ -175,17 +175,15 @@ class XGBoost:
         if self.booster is None:
             raise ValueError("The fit method must be called before evaluate.")
 
-        X, _, qid = dataset_to_xy(
+        X, y, qid = dataset_to_xy(
             dataset, self.feature_columns, self.target_columns, self.qid_column
         )
-        data = xgb.dask.DaskDMatrix(self.dask_client, X, qid=qid)
-        preds = xgb.dask.predict(self.dask_client, self.booster, data, **predict_kwargs)
 
         # convert to DMatrix
         # (eval doesn't have dask support currently)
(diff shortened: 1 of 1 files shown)


radekosmulski (approved):
Looks great! Thank you very much for the fix, @karlhigley!

Issue #613, opened 2022-08-01
[Task] Update DLRM to use the new EmbeddingTable


gabrielspmoreira:
### Description
Currently DLRM is using a custom input block and the previous `EmbeddingFeatures`. 
This task is about replacing `EmbeddingFeatures` by the new `EmbeddingTable` and also analyzing if DLRM can use the `InputBlockV2` instead of its custom approach to represent features.
Associated to https://github.com/NVIDIA-Merlin/Merlin/issues/479, which you can check for more context

oliverholworthy:
- PR #620 demonstrates how `InputBlockV2` can be used inside the `DLRMBlock` (the existing tests pass, note that the layer names and structure has changed slightly with this change). However, the way it stands, I think it makes the code less clear. And I'm wondering if that may be true in more cases where we are considering replacing usage of the InputBlock with InputBlockV2.

oliverholworthy:
- PR #619 demonstrates how the `EmbeddingFeatures` can be replaced with the `Embeddings` in the `DLRMBlock` (with existing tests passing, note that there is a change to the structure of the layers here). 
  - As it stands, this doesn't make the code any clearer. And seems to remove some functionality - with this change, the intializers and l2_reg from EmbeddingOptions are ignored.

sararb:
Thanks, @oliverholworthy and @gabrielspmoreira for the refactor of DLRM with the updated Embeddings API.  There is an example [notebook](https://github.com/NVIDIA-Merlin/models/blob/main/examples/06-Define-your-own-architecture-with-Merlin-Models.ipynb) that shows how to build the DLRM model from scratch. Should we include the update of this notebook in this ticket?

gabrielspmoreira:
Fixed by #619

PR #619, opened 2022-08-04
Enable customization of embeddings in DLRM Model

oliverholworthy:
Part of #613

### Goals :soccer:

- Allow overriding of categorical embeddings block. 
- Removing usage of `EmbeddingFeatures` in DLRM implementation. We would like to deprecate this block in favor of the simpler `Embeddings` block

### Implementation Details :construction:

- Adding new option argument `embeddings` to `DLRMBlock` and `DLRMModel`. An alternative to the `emedding_options`.
- Replacing use of `EmbeddingFeatures` with `Embeddings`

### Testing Details :mag:

- Added tests for new `embeddngs` parameter used in DLRMBlock.
- Checking that this is not used at the same time as `embedding_options`

PR #620, opened 2022-08-04
Update `DLRMBLock` to use `InputBlockV2`

oliverholworthy:
Part of #613 

### Goals :soccer:

Demonstrate how `InputBlockV2` can be used inside the `DLRMBlock`

Replace existing construction of the ParallelBlock with the InputBlockV2

### DIFF
--- a/merlin/models/tf/blocks/dlrm.py
+++ b/merlin/models/tf/blocks/dlrm.py
@@ -18,8 +18,8 @@ from typing import Optional
 
 from merlin.models.tf.blocks.interaction import DotProductInteraction
 from merlin.models.tf.core.base import Block, Debug
-from merlin.models.tf.core.combinators import Filter, ParallelBlock, SequentialBlock
-from merlin.models.tf.inputs.continuous import ContinuousFeatures
+from merlin.models.tf.core.combinators import Filter, SequentialBlock
+from merlin.models.tf.inputs.base import InputBlockV2
 from merlin.models.tf.inputs.embedding import EmbeddingFeatures, EmbeddingOptions
 from merlin.schema import Schema, Tags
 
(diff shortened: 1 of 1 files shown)


merlin/models/tf/blocks/dlrm.py
```diff
@@ -108,9 +108,12 @@ def DLRMBlock(
                 "The bottom_block is required by DLRM when "
                 "continuous features are available in the schema"
             )
-        con = ContinuousFeatures.from_schema(con_schema)
-        bottom_block = con.connect(bottom_block)  # type: ignore
-        interaction_inputs = ParallelBlock({"embeddings": embeddings, "bottom_block": bottom_block})
+        interaction_inputs = InputBlockV2(
```
oliverholworthy, line 111: It may be the case (demonstrated here) that use of the `InputBlockV2` doesn't make the code any clearer or simpler. I would argue that in this case it would be clearer to keep the construction of the ParallelBlock as it exists already.

gabrielspmoreira: I agree @oliverholworthy. The code without `InputBlockV2` is indeed clearer. 
I think the main benefit objective here is to use the new `EmbeddingTable`, so maybe we should replace the `EmbeddingFeatures.from_schema()` by `Embeddings(schema)`.
What do you think?

gabrielspmoreira: Just like you do [here](https://github.com/NVIDIA-Merlin/models/pull/619) :)

oliverholworthy:
Closing this PR, since it doesn't add any value at this stage to use `InputBlockV2` in the implementation of DLRM

PR #623, opened 2022-08-05
Add Wide Deep model

timmy00:
### Goals :soccer:
Add Wide & Deep Model into ranking.py
Refactor previous PR of Add Wide and Deep model into benchmark  #502 
Replace InputBlock by InputBlockV2

Arguments of wide_schema and deep_schema are used to specify features for wide model and for deep model, for now it is only suitable for binary classification, which is similar to [tf.estimator.DNNLinearCombinedClassifier](https://www.tensorflow.org/api_docs/python/tf/estimator/DNNLinearCombinedClassifier)

### Testing Details :mag:
Tests include different preprocess block for wide model:
test_wide_deep_model: No preprocess
test_wide_deep_model_categorical_one_hot: CategoricalOneHot as wide_preprocess
test_wide_deep_model_hashed_cross: HashedCross as wide_preprocess
test_wide_deep_embedding_custom_input:  test custom embeddings and inputblock for deep model


merlin/models/tf/models/ranking.py
```diff
@@ -235,3 +236,131 @@ def DeepFMModel(
...
+        if len(wide_schema) > 0:
+            wide_input_block = ParallelBlock(
+                TabularBlock.from_schema(schema=wide_schema, pre=wide_preprocess),
+                is_input=True,
+                aggregation="concat",
+            )
+    wide_body = wide_input_block.connect(MLPBlock([1], no_activation_last_layer=True))
+
+    branches = {"wide": wide_body, "deep": deep_body}
+    wide_and_deep_body = ParallelBlock(branches, aggregation="element-wise-sum")
```
gabrielspmoreira, line 363: In the W&D paper they say "The wide component and deep component are combined using a weighted sum of their output log odds as the prediction". But I have checked the [TF implementation](tf.estimator.DNNLinearCombinedClassifier ) and they just sum the logits of both heads, so I think we are fine here

gabrielspmoreira (approved):
@timmy00 Thanks for this PR! I like the fact that you have sliced the W&D architecture in many reusable building blocks, introduced in other PRs. I like the flexibility provided by the args `wide_preprocess` (e.g. `CategoricalOneHot`, `HashedCross`) and `deep_input_block`, which allows defining advanced embeddings options of needed.
The tests and docstrings for the new building blocks are great.
I just think we should add a test showcasing the standard W&D architecture from the paper, including the [FTRL](https://github.com/tensorflow/estimator/blob/eb9d58bf952fbeb9fd4e686b7528d6adf0021320/tensorflow_estimator/python/estimator/canned/optimizers.py#L104) for wide part and AdaGrad for deep part.

timmy00:
> @timmy00 Thanks for this PR! I like the fact that you have sliced the W&D architecture in many reusable building blocks, introduced in other PRs. I like the flexibility provided by the args `wide_preprocess` (e.g. `CategoricalOneHot`, `HashedCross`) and `deep_input_block`, which allows defining advanced embeddings options of needed. The tests and docstrings for the new building blocks are great. I just think we should add a test showcasing the standard W&D architecture from the paper, including the [FTRL](https://github.com/tensorflow/estimator/blob/eb9d58bf952fbeb9fd4e686b7528d6adf0021320/tensorflow_estimator/python/estimator/canned/optimizers.py#L104) for wide part and AdaGrad for deep part.

Hi @gabrielspmoreira , thank you for your review and suggestion, I agree with you that we could add an example showcasing the standard W&D architecture from the paper, including the FTRL for wide part and AdaGrad for deep part. But I need to merge another PR at first. 
Because when user create the model, they do not create the wide model and deep model by themselves, so we need to find a way to let them get the blocks (wide model and deep model), I implemented another method in Model base: find_blocks_by_name, and we can get wide model and deep model by name, then assign different optimizers to them:

PR #625, opened 2022-08-05
CategoricalPrediction block with negatives support

### LINKED ISSUE #393: Improve the design of negative-sampling
We want to improve the user experience of applying negative sampling + make it more generic to make it work for ranking use-cases as well.

In [Cross-Batch Negative Sampling for Training Two-Tower Recommenders](https://arxiv.org/abs/2110.15154) they describe a strategy where they start with in-batch negative sampling and then add cross-batch after _n_ steps. This is tricky to do in our current API since negative-sampling is part of the model. 

We are proposing a new type of block to make this easier: `PreLossBlock`. Since this is only used in the training/evaluation procedure (just like losses & metrics) we propose to make this part of `model.compile`. The API would then become as follows:

**Retrieval case:**
```python
model = TwoTowerModel(schema, MLPBlock([512, 128]))
model.compile(pre_loss=ContrastiveLearning(schema, "in-batch"))
model.save(save_pre_loss=False)   # user can choose whether or not to save the pre_loss

# Model can easily be converted to a youtube-dnn style retrieval model
# The model then only takes in user-features and holds all item-embeddings in an embedding-table
new_model = model.to_item_recommender(train)
model.compile(pre_loss=ContrastiveLearning(schema, "in-batch", "cross-batch"))
```

**Ranking case:**
```python
pred_task = NextItemPredictionTask(schema)
model = DLRMModel(schema, 128, MLPBlock([512, 128]), prediction_tasks=pred_task)
model.compile(pre_loss=ContrastiveLearning(schema, "popularity-based"))
```

sararb:
Fixes #393

### Goals :soccer:

- Refactor the prediction block related to multi-classification and next item prediction task. 
- The new prediction block also supports contrastive learning for sampled softmax support.  

- [x] Add CategoricalPrediction layer : This new class allows the definition of multi-classification tasks from a schema, with either a dense layer or an embedding table as an output layer (weight-tying use-case).

- [x] The class also supports negative sampling strategies (also known as sampled softmax) during training/evaluation to obtain top-k logits over a small subset of vocabulary. 

- [x] The negative sampling is handled by the new class `ContrastiveLookUps`. From a list of samples, it samples items ids and looks up their weights or embeddings. 

- [x] Weight-tying is handled via the new class EmbeddingTablePrediction. The logits are computed by multiplying the embedding weights by the hidden representations given by the neural network (`inputs`). 

- [x] Add a popularity-based sampler. It uses tf.random.log_uniform_candidate_sampler to approximate item probabilities using log_uniform (zipfisan).  The only requirement is that the item ids are decreasingly sorted by their count frequency (which is the default in the NVTabular `Categorify` op).

### Testing Details :mag:

- Testing new PopularityBasedSamplerV2
- Testing model with multi-classification task 
- Testing model with next item prediction task 
- Testing setting different negative sampling strategies: None, PopularitySampler,  PopularitySampler + In-batch sampling


sararb:
rerun tests

sararb:
rerun tests

sararb:
close this PR to create a new one because Jenkins tests are re-running indefinitely

PR #631, opened 2022-08-09
Enable `TabularData` to be used with `EmbeddingTable`

oliverholworthy:
Relates to https://github.com/NVIDIA-Merlin/Merlin/issues/479

### Goals :soccer:

The `EmbeddingTable` layer has partial support for TabularData inputs and masking. (It was written with the intent to support both. This PR is a follow-up to add further implementation for these features.

- Enable `EmbeddingTable` layer to be used with TabularData inputs (`Dict[str, Tensor]`).
- Enable `EmbeddingTable` layer to be used with masking

- Trying to use `EmbeddingTable` in contexts where a dict of tensors is passed currently throws an exception from the `compute_output_shape` method. Extending support for this input format.
- Setting `supports_masking = True` in the constructor of `EmbeddingTable` to 

### Testing Details :mag:

Added tests for `compute_output_shape` with different inputs.

### DIFF
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -190,6 +190,7 @@ class EmbeddingTable(EmbeddingTableBase):
                 **table_kwargs,
             )
         self.combiner = combiner
+        self.supports_masking = True
 
     @classmethod
     def from_pretrained(
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -299,7 +299,12 @@ class EmbeddingTable(EmbeddingTableBase):
 
         return out
 
-    def compute_output_shape(self, input_shape):
+    def compute_output_shape(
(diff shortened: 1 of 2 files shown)


merlin/models/tf/inputs/embedding.py
```diff
@@ -190,6 +190,7 @@ def __init__(
                 **table_kwargs,
             )
         self.combiner = combiner
+        self.supports_masking = True
```
gabrielspmoreira, line 193: It seems that `supports_masking` is currently only used by `layer_test()`, as masking was removed before.
Not sure if the new solution for masking (when we add session/sequential based support will use `support_masking`

gabrielspmoreira (approved):

merlin/models/tf/inputs/embedding.py
oliverholworthy: According to the keras docs, the Embedding layer supports masking, so I put this in to maintain the functionality of the `Embedding` layer. However, we don't have any tests that use the masking with this `EmbeddingTable` wrapper.

PR #632, opened 2022-08-09
Enable static weights to be used with `EmbeddingTable.from_pretrained`

oliverholworthy:
Relates to https://github.com/NVIDIA-Merlin/Merlin/issues/479

### Goals :soccer:

Enable static weights to be used with the `EmbeddingTable.from_pretrained` method

The current implementation initializes the embedings using the `weights` parameter.

However, this only works if the layer is called first before the model is compiled/fit.

Switching to use the `embeddings_initializer` parameter instead which works as expected with trainable False.

### Testing Details :mag:

Removing part of the test for the from_pretrained method to capture the behaviour of the embeddings correctly.

### DIFF
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -180,6 +180,7 @@ class EmbeddingTable(EmbeddingTableBase):
                 embeddings_constraint=embeddings_constraint,
                 mask_zero=mask_zero,
                 input_length=input_length,
+                trainable=trainable,
             )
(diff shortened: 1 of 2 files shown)


gabrielspmoreira (approved):
This is good, as setting `EmbeddingTable(trainable=False)` before was not really setting trainable of `Embedding.trainable`.

rnyak:
rerun tests

rnyak (approved):

rnyak:
rerun tests

Issue #627, opened 2022-08-05
The `Embeddings` function is very handy to infer the embeddings definition from the schema. Make it more flexible by adding a `trainable: Optional[List[str]] = None` argument, to set `EmbeddingTable.trainable=False` (default is True) for the features present in the list.


viswa-nvidia:

gabrielspmoreira:
Fixed in #634

PR #634, opened 2022-08-09
Add `trainable` and `embeddings_initializers` args to `Embeddings` function

rnyak:
This PR is addressing  https://github.com/NVIDIA-Merlin/models/issues/627


merlin/models/tf/inputs/embedding.py
```diff
@@ -347,6 +347,10 @@ def Embeddings(
     infer_embedding_sizes: bool = True,
     infer_embedding_sizes_multiplier: float = 2.0,
     infer_embeddings_ensure_dim_multiple_of_8: bool = True,
+    non_trainable: List[str] = [],
```
oliverholworthy, line 350: Wondering if it would be more familiar to be called `trainable` to be consistent with the layers?

Passing as a `Dict[str, bool]`. For example `trainable = {"col_a": False}`

merlin/models/tf/inputs/embedding.py
```diff
@@ -414,7 +426,14 @@ def Embeddings(
                 combiner = sequence_combiner
 
         embedding_size = embedding_dims.get(col.name, embedding_dim_default)
-        tables[col.name] = EmbeddingTable(embedding_size, col, combiner=combiner, **kwargs)
+        tables[col.name] = EmbeddingTable(
+            embedding_size,
+            col,
+            combiner=combiner,
+            trainable=col not in non_trainable,
+            embeddings_initializer=embeddings_initializers.get(col.name, "uniform"),
```
oliverholworthy, line 434: Need to handle the case where embeddings_initializers is None or not a dict. Similar to this part in the EmbeddingFeatures

https://github.com/NVIDIA-Merlin/models/blob/c63daf65efc653cd84cc32ba64c8ce8f110af220/merlin/models/tf/inputs/embedding.py#L629-L632

merlin/models/tf/inputs/embedding.py
rnyak: makes sense.

merlin/models/tf/inputs/embedding.py
rnyak: makes sense.

merlin/models/tf/inputs/embedding.py
```diff
@@ -381,6 +385,14 @@ def Embeddings(
     infer_embeddings_ensure_dim_multiple_of_8 : bool, optional
         Ensures that the inferred embedding sizes are rounded up to the next multiple of 8,
         for better performance for embedding looks on GPUs. By default True
+    non_trainable : Optional[Dict[str, bool]] = None
+        Name of the column(s) whose embeddings should be frozen during training
```
oliverholworthy, line 389: Docstring out of date with type of arg

merlin/models/tf/inputs/embedding.py
```diff
@@ -381,6 +385,14 @@ def Embeddings(
     infer_embeddings_ensure_dim_multiple_of_8 : bool, optional
         Ensures that the inferred embedding sizes are rounded up to the next multiple of 8,
         for better performance for embedding looks on GPUs. By default True
+    non_trainable : Optional[Dict[str, bool]] = None
```
oliverholworthy, line 388: Removing _non_ prefix for consistency based on previous comment I think worth doing now that it's a Dict.

oliverholworthy, line 388: Removing _non_ prefix for consistency based on previous comment I think worth doing now that it's a Dict.

merlin/models/tf/inputs/embedding.py
```diff
@@ -414,7 +426,21 @@ def Embeddings(
                 combiner = sequence_combiner
 
         embedding_size = embedding_dims.get(col.name, embedding_dim_default)
-        tables[col.name] = EmbeddingTable(embedding_size, col, combiner=combiner, **kwargs)
+        initializer_default = "uniform"
+        if not embeddings_initializers:
+            emb_initializer = initializer_default
+        elif isinstance(embeddings_initializers, dict):
+            emb_initializer = embeddings_initializers.get(col.name, "uniform")
```
oliverholworthy, line 433: Could reuse the default variable here in the .get

oliverholworthy (approved):

gabrielspmoreira (commented):
@rnyak I think we should create a test showcasing the typical use case where we want to initialize an embedding table with pre-trained embeddings and set them to non-trainable, so that we can both automate testing of that functionality and use it as an example for our customers.

gabrielspmoreira:
Test was added in #648

PR #635, opened 2022-08-10
Add Layer freezing 

timmy00:
### Goals :soccer:
Add freeze_blocks in Model class. Freezing a layer prevents its weights from being modified. This technique is often used in transfer learning. This feature is similar to [keras Transfer learning & fine-tuning](https://keras.io/guides/transfer_learning/), which allow users to freeze blocks (layers) and unfreeze.

Example:
```            input_block = ml.InputBlockV2(schema)
            layer_1 = ml.MLPBlock([64], name="layer_1")
            layer_2 = ml.MLPBlock([1], no_activation_last_layer=True, name="layer_2")
            two_layer = ml.SequentialBlock([layer_1, layer_2], name="two_layers")
            body = input_block.connect(two_layer)
            model = ml.Model(body, ml.BinaryClassificationTask("click"))

            # Compile(Make sure set run_eagerly mode) and fit -> model.freeze_blocks -> compile and
            # fit Set run_eagerly=True in order to avoid error: "Called a function referencing
            # variables which have been deleted". Model needs to be built by fit or build.

            model.compile(run_eagerly=True, optimizer=tf.keras.optimizers.SGD(lr=0.1))
            model.fit(ecommerce_data, batch_size=128, epochs=1)

            model.freeze_blocks(["user_categories", "layer_2"])
            # From the result of model.summary(), you can find which block is frozen (trainable: N)
            print(model.summary(expand_nested=True, show_trainable=True, line_length=80))

            model.compile(run_eagerly=False, optimizer="adam")
            model.fit(ecommerce_data, batch_size=128, epochs=10)
```

Users may only call one line to create a model, but they can `print(model.summary(expand_nested=True))` to find the names of each block, (you can find the example of `model.summary` 's output in `test_base.py` ) then they could use `model.freeze_blocks(["user_categories", "layer_2"])` to freeze any layers they want, in this example the embedding table for `user_categories` and the MLP block with name `layer_2` are frozen.

Include 5 methods and a property added in Model calss (merlin/models/tf/models/base.py)

1. Freeze all sub-blocks of given blocks recursively. `freeze_blocks(self, blocks: Union[Sequence[Block], Sequence[str]] )`
2. Unfreeze all sub-blocks of given blocks recursively. `unfreeze_blocks(self, blocks: Union[Sequence[Block], Sequence[str]] )`
3. Unfreeze all blocks (including blocks and sub-blocks) of this model recursively. `unfreeze_all_frozen_blocks(self)`
4. Get blocks by given block_names, return a list of blocks. `find_blocks_by_name`
5. Get all sub-blocks of given blocks, including blocks themselves. `get_sub_blocks(self, blocks: Sequence[Block]) -> List[Block]`
6. Get frozen blocks of model.
```@property
    def frozen_blocks(self):
```
        

### Testing Details :mag:

1. test_find_blocks_and_sub_blocks
2. test_freeze_parallel_block
3. test_freeze_sequential_block
4. test_wide_deep_model
5. test_freeze_unfreeze
6. test_unfreeze_all_blocks

### TODO
Currently works well with run_eagerly=True, but for graph mode (run_eagerly=False), when re-compile and fit, it would raise the same issue as https://github.com/NVIDIA-Merlin/models/issues/647.
TODO: fix it for graph mode (run_eagerly = False), please find example test_freeze_parallel_block in tests/unit/tf/models/test_base.py


merlin/models/tf/models/base.py
```diff
@@ -1025,6 +1028,148 @@ def _set_save_spec(self, inputs, args=None, kwargs=None):
...
+                    include_self=False, recursive=False
+                ):
+                    # Filter out modelcontext
+                    if type(sub_module) != ModelContext:
+                        deque.append(sub_module)
+            if len(block_names) > 0:
+                raise ValueError(f"Cannot find block with the name of {block_names}")
+        return list(result_blocks)
+
+    def get_sub_blocks(self, blocks: Sequence[Block]) -> List[Block]:
```
marcromeyn, line 1150: I think it would be better to add this method to the utils, and not make it part of the `Model`. This since it seems useful outside the model as well.

merlin/models/tf/models/base.py
```diff
@@ -1025,6 +1028,148 @@ def _set_save_spec(self, inputs, args=None, kwargs=None):
...
+                self._frozen_blocks.remove(b)
+            b.trainable = True
+
+    def unfreeze_all_frozen_blocks(self):
+        """Unfreeze all blocks (including blocks and sub-blocks) of this model recursively"""
+        for b in self._frozen_blocks:
+            b.trainable = True
+        self._frozen_blocks = set()
+
+    def find_blocks_by_name(self, block_names: Sequence[str]) -> List[Block]:
```
marcromeyn, line 1119:
```suggestion
    def get_blocks_by_name(self, block_names: Sequence[str]) -> List[Block]:
```

merlin/models/tf/core/combinators.py
```diff
@@ -239,9 +239,9 @@ def trainable(self):
         Returns
         -------
         bool
-            True if all layer within SequentialBlock are trainable, otherwise False
+            True if any layer within SequentialBlock are trainable, otherwise False
         """
-        return all(layer.trainable for layer in self.layers)
+        return any(layer.trainable for layer in self.layers)
```
marcromeyn, line 244: What the reasoning behind this change? `trainable` is a standard keras-method, and if it returns `any` I am worried it will have some unforeseen side-effects.

rnyak:
blocked by https://github.com/NVIDIA-Merlin/models/issues/647

timmy00:
> blocked by #647

Currently works well with run_eagerly=True, for graph mode (run_eagerly=False), when re-compile and fit, it would raise the same issue as #647.
TODO: fix it for graph mode (run_eagerly = False)

merlin/models/tf/core/combinators.py
timmy00: There are many `ParallelBlock`s in a model, and sometime we would freeze some children blocks of it, so parts of its children blocks have `trainable=True` and other children blocks have `trainable=False`, if we keep `all` here, then the entire `ParallelBlock` would have `trainable` as False, it is not we want, we still want to change the unfrozen children blocks.

marcromeyn (approved):

PR #637, opened 2022-08-10
Update stacking in `DLRMBlock` to use correct axis for the dot product.

oliverholworthy:
### Goals :soccer:

Correct the implementation of the dot product component of the DLRM Layer.

While adding a test in #619
Noticed that the output dimension didn't feel right.

The stacking aggregation is using axis -1 (the default in `StackFeatures`).

Which results in the `DotProductInteraction` receiving inputs with shape `(batch_size, embedding_dim, num_features)` instead of the expected shape `(batch_size, num_features, embedding_dim)`.

This changes the axis to 1 so that the embedding dim comes last.

### Testing Details :mag:

Updated existing tests for the expected output shape of the DLRMBlock.

### DIFF
--- a/merlin/models/tf/blocks/dlrm.py
+++ b/merlin/models/tf/blocks/dlrm.py
@@ -17,6 +17,7 @@
 from typing import Optional
 
 from merlin.models.tf.blocks.interaction import DotProductInteraction
+from merlin.models.tf.core.aggregation import StackFeatures
 from merlin.models.tf.core.base import Block, Debug
 from merlin.models.tf.core.combinators import Filter, ParallelBlock, SequentialBlock
 from merlin.models.tf.inputs.continuous import ContinuousFeatures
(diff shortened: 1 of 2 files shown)


gabrielspmoreira:
Good catch @oliverholworthy! I think your fix is correct as according to the [DLRM paper](https://arxiv.org/pdf/1906.00091.pdf)...

```
We will compute second-order interaction of different features explicitly, following the intuition for
handling sparse data provided in FMs (4), optionally passing them through MLPs. This is done by
taking the dot product between all pairs of embedding vectors and processed dense features. 
```

I debugged the `test_dlrm_block()` and it seems correct now, as the feature interaction between 5 vectors (4 embeddings + bottom MLP output) results into 10 values (the number of pairwise combinations with 5 elements) instead of 2016 values as before.

I think we can merge it!

gabrielspmoreira (approved):

Issue #641, opened 2022-08-11
[BUG] ValueError for Two-tower model with InputBlockV2 as input


timmy00:
### Bug description
Error when create TwoTowerModel with InputBlockV2 as input.

### Steps/Code to reproduce bug
```
    @pytest.mark.parametrize("run_eagerly", [True, False])
    def test_two_tower_with_custom_input_block(ecommerce_data: Dataset, run_eagerly):
        schema = ecommerce_data.schema.select_by_name(
            ["user_brands", "user_categories", "item_brand",  "item_id"])
        item_embeddings = mm.Embeddings(schema.select_by_tag(Tags.ITEM),infer_embedding_sizes=True)
        query_embeddings = mm.Embeddings(schema.select_by_tag(Tags.USER),infer_embedding_sizes=True)
>       model = mm.TwoTowerModel(schema,
                                query_tower=mm.MLPBlock([2]),
                                item_tower_input=mm.InputBlockV2(schema.select_by_tag(Tags.ITEM),embeddings=item_embeddings),
                                query_tower_input=mm.InputBlockV2(schema.select_by_tag(Tags.USER),embeddings=query_embeddings)
                                )

tests/unit/tf/models/test_retrieval.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
merlin/models/tf/models/retrieval.py:186: in TwoTowerModel
    two_tower = TwoTowerBlock(
merlin/models/tf/blocks/retrieval/two_tower.py:122: in __init__
    super().__init__(query_tower, _item_tower, post=post, **kwargs)
merlin/models/tf/blocks/retrieval/base.py:104: in __init__
    "query": Filter(query_block.schema).connect(self._query_block),
[...]

Issue #604, opened 2022-07-25
[Task] InputBlockV2 - Usage in ranking models high-level API


gabrielspmoreira:
Change the following higher-level API ranking models to use the new `InputBlockV2`.
This is part of RMP #479, which you can check for more context

- [x] `DCNModel`
- [x] `DeepFMModel`

viswa-nvidia:
@gabrielspmoreira , please follow the template and add the problem statement, goal, constraints and starting point.

PR #644, opened 2022-08-11
Allow custom input block for two tower and fix [Bug] InputBlockV2 without schema 

timmy00:
Fixes # (issue)
#641 

### Goals :soccer:
Fix InputBlockV2 without schema #641 

1. Change the schema setting in ParallelBlock, in previous development, only if all sub blocks have schema, the ParallelBlock (parent of sub blocks) would be set schema. I changed it to that if any sub block has schema,  the ParallelBlock (parent of sub blocks) would be set to the collection of sub-blocks' schemas.
2. InputBlockV2 dose not have schema. Thus adding set_schema for InputBlockV2
3. Allow flexibility for TwoTower API, allow users create custom input block as item/query tower input block

### Testing Details :mag:
Test custom input block (InputBlockV2) in two-tower model


gabrielspmoreira:
@timmy00  This PR is related to these issues
https://github.com/NVIDIA-Merlin/Merlin/issues/479
https://github.com/NVIDIA-Merlin/models/issues/604

gabrielspmoreira:
rerun tests

gabrielspmoreira (approved):
This PR looks good to me!

marcromeyn (commented):
Let’s hold off this one, seems like we should first look into the root-cause of the bug a bit more.

timmy00:
Fixed by #649 .

PR #649, opened 2022-08-15
Restructuring Embeddings

marcromeyn:
Supports https://github.com/NVIDIA-Merlin/Merlin/issues/479

### Goals :soccer:
This PR makes some improvements to `Embeddings` & `EmbeddingTable`:

- names of arguments to both now match for consistency. And they can all be set with either a fixed value to be used for all embedding tables `T` or a dict mapping column name to different values `Dict[str, T]`
- Remove the need to wrap the embeddings inside a sequential-block with a `AsRaggedFeatures`. This is done by allowing a list-column represented as a tuple to be passed to a `EmbeddingTable`.
- Break out some helper-functions to improve code readability of `Embeddings`.
- Simplify the way we infer embedding-dims
  - A function may now be passed as `infer_dim_fn`. This function is passed a ColumnSchema corresponding to the column in the schema each embedding table is being constructed for.


oliverholworthy:
If we want the ragged tensor support for the EmbeddingTable in 22.08. perhaps could pull that out into a separate PR from the change to the `Embedding` function.

merlin/models/tf/inputs/embedding.py
```diff
@@ -335,120 +343,115 @@ def get_config(self):
...
     ] = "mean",
-    embedding_dims: Optional[Dict[str, int]] = None,
-    embedding_dim_default: Optional[int] = 32,
-    infer_embedding_sizes: bool = True,
-    infer_embedding_sizes_multiplier: float = 2.0,
-    infer_embeddings_ensure_dim_multiple_of_8: bool = True,
-    trainable: Optional[Dict[str, bool]] = None,
     embeddings_initializers: Optional[
-        Union[Dict[str, Callable[[Any], None]], Callable[[Any], None]]
+        Union[Dict[str, Callable[[Any], None]], Callable[[Any], None], str]
```
oliverholworthy, line 353: I think the type of of the initializers in `EmbeddingOptions` needs to be updated to match this for consistency.

tests/unit/tf/inputs/test_tabular.py
```diff
@@ -220,12 +226,9 @@ def test_tabular_seq_features_avg_embeddings_with_mapvalues(sequence_testing_dat
 def test_embedding_tables_from_schema_infer_dims(sequence_testing_data: Dataset, aggregation: str):
     cat_schema = sequence_testing_data.schema.select_by_tag(Tags.CATEGORICAL)
     embeddings_block = ml.Embeddings(
-        cat_schema,
+        cat_schema.select_by_tag(Tags.CATEGORICAL),
         embedding_dims={"item_id_seq": 15, "test_user_id": 21},
-        infer_embedding_sizes=True,
-        infer_embedding_sizes_multiplier=2.0,
-        infer_embeddings_ensure_dim_multiple_of_8=True,
-        embeddings_initializer="truncated_normal",
```
oliverholworthy, line 228: If I'm reading this correctly. This was working as before because **kwargs are passed down to `EmbeddingTable`. 

It looks like this should now fail if passed and will complain about multiple params with the same name passed to `EmbeddingTable`.

Are there any params to **kwargs would be useful to pass to EmbeddingTable now or can that be removed?

merlin/models/tf/inputs/embedding.py
```diff
@@ -399,25 +402,49 @@ def Embeddings(
...
-            _get_dim(col, embedding_dims, infer_dim_fn),
-            col,
-            combiner=_get_combiner(col, sequence_combiner),
-            embeddings_initializer=_get_initializer(col, embeddings_initializers),
-            trainable=trainable.get(col.name, True),
-            **kwargs,
-        )
+        table_kwargs = _forward_kwargs_to_table(col, table_cls, kwargs)
+        if sequence_combiner:
+            table_kwargs["combiner"] = _get_combiner(col, sequence_combiner)
```
oliverholworthy, line 425: It's not clear to my why this param requires a different name in this function. If `sequence_combiner` is preferred over `combiner` it seems like that should be the name in the `EmbeddingTable` too. Or we change this to match `combiner` from the EmbeddingTable?

oliverholworthy (approved):

merlin/models/tf/inputs/embedding.py
```diff
@@ -318,135 +327,132 @@ def compute_call_output_shape(self, input_shapes):
...
+
+
+def _get_dim(col, embedding_dims, infer_dim_fn):
+    dim = None
+    if isinstance(embedding_dims, dict):
+        dim = embedding_dims.get(col.name)
+    elif isinstance(embedding_dims, int):
+        dim = embedding_dims
+
+    if not dim:
```
gabrielspmoreira, line 452: @marcromeyn Before we were able to define embedding dims to specific features and infer embeddding dims for all other features. But now it seems you can either set dims or infer them, but not both. 
That was useful for example for the retrieval experiments, so that we can hypertune the user and item dims independently (so that capacity of TwoTower and MF is comparable).

oliverholworthy: @gabrielspmoreira it should only reach this line (to infer the `dim` parameter) if it's not already specified. The difference is that all the args can now be specified as a single value (`T` - int in this case) or `Dict[str, T]`. 

Example below of specifying a dim for one column (`item_id`), leaving the rest to be inferred (`item_feature`)

```python
import pyarrow as pa
import numpy as np
import merlin.models.tf as mm
from merlin.schema import Schema, ColumnSchema, Tags

schema = Schema(
    [
        ColumnSchema(
            "item_id",
            properties={"domain": {"min": 0, "max": 10}},
            dtype=np.int32,
        ),
        ColumnSchema(
            "item_feature",
            properties={"domain": {"min": 0, "max": 10}},
            dtype=np.int32,
        ),
    ]
)


embeddings = mm.Embeddings(schema, dim={"item_id": 4})

embeddings["item_id"].table.output_dim       # => 4 (specified)
embeddings["item_feature"].table.output_dim  # => 8 (inferred)
```

PR #653, opened 2022-08-15
Breaks ties of top-k metrics

### LINKED ISSUE #617: [BUG] TopK metrics are wrong when model predicts same value for all items.
### Bug description
When the model predicts a constant value, the topk metrics scores perfect.

When the model predicts the same value for all items, the sort order doesn't change
https://github.com/NVIDIA-Merlin/models/blob/main/tests/unit/tf/metrics/test_metrics_topk.py#L53

The negative sampler creates the batches by putting the positive item on the first position and the negatives on the next ones:
https://github.com/NVIDIA-Merlin/models/blob/7f6170ca88848668ea5f54ca029e58f686163425/merlin/models/tf/blocks/retrieval/base.py#L363

In that way, we get perfect scores by predicting a constant.

Based on the unttest

```
labels = tf.convert_to_tensor([[0, 1, 0, 1, 0], [1, 0, 0, 1, 0], [0, 0, 0, 0, 1]], tf.float32)
predictions = tf.convert_to_tensor(
    [[10, 10, 10, 10, 10], [5, 5, 5, 5, 5], [6, 6, 6, 6, 6]], tf.float32
)
label_relevant_counts = tf.convert_to_tensor([2, 2, 1], tf.float32)
print(labels)
predictions, labels, label_relevant_counts = extract_topk(5, predictions, labels)
print(labels)
```

gabrielspmoreira:
Fixes #617
With this fix, we break ties of top-k metrics.
That is done by adding a small random epsilon to the predictions.
That way, we avoid issues when the score of all predictions is equal, like reported in #617

Also added tests and docstrings to `extract_topk()`

### DIFF
--- a/merlin/models/tf/metrics/topk.py
+++ b/merlin/models/tf/metrics/topk.py
@@ -257,7 +257,9 @@ class TopkMetric(Mean, TopkMetricWithLabelRelevantCountsMixin):
 
     def _maybe_sort_top_k(self, y_pred, y_true, label_relevant_counts: tf.Tensor = None):
         if not self.pre_sorted:
-            y_pred, y_true, label_relevant_counts = extract_topk(self.k, y_pred, y_true)
+            y_pred, y_true, label_relevant_counts = extract_topk(
+                self.k, y_pred, y_true, shuffle_ties=True
+            )
         else:
             if label_relevant_counts is None:
                 raise Exception(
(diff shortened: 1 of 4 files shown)


bschifferer (approved):
It seems like my workaround. I dont know, if the workaround capture all edge cases and/or if it  changed all functions, we need to.

One edge case, we probably miss is: Tensorflow predicts tf.nan values. If learning rate is pretty high, the model can collapse and predicts only tf.nan. In that case, the shuffle by adding a random dataframe doesnt work

gabrielspmoreira:
> It seems like my workaround. I dont know, if the workaround capture all edge cases and/or if it changed all functions, we need to.
> 
> One edge case, we probably miss is: Tensorflow predicts tf.nan values. If learning rate is pretty high, the model can collapse and predicts only tf.nan. In that case, the shuffle by adding a random dataframe doesnt work

Thanks for the review @bschifferer . Indeed the solution is based in your code snippet. It is somewhat aligned with the way [TF Ranking](https://github.com/tensorflow/ranking/blob/de7d7c2a8b8663a9d813a68ef036ef0c7585ab37/tensorflow_ranking/python/utils.py#L96) implements tie break - the difference there is that they shuffle the items randomly before tf.math.topk (as it is stable, keeping the order of items if they have equal value), while in this PR we just add a very small epsilon directly to prediction scores.
Regarding metrics if any prediction score is Nan, then the aggregated (averaged) metric result will be Nan. This already happens. And I think that is fine and better than having "silent" Nan prediction scores that user is not aware of.

PR #654, opened 2022-08-15
enhance pretrained-embeddings example to set trainable=False for  the pre-trained embeddings

rnyak:
This PR enhances the existing [entertainment-with-pretrained-embeddings.ipynb](https://github.com/NVIDIA-Merlin/models/blob/main/examples/usecases/entertainment-with-pretrained-embeddings.ipynb) notebook in the following ways:
- users now can set the trainable param in the Embeddings function
-  users can use InputBlockV2 and feed to the model


radekosmulski (dismissed):

radekosmulski (approved):

radekosmulski:
rerun tests

bschifferer:
Should the unittest include testing that the embedding weights did not change?

bschifferer (approved):
We might want to add a unittest to check, if the weights were freezed (comparing the embeddings after training with the original numpy array)

rnyak:
> We might want to add a unittest to check, if the weights were freezed (comparing the embeddings after training with the original numpy array)

we do have a unit test [here](https://github.com/NVIDIA-Merlin/models/blob/main/tests/unit/tf/inputs/test_embedding.py#L219) if that' what you meant.

rnyak:
rerun tests

rnyak:
rerun tests

PR #663, opened 2022-08-18
Top-k recommender model

### LINKED ISSUE #622: [Task] Simplify Top-k recommender API
## Problem 
- Generating the top-k prediction from a trained the retrieval model currently requires the user to follow these steps (Not straightforward). Besides, the top-k recommender is a ModelBlock that cannot be saved / re-loaded. 
- This feature should also fix issue #499
## Goals
- [ ] Decouple the top-k local prediction top-k evaluation from the retrieval contrastive learning task. 
- [ ] Convert retrieval models to a top-k recommender model : Matrix Factorization, Two-Tower, and YoutubeDNN
- [ ] Ensure the Keras analogy for the top-k recommender where the user can call .predict, .evaluate, .save, and load the model. 
- [ ] ItemRecommender should support different top-k strategies: Brute-force || [Streaming](https://github.com/tensorflow/recommenders/blob/18712407c799cfb7ea53440f3fccd576e69ca580/tensorflow_recommenders/layers/factorized_top_k.py#L334) || [ANN ](https://www.tensorflow.org/recommenders/api_docs/python/tfrs/layers/factorized_top_k/ScaNN) or any user-specific top-k strategy. 
- [ ] Ensure retrieval experiments and CI performance tests are returning same level of performance with the new Retrieval API

## Starting Point:
- **Definition:** The Top-k recommender is a model with : Query encoder + Top-k layer 

- Prerequisite of the top-k recommender: 
	- **Predict method:**  returns top-k items (scores and ids) for  a given query (user)
	- **Evaluate method**: compute ranking metrics for a dataset of users/queries
	- **batch_predict**: return a dataset with top-k items  for a dataset of users/queries
	- **save**: The top-k model is the 'useful' part of the retrieval pipeline as it is the one that generates the prediction for the external endpoint. The user needs to save this model and reload it later for evaluation or local prediction
        - Supporting different top-k strategies

- Arguments of the top-k layer: 
	- A cut-off  `k` 
	- The dataset of candidates: pre-trained item embeddings 
	- The method `index_from_dataset`:  to set the index for the top-k search 
	- The method `score`:  the distance metric to use for computing the score between the query and the item embeddings. (default: dot-product) 
	- Call method: 
		* Takes as input the query embeddings 
		* Define the logic of how to retrieve the top-k items :   		* The scope of this first work is "Brute-Force"

Open questions:
- Do we need to re-train the new recommender model with the pre-trained item embeddings? ==> e.g., convert a two-tower model to a  youtube-dnn like model.  ==> This can be done outside of the top-k recommender class. We should simplify the top-k recommender for supporting different top-k strategies

- Should we define the Top-k layer as a sub-class of the CategoricalPrediction block? 

- implementation starting points: 
--> https://github.com/NVIDIA-Merlin/models/compare/main...tf/retrieval-models 
--> https://github.com/NVIDIA-Merlin/models/pull/663/files

sararb:
Fixes #622

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Dict, List, Optional, Protocol, Sequence, Unio
 
 import six
 import tensorflow as tf
+from keras.engine import data_adapter
 from keras.utils.losses_utils import cast_losses_to_common_dtype
 from packaging import version
 from tensorflow.keras.utils import unpack_x_y_sample_weight
@@ -26,7 +27,9 @@ from merlin.models.tf.metrics.topk import TopKMetricsAggregator, filter_topk_met
 from merlin.models.tf.models.utils import parse_prediction_tasks
 from merlin.models.tf.prediction_tasks.base import ParallelPredictionBlock, PredictionTask
 from merlin.models.tf.predictions.base import ContrastivePredictionBlock, PredictionBlock
+from merlin.models.tf.predictions.topk import TopKLayer, TopKPrediction
 from merlin.models.tf.typing import TabularData
+from merlin.models.tf.utils.batch_utils import TFModelEncode
 from merlin.models.tf.utils.search_utils import find_all_instances_in_layers
 from merlin.models.tf.utils.tf_utils import (
     call_layer,
@@ -318,6 +321,9 @@ class BaseModel(tf.keras.Model):
                     )
                 self.prediction_blocks[0].compile(negative_sampling=negative_sampling)
 
+            k = kwargs.pop("k", None)
+            if k:
+                self.prediction_blocks[0].compile(k=k)
         # This flag will make Keras change the metric-names which is not needed in v2
         from_serialized = kwargs.pop("from_serialized", num_v2_blocks > 0)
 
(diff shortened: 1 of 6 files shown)


tests/unit/tf/models/test_retrieval.py
```diff
@@ -368,3 +368,37 @@ def last_interaction_as_target(inputs, targets):
...
+
+    dataloader = BatchedDataset(ecommerce_data, batch_size=50)
+    dataloader = dataloader.map(_item_id_as_target)
+
+    model = mm.TwoTowerModelV2(ecommerce_data.schema, query_tower=mm.MLPBlock([64, 128]))
+    model.compile(run_eagerly=run_eagerly, optimizer="adam")
+    _ = model.fit(ecommerce_data, batch_size=50, epochs=5, steps_per_epoch=1)
+
+    recommender = model.to_item_recommender(dataset=ecommerce_data, batch_size=50, k=10)
+    recommender.compile(run_eagerly=run_eagerly, optimizer="adam")
```
gabrielspmoreira, line 387: Do we need to set the optimizer of item_recommender if it is just to be used for evaluation and prediction?

merlin/models/tf/models/base.py
```diff
@@ -1406,3 +1412,102 @@ def _maybe_convert_merlin_dataset(data, batch_size, shuffle=True, **kwargs):
...
+        `Model` on data.
+        """
+        x, _, _ = data_adapter.unpack_x_y_sample_weight(data)
+        outputs, context = self(x, training=False, output_context=output_context)
+        return outputs, context.top_ids
+
+    def batch_predict(
+        self, dataset: merlin.io.Dataset, batch_size: int, top_k=1, **kwargs
+    ) -> merlin.io.Dataset:
+        pass
```
gabrielspmoreira, line 1501: Should raise NotImplementedError()

rnyak:
we can close it since there will be a new PR for top_k_recommender via https://github.com/NVIDIA-Merlin/models/issues/718

PR #638, opened 2022-08-10
Improvements in Wide and Deep architecture and example test based on the paper

timmy00:
Fixes # (issue)
https://github.com/NVIDIA-Merlin/models/pull/623#pullrequestreview-1065967254

### Goals :soccer:
- Add an example (test) of Wide and Deep architecture which is equivalent to the one proposed in [paper](https://arxiv.org/abs/1606.07792), including multiple optimizers ([FTRL](https://github.com/tensorflow/estimator/blob/eb9d58bf952fbeb9fd4e686b7528d6adf0021320/tensorflow_estimator/python/estimator/canned/optimizers.py#L104) for wide part and AdaGrad for deep part) and 2nd-level feature interactions in the wide part.
- Changed docstrings of WideAndDeepModel() to provide more examples of configurations
- Change default value of output_mode for `HashedCross` and `HashedCrossAll` from 'int' to 'one_hot', and makes them accept now 2D features (for feature interaction with `multi_hot` features)
- Provide an option for `HashedCrossAll` to ignore meaningfuless feature combinations (that always lead to the same values), for example: `[["item_id", "item_category"], ["user_id", "user_birth_city"]]`
- Replaces the `CategoricalOneHot` by `CategoryEncoding` transformation, as it supports multi-hot and sparse input/output
- Added support of multi-hot categ features into HashedCross, fixed bugs on graph mode and updated tests

Making either wide or deep part optional for wide&deep

### Implementation Details :construction:
- Use multi optimzier and model.find_blocks_by_name()
- It is important to enable the output of one-hot representations of high-cardinality features to be sparse tensors `e.g. CategoryEncoding(..., sparse=True), HashedCrossAll(..., sparse=True), HashedCross(..., sparse=True)`, otherwise memory consumption is two high and leads to OOM errors.

### Testing Details :mag:
Model test with two mode (run_eagerly = True / False)

PR #666, opened 2022-08-19
Change `output_mode` of HashedCross from `int` to `one_hot`

timmy00:
### Goals :soccer:
Change the default `output_mode` of HashedCross from `int` to `one_hot`


timmy00:
Same as #638 .

Issue #650, opened 2022-08-15
[Task] Change UniformNegativeSampling class name


rnyak:
### Description

The `UniformNegativeSampling` class name might be misleading since it is not really doing uniform negative sampling. It'd be useful to find another name that'd be more accurate and also generic.

oliverholworthy:
Rename completed in #710 to `InBatchNegatives`

PR #668, opened 2022-08-21
Rename UniformNegativeSampling class

rnyak:
Addresses https://github.com/NVIDIA-Merlin/models/issues/650


oliverholworthy:
@rnyak After discussion with @marcromeyn . Seems like adding something about 'in batch' to the name may help make the name closer to it's natural sense.  Something like `InBatchNegatives` or `InBatchNegativeSampling`.

rnyak:
> @rnyak After discussion with @marcromeyn . Seems like adding something about 'in batch' to the name may help make the name closer to it's natural sense. Something like `InBatchNegatives` or `InBatchNegativeSampling`.

Wonder if this would lead confusion for users. Since we also have in batch negative sampling in retrieval,  we dont want people to think that they can use `InBatchNegativeSampling`  class for retrieval models as well..  bcs they wont be able to, if I am not mistaken. so did we decide if InBatchNegativeSampling is only for ranking? If yes, we would explicit explanations in the docstrings to avoid confusion.

sararb:
> > @rnyak After discussion with @marcromeyn . Seems like adding something about 'in batch' to the name may help make the name closer to it's natural sense. Something like `InBatchNegatives` or `InBatchNegativeSampling`.
> 
> @marcromeyn and @oliverholworthy Wonder if this would lead confusion for users. Since we also have in batch negative sampling in retrieval, we dont want people to think that they can use `InBatchNegativeSampling` class for retrieval models as well.. bcs they wont be able to, if I am not mistaken. so did we decide if InBatchNegativeSampling is only for ranking? If yes, we would explicit explanations in the docstrings to avoid confusion.

How about renaming the `UniformNegativeSampling` class to `SampleFromBatch`?

oliverholworthy:
Closing this since the rename has been made in #710 to `InBatchNegatives`

Issue #677, opened 2022-08-25
[Task] Implement save method on Tensorflow implementation of models.


oliverholworthy:
Part of #669 

### Description

Implement the save method on tensorflow models to write out both the tensorflow/keras artifacts as well as the merlin schema in a separate file.

viswa-nvidia:
Pending - Writing output schema, saving the encoder

PR #680, opened 2022-08-25
Add a protocol for expected interface of a model and save input schema to metadata directory

### LINKED ISSUE #676: [Task] Create a protocol for common methods across model frameworks
Part of #669 
Relates to https://github.com/NVIDIA-Merlin/Merlin/issues/545

### Background

Based on the discussions around having a common API across the backend framework implementations. We would like to make this concrete with a protocol that each model can be checked against to help make it clear which methods are part of a common API and which are specific to the various model implementations.

### Description

Write down the common API methods supported by each model in a protocol that can be runtime checked. This can be extended in future. For the purposes of this task. The following may be enough to consider this task complete:

- training methods:
  - `fit` Fit (train) the model on a dataset provided
  - `evaluate` Evaluate the model on a dataset provided (returns metrics)
  - `predict` Run inference on the model returning the output predictions of the model
- io methods:
  - `save` Write out artifact(s) that can be reloaded by the `load` method
  - `load` classmethod to re-load a model from a previously saved model instance.

oliverholworthy:
Fixes #676 and #677 

### Goals :soccer:

- Improve consistency and clarity of models API.
  - adding protocol to document expected methods on a Model.
- Enable use of Merlin Model artifacts to be used in Serving concepts more easily.
  - Allow additional metadata to be stored about inputs by storing Merlin Schema alongside model artifacts

- Provide a protocol that is runtime checkable and serves as a documentation reference of what is expected of a model across the backend implementations.
- Saving input and output Melrin Schema of model (where available) to a Merlin Metadata directory.
  - There is more to do here, like dealing with models where we don't always have a schema available. This is a first step in that direction.


merlin/models/io.py
```diff
@@ -0,0 +1,47 @@
...
+#
+import os
+import pathlib
+from typing import Optional, Union
+
+from merlin.models.api import MerlinModel
+from merlin.models.utils.schema_utils import schema_to_tensorflow_metadata_json
+from merlin.schema import Schema
+
+_MERLIN_METADATA_DIR_NAME = "merlin_metadata"
```
oliverholworthy, line 24: Calling this metadata directory `merlin_metadata` here. 

Another alternative name could be something like `.merlin`.

marcromeyn (approved):

oliverholworthy:
rerun tests

PR #682, opened 2022-08-25
XGBoost - Use DaskDMatrix for evals data to ensure metrics in logs match result of evaluate

oliverholworthy:
### Goals :soccer:

Ensure metrics in logs match the result of the `.evaluate` method.

Example in this gist of the metrics difference:
https://gist.github.com/oliverholworthy/db8dfca6e08d29fa6481f5f9d66c8049

Issue reported by @rnyak 

When using the DaskDeviceQuantileDMatrix in the `evals` passed to the train function, there is a difference between the metrics reported in the logs printed to the screen and the metric you will get from the call to `model.evaluate`.

We can resolve this by always using the DaskDMatrix for evals data (Thanks to @radekosmulski for this suggestion). The downside of this might be if you have a sufficiently large dataset you want to pass to evals that requires the efficiency of the DeviceQuantileDMatrix when running on the GPU. 

### Testing Details :mag:

Added a test to ensure that dmatrix passed to evals is not the device quantile version.

### DIFF
--- a/merlin/models/xgb/__init__.py
+++ b/merlin/models/xgb/__init__.py
@@ -171,7 +171,10 @@ class XGBoost:
                 self.target_columns,
                 self.qid_column,
             )
-            d_eval = dmatrix_cls(self.dask_client, X, label=y, qid=qid)
+            # using the quantile DMatrix as part of evals results in a
+            # discrepancy between metrics reported in logs and result
+            # of evaluate
+            d_eval = xgb.dask.DaskDMatrix(self.dask_client, X, label=y, qid=qid)
             watchlist.append((d_eval, name))
 
         train_res = xgb.dask.train(
--- a/tests/unit/xgb/test_xgboost.py
+++ b/tests/unit/xgb/test_xgboost.py
@@ -121,14 +121,13 @@ class TestXGBoost:
     ],
 )
 @patch("xgboost.dask.train", side_effect=xgboost.dask.train)
-def test_gpu_hist_dmatrix(
-    mock_train, fit_kwargs, expected_dtrain_cls, dask_client, music_streaming_data: Dataset
-):
-    schema = music_streaming_data.schema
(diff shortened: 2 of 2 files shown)


radekosmulski (approved):

rnyak:
@oliverholworthy regarding your statement `he downside of this might be if you have a sufficiently large dataset you want to pass to evals that requires the efficiency of the DeviceQuantileDMatrix when running on the GPU` 
Can we not make this optional?   this is an important feature and we actually used it for fitting data on a single GPU for a customer use case. So how about adding an arg in the evals in the model.fit() and exposing it to the user so they can choose if they want DaskDMatrix or DeviceQuantileDMatrix? default it can be `DaskDMatrix`. What do you think?

oliverholworthy:
> this is an important feature and we actually used it for fitting data on a single GPU for a customer use case. So how about adding an arg in the evals in the model.fit() and exposing it to the user so they can choose if they want DaskDMatrix or DeviceQuantileDMatrix?

Similar to the param `use_quantile`  - which controls the dmatrix type for both the training data and the eval data currently. An additional parameter to control the dmatrix type of the evals data specifically (e.g. `evals_use_quantile`) would work. However, I wonder if the additional parameter would add more complexity than required to this interface.

If using the device quantile version of the dmatrix is very important when using the GPU (even for eval data) then perhaps it would make sense to keep the code as-is (closing this PR)? And instead add notes/documentation about the metrics - that logged metrics won't match the result of the `.evaluate` method exactly. If we added a param to control the dmatrix type of the evals data, we'd have to handle this anyway. 

The only reason I can think of the extra param would be required if someone was training on GPU with a sufficiently small validation dataset (where they know for sure that the benefits of the device quantile dmatrix are not required) and it was important that the logged metrics matched up with evaluation metrics after the training. If that's a likely requirement then adding this extra param is the way to go.

Issue #626, opened 2022-08-05
[BUG] getting error from DLRM/DCN when using UniformNegativeSampling


rnyak:
### Bug description
I am getting the following error from DLRM/DCN model when I use `UniformNegativeSampling`.

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Input In [14], in <cell line: 10>()
      1 model = mm.DLRMModel(
      2     schema,
      3     embedding_dim=64,
   (...)
      6     prediction_tasks=mm.BinaryClassificationTask('rating_binary'),
      7 )
      9 model.compile("adam", run_eagerly=False, metrics=[tf.keras.metrics.AUC()])
---> 10 model.fit(dataset, batch_size=batch_size)

File /usr/local/lib/python3.8/dist-packages/merlin/models/tf/models/base.py:713, in BaseModel.fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq, max_queue_size, workers, use_multiprocessing, train_metrics_steps, **kwargs)
    705 callbacks = self._add_metrics_callback(callbacks, train_metrics_steps)
    707 fit_kwargs = {
    708     k: v
    709     for k, v in locals().items()
    710     if k not in ["self", "kwargs", "train_metrics_steps", "__class__"]
    711 }
--> 713 return super().fit(**fit_kwargs)

File /usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py:67, in filter_traceback.<locals>.error_handler(*args, **kwargs)
     65 except Exception as e:  # pylint: disable=broad-except
     66   filtered_tb = _process_traceback_frames(e.__traceback__)
---> 67   raise e.with_traceback(filtered_tb) from None
     68 finally:
     69   del filtered_tb

File /usr/local/lib/python3.8/dist-packages/keras/engine/data_adapter.py:1577, in unpack_x_y_sample_weight(data)
   1574 else:
   1575   error_msg = ("Data is expected to be in format `x`, `(x,)`, `(x, y)`, "
   1576                "or `(x, y, sample_weight)`, found: {}").format(data)
-> 1577   raise ValueError(error_msg)

[...]

Issue #596, opened 2022-07-20
[BUG] The negative sampling for ranking gives AUC 0.00 value when we add `sampling` layer  to the Model class


rnyak:
### Bug description

The negative sampling for ranking gives AUC 0.00 value and binary classification acc 0.999 when we add `sampling` layer to the `mm.Model()` class. 

### Steps/Code to reproduce bug

Please run the code below with any synthetic dataset and corresponding schema to repro the issue:

```
sampling = UniformNegativeSampling(schema, 5, seed=42)

model = mm.Model(
    mm.InputBlock(schema),
    sampling,
    mm.MLPBlock([64]),
    mm.BinaryClassificationTask("label"),
)


BATCH_SIZE=2048
model.compile("adam", run_eagerly=False, metrics=[tf.keras.metrics.AUC()])
model.fit(train, batch_size=BATCH_SIZE)
``` 

### Expected behavior

We should be getting AUC > 0

### Environment details
- Merlin version: 
- Platform:
- Python version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?):

`merlin-tensorflow-training:22.05` image with the latest main branches pulled.

viswa-nvidia:
@rnyak , please triage this bug

oliverholworthy:
So it looks like this is something to do with eager/graph mode. Running with `model.compile(..., run_eagerly=True)` produces a non-zero AUC. Example below:

```python
import pyarrow as pa
import tensorflow as tf
from merlin.models.tf.data_augmentation.negative_sampling import UniformNegativeSampling
import merlin.models.tf as mm
from merlin.io import Dataset
from merlin.datasets.entertainment import get_movielens


train, valid = get_movielens(variant="ml-100k")
schema = train.schema
schema = schema.without(["rating"])

# keep only positives
train_df = train.to_ddf().compute()
train = Dataset(train_df[train_df["rating_binary"] == 1])
train.schema = schema

model = mm.Model(
    mm.InputBlock(schema),
    UniformNegativeSampling(schema, 5, seed=42),
    mm.MLPBlock([64]),
    mm.BinaryClassificationTask("rating_binary"),
)

model.compile("adam", metrics=[tf.keras.metrics.AUC()], run_eagerly=True)
model.fit(train, batch_size=2048, epochs=1)
# => 25/25 [======] - 5s 187ms/step - loss: 0.4996 - auc_1: 0.5002
```

oliverholworthy:
[...]

PR #685, opened 2022-08-26
Update In Batch Negative Sampling Augmentation Layer for primary use with dataloader

oliverholworthy:
Related to #596 and #626

### Goals :soccer:

- Improve speed of in-batch negative sampling when used as part of the dataloader
- Simplify use of in-batch negative sampling when used as part of the dataloader

- Adding the `tf.function` decorator to the `call` method of the in-batch negative sampling data augmentation layer.
- Change default value of `return_tuple` so that this isn't required when used with `BatchedDataset.map`
 

### Testing Details :mag:

slightly faster when wrapped with a tf.function. though not much practical difference unless used with very with large datasets.

```python
import pyarrow as pa
from merlin.models.tf.data_augmentation.negative_sampling import UniformNegativeSampling
from merlin.models.tf.dataset import BatchedDataset
from merlin.datasets.synthetic import generate_data
from toolz import last
from merlin.io import Dataset
import tensorflow as tf

train, valid = generate_data("movielens-100k", num_rows=int(5e5), set_sizes=(0.8, 0.2))

schema = train.schema
schema = schema.without(["rating"])

# keep only positives
train_df = train.to_ddf().compute()
train = Dataset(train_df[train_df["rating_binary"] == 1])
train.schema = schema

sampling = UniformNegativeSampling(schema, 5, seed=42, return_tuple=True)
tf_sampling = tf.function(sampling)

%timeit -n 20 last(BatchedDataset(train, batch_size=100).map(sampling))
# => 71.5 ms ┬ 1.69 ms per loop (mean ┬ std. dev. of 7 runs, 20 loops each)
%timeit -n 20 last(BatchedDataset(train, batch_size=100).map(tf_sampling))
# => 41.3 ms ┬ 1.09 ms per loop (mean ┬ std. dev. of 7 runs, 20 loops each)
```


merlin/models/tf/data_augmentation/negative_sampling.py
```diff
@@ -52,7 +52,7 @@ def __init__(
         n_per_positive: int,
         seed: Optional[int] = None,
         run_when_testing: bool = True,
-        return_tuple: bool = False,
+        return_tuple: bool = True,
```
marcromeyn, line 55: In what cases does it matter whether it’s a standard tuple versus our `Prediction`-object which is a NamedTuple.

oliverholworthy: This matters when this sampling layer is used as an argument to `BatchedDataset.map(...)`. When returning the Prediction NamedTuple (`return_tuple=False`) and passed to a model.fit an exception is raised from keras. Reported in #626 

```
ValueError: Data is expected to be in format `x`, `(x,)`, `(x, y)`, or `(x, y, sample_weight)`, found: Prediction
```

This `return_tuple` flag was added to work around this issue. However, it doesn't feel like the most natural thing to use.

The proposal in this PR is to change the default so that the supported use-case (in the dataloader) works without requiring this parameter to be specified. If we want to support running this as part of the model, we should aim to find a way for it to work without requiring this parameter.

marcromeyn: I would prefer to keep the default of this flag `False` and to put in a small PR to implement our own version of `unpack_x_y_sample_weight` (and make use of that in [train_step](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/models/base.py#L580) & [test_step](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/models/base.py#L598))

oliverholworthy: I've tried out the idea of updating `train_step` and `test_step`  in #702 . However, it seems that inside the keras model there are methods we're calling on the keras model base class that use  `unpack_x_y_sample_weight`. So updating `train_step` and `test_step` is not enough. I guess monkey-patching the function might work around this, though doesn't feel like the preferred solution.

oliverholworthy: An alternative that borrows from this idea, is to use the fact that our `train_step` and `test_step` pass in the parameters `training` and `testing`. We can use those to figure out if the layer is being called from a Merlin Model. And figure out the return type based on those.

Implemented this idea in PR #703

oliverholworthy:
Replaced by #703

PR #664, opened 2022-08-19
[WIP] Add Dressipi RecSys Util Function to Merlin Models

bschifferer:
Similar to ALICCP, we added util function for dressipi recsys dataset + a synthetic dataset schema file

PR #690, opened 2022-08-30
Fix ecommerce session based test

edknv:
### Goals :soccer:
This PR addresses flaky unit test in [tests/unit/tf/examples/test_usecase_ecommerce_session_based.py](https://github.com/NVIDIA-Merlin/models/pull/664/files#diff-dff76fa6ac53b51fdb34a6fab955589be39a2694209374729eccce41e56d6313) added in #664. The new test has been failing non-deterministically, e.g., [in this run](https://github.com/NVIDIA-Merlin/models/runs/8076039594?check_suite_focus=true) or [this run](https://github.com/NVIDIA-Merlin/models/runs/8078235931?check_suite_focus=true).

The exact root cause is not 100% clear, but it seems that there is not enough data in `valid` dataset to split across 10 parquet files. This should probably investigated as a follow up. For now, this PR proposes to create a single partquet file in order to unblock CI.

### Testing Details :mag:
Unblocks CI


examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb
```diff
@@ -687,7 +687,7 @@
     "\n",
     "# transform and save data\n",
     "workflow.transform(train).to_parquet(os.path.join(DATA_FOLDER, \"train/\"), output_files=10)\n",
```
bschifferer, line 689: should we do the same for train, as well?

bschifferer (approved):

PR #700, opened 2022-08-30
Extend `EmbeddingTable` and `Embeddings` with support for shared embeddings

### LINKED ISSUE #697: [TASK] Add capability to EmbeddingTable to hold multiple features that share id-encoding
As of now a `EmbeddingTable` holds a single feature which doesn't allow for shared embedding-tables. To support this we propose to add the capability to a table to add features as follows:

```python
item_id_col = ...
user_item_history = ...

table = EmbeddingTable(64, item_id_col, user_item_history)

table.add_feature(col)

table.schema # -> Schema(item_id_col, user_item_history)

# Ways to call this
table(tf.Tensor([1,2,3])) # Tensor[3, 64]
table({"item_id": tf.Tensor([1,2,3])}) # Dict[str, Tensor[3, 64]]
table({
	"item_id": tf.Tensor([1,2,3]),
	"user_item_history": tf.RaggedTensor([1], [2, 4], [3])
}) # Dict[str, Tensor[3, 64]]
```

oliverholworthy:
Resolves #697 

### Goals :soccer:

Support shared embedding tables with the `EmbeddingTable` layer.

- Update `EmbeddingTable` to allow lookup of embeddings from multiple features passed in.
- Add support for `has_schema` and `schema`  (from the `SchemaMixin`) on the `EmbeddingTable`
- Adds support for creating an `EmbeddingTable` from more than one ColumnSchema
  - When more than one coluum schema is provided they are required to have the same value of int_domain.max and int_domain.name (if set)
```python
embedding_table = EmbeddingTable(dim, col_schema_a, col_schema_b)
# an alternative / additional method is provided for adding more features
embedding_table.add_feature(col_schema)
```
- Adds support for creating shared embedding table in `Embeddings` function where some column schemas share the same domain name

### Testing Details :mag:

- Tests added for shared embedding domain functionality.

### DIFF
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -62,7 +62,7 @@ EMBEDDING_FEATURES_PARAMS_DOCSTRING = """
 
 
 class EmbeddingTableBase(Block):
-    def __init__(self, dim: int, col_schema: ColumnSchema, trainable=True, **kwargs):
+    def __init__(self, dim: int, col_schema: ColumnSchema, *, trainable=True, **kwargs):
         super(EmbeddingTableBase, self).__init__(trainable=trainable, **kwargs)
         self.dim = dim
 
@@ -184,6 +184,7 @@ class EmbeddingTable(EmbeddingTableBase):
         self,
         dim: int,
         col_schema: ColumnSchema,
+        *,
         embeddings_initializer="uniform",
         embeddings_regularizer=None,
         activity_regularizer=None,
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -342,9 +342,10 @@ class EmbeddingTable(EmbeddingTableBase):
         if isinstance(input_shape, dict):
             output_shapes = {}
             for feature_name in self.feature_names:
-                output_shapes[feature_name] = self._compute_output_shape_table(
-                    input_shape[feature_name]
-                )
+                if feature_name in input_shape:
+                    output_shapes[feature_name] = self._compute_output_shape_table(
+                        input_shape[feature_name]
+                    )
         else:
             output_shapes = self._compute_output_shape_table(input_shape)
 
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -483,7 +483,13 @@ def Embeddings(
 
     for col in schema:
         table_kwargs = _forward_kwargs_to_table(col, table_cls, kwargs)
-        tables[col.name] = table_cls(_get_dim(col, dim, infer_dim_fn), col, **table_kwargs)
+        table_name = col.int_domain.name or col.name
+        if table_name in tables:
+            tables[table_name].add_feature(col)
+        else:
+            tables[table_name] = table_cls(
+                _get_dim(col, dim, infer_dim_fn), col, name=table_name, **table_kwargs
+            )
 
     return ParallelBlock(
         tables, pre=pre, post=post, aggregation=aggregation, name=block_name, schema=schema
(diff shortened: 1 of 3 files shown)


merlin/models/tf/inputs/embedding.py
```diff
@@ -85,6 +87,32 @@ def from_pretrained(
...
+            col_schema.int_domain.name
+            and self.col_schema.int_domain.name
+            and col_schema.int_domain.name != self.col_schema.int_domain.name
+        ):
+            raise ValueError(
+                "`col_schema` int-domain name does not match table domain name. "
+                f"{col_schema.int_domain.name} != {self.col_schema.int_domain.name} "
+            )
+
+        if col_schema.int_domain.max != self.col_schema.int_domain.max:
```
gabrielspmoreira, line 108: Shouldn't we be using `cardinality` instead of `int_domain.max`?

merlin/models/tf/inputs/embedding.py
```diff
@@ -62,7 +62,7 @@
 
 
 class EmbeddingTableBase(Block):
-    def __init__(self, dim: int, col_schema: ColumnSchema, trainable=True, **kwargs):
+    def __init__(self, dim: int, col_schema: ColumnSchema, *, trainable=True, **kwargs):
```
gabrielspmoreira, line 65: I have just a question here.
Does the `*,` allows providing a list to `col_schema`?
Is it equivalent to passing `*col_schema`?
We could have a test demonstrating how multiple col schemas can be passed to `EmbeddingTable` (in addition to the `embedding_table.add_feature()` option)

oliverholworthy: the `*` here forces the subsequent parameters to be passed as keyword-only (https://peps.python.org/pep-3102/). Without this, it would be valid to pass `trainable` as a positional parameter (without the name). 

The * is an 'unnamed' varargs. So you can't use it along side `*col_schema` here. It's one or the other. Both will turn subsequent params into keyword-only params.

_Related note_
There's also a way to make some args positional-only with `/`. That may be useful for some places in our codebase too.However, it was introduced in Python 3.8 (https://peps.python.org/pep-0570/), so would force 3.8 to be the minimum python we support. (We're only running tests on 3.9 currently, so there may already be some things we're using in Python that implicitly have pushed our min-version to at least 3.8)

oliverholworthy: I've updated to support passing multiple column schemas to the __init__. And implemented support for the `schema` property.

merlin/models/tf/inputs/embedding.py
oliverholworthy: the implementation of the embedding table uses int_domain.max so I think changing that should be a separate PR if we wanted to do that.

I've been wondering what benefit we'd get from changing to use the `embedding_sizes.cardinality` property. The cardinality makes sense as a property of the data, but doesn't feel like `embedding_sizes` is an attribute of the data depending on what we intend the schema to describe. If the main reason is to improve the code clarity we could perhaps implement a helper property on the columnschema. 

The support for shared embedding tables is based on the int domain name. Which suggests that the int domain is a closer match for computing the cardinality too.

gabrielspmoreira:
@oliverholworthy I have tested your PR in a schema with multiple features sharing the same domain and it seems to be working fine.
I have tested locally a change of `Embeddings()` function to create a table per domain and add the features with the same domain automatically. Maybe you include that change in this PR.

```python
def Embeddings(...):
   ...
    for col in schema:
        table_kwargs = _forward_kwargs_to_table(col, table_cls, kwargs)
        domain_name = col.int_domain.name or col.name
        if domain_name not in tables:
            tables[domain_name] = table_cls(_get_dim(col, dim, infer_dim_fn), col, **table_kwargs)
        else:
            # Sharing embeddings for features with the same int_domain
            tables[domain_name].add_feature(col)
    ...
```

marcromeyn (approved):

PR #701, opened 2022-08-31
Add select_by_tag to ParallelBlock

### LINKED ISSUE #694: [TASK] Add select_by_tag to ParallelBlock
We would like to add capabilities to a `ParallelBlock` to select sub-graphs. This can be useful for instance to select the `item-id` embedding-table from a `InputBlock`. Another instance where this could be useful is to enable shared-embeddings for two-tower like models. At the moment, we would instantiate a different `InputBlock` per tower which doesn't allow for shared-embeddings. A way to enable this is to select the right feature-branches from a `InputBlock`.

As an example, let's say we have the following schema:
- User features: `user-id`, `last-purchase` (is shared encoded with the item-id)
- Item features: `item-id`

```python
all_inputs = InputBlockV2(schema)
# This would result in a parallel-block with 2 branches:
# user-id -> EmbeddingTable(user_id)
# item-id, last-purchace -> EmbeddingTable(item_id, last_purchase)

user_inputs = all_inputs.select_by_tag(Tags.USER)
# Results in a parallel-block with 2 branches:
# user-id -> EmbeddingTable(user_id)
# last-purchace -> EmbeddingTable(item_id, last_purchase)

item_inputs = all_inputs.select_by_tag(Tags.ITEM)
# Results in a parallel-block with 2 branches:
# item-id -> EmbeddingTable(item_id, last_purchase)

```

As can be seen in the previous example, `select_by_tag` sub-selects the branches in a `ParallelBlock` from a feature-perspective.

edknv:
Fixes #694


tests/unit/tf/core/test_combinators.py
```diff
@@ -103,6 +103,54 @@ def test_parallel_block_select_from_names():
         blocks.select_by_names(["0", "2"])
 
 
+def test_parallel_block_select_by_tags(music_streaming_data):
+    continuous_block = mm.Filter(Tags.CONTINUOUS)
+    embedding_block = mm.EmbeddingFeatures.from_schema(
```
gabrielspmoreira, line 108: I think we could use here `Embeddings()` instead of `EmbeddingFeatures`, as the latter is going to be deprecated in the future.

gabrielspmoreira (commented):
Just to double-check, does the current implementation of this method returns branches that contain at least one feature with the filter tag or only columns that contains the Tag.
I think the latter would be preferred. 
For example, let's say I want to filter only sequential features (Tags.SEQUENCE / Tags.LIST). Typically sequential features might be in both "categorical" (e.g. sequence of product ids) and "continuous" (sequence of product prices) branches.
If a user calls `select_by_tag(Tags.SEQUENCE)`, he would hope that only sequential features are returned, and not the whole "categorical" and "continuous" branches because they contain sequential features.

edknv:
> Just to double-check, does the current implementation of this method returns branches that contain at least one feature with the filter tag or only columns that contains the Tag. I think the latter would be preferred. For example, let's say I want to filter only sequential features (Tags.SEQUENCE / Tags.LIST). Typically sequential features might be in both "categorical" (e.g. sequence of product ids) and "continuous" (sequence of product prices) branches. If a user calls `select_by_tag(Tags.SEQUENCE)`, he would hope that only sequential features are returned, and not the whole "categorical" and "continuous" branches because they contain sequential features.

@gabrielspmoreira It's the former with this implementation, i.e., we return `categorical` and `continuous` branches with filtered columns if the user had those branches in the original parallel block before the filtering. Wouldn't the user want to keep the whole `categorical` and `continuous` branches if they were both in the original parallel block and they both contain sequential features?

One option would be to add an optional parameter, something like `return_columns_only=True`, and return only the columns if True and keep the original branches if False. I'm not sure what the default behavior should be though. I would have guessed `return_columns_only=False` would be a preferred default.

marcromeyn:
I think it could be useful to add some better docs around this method since it might be slightly confusing for users. This because we select branches by features and not by branch-name. I started with trying to explain it better in this [issue](https://github.com/NVIDIA-Merlin/models/issues/694#issue-1355532644).

oliverholworthy:
> returns branches that contain at least one feature with the filter tag or only columns that contains the Tag.

@gabrielspmoreira  Are you thinking of the case where a branch of the ParallelBlock is itself another ParallelBlock that we can also call `select_by_tag` on? or something else?

If we wanted some other way to select sub-sets of branches, we'd need to come up with a mechanism for them to declare how to be sub-selected (perhaps a check for the attribute `select_by_tag` would work. And if there are other types of blocks that are valid to select a subset of by tag we could implement this method).

One thing about the logic '_filter branches that contain at least one feature with the tag_' - We may run into issues if a branch requires all features (or a superset of those specified by the tag). Depending on the expectations about how this tag affects the input features to these branches. If for example a branch inputs contains features tagged with (`{"user", "item"}`) and requires all of them to function correctly and we call `.select_by_tag("user")`, this will return the branch expecting both user and item features to be passed. 

We might need a way to specify on a branch schema which features are required/optional. And potentially an option to `select_by_tag` to control whether the branch selection criteria is 'at least one input feature with the tag' or 'all input features with the tag'

edknv:
Update:

I changed the approach so that `ParallelBlock` calls the `select_by_tag` method of the branches: `EmbeddingTable` and `TabularBlock`. Now, `ParallelBlock` simply calls `EmbeddingTable.select_by_tag()` and `TabularBlock.select_by_tag()` (and also `Filter.select_by_tag`.

With the previous implementation, the schema change was not being propagated to the branches, so even though the `ParallelBlock`'s schema was filtered correctly, the output features were not being filtered correctly. That is, the following didn't work with the previous implemention, but now it does:
```python
    batch = mm.sample_batch(music_streaming_data, batch_size=10, include_targets=False)
    continuous_outputs = continuous_inputs(batch)
    assert sorted(continuous_outputs.keys()) == sorted(continuous_inputs.schema.column_names)
```

Also, more on Gabriel's earlier point:
> Just to double-check, does the current implementation of this method returns branches that contain at least one feature with the filter tag or only columns that contains the Tag. I think the latter would be preferred.
I double-checked and compared the output formats of `ParallelBlock` with and without `select_by_tag`, and realized that I was doing it wrong. `ParallelBlock` should not have the branch names in its schema (the branch names are in the attribute `parallel_layers`) and it had only the columns in the schema. Thus, now it's changed so that even after `select_by_tag` the schema has only column that contains the tags.

edknv:
Note: I'm not sure why `tensorflow/examples` test is failing and can't reproduce it locally. On my local machine, all tests under `examples` passed.

merlin/models/tf/inputs/embedding.py
```diff
@@ -244,6 +245,39 @@ def __init__(
...
+
+        Returns
+        -------
+        An EmbeddingTable if the tags match. If no features match, it returns None.
+        """
+        if not isinstance(tags, collections.Sequence):
+            tags = [tags]
+
+        config = self.get_config()
+        schema = config.pop("schema", None)
```
oliverholworthy, line 269: could these 6 lines could be replaced by `selected_schema = self.schema.select_by_tag(tags)`, since the schema is always going to be set on the embedding table?

merlin/models/tf/inputs/embedding.py
```diff
@@ -244,6 +245,39 @@ def __init__(
...
+        if not schema:
+            return
+        if not isinstance(schema, Schema):
+            schema = schema_utils.tensorflow_metadata_json_to_schema(schema)
+        selected_schema = schema.select_by_tag(tags)
+        if not selected_schema:
+            return
+        config["schema"] = schema_utils.schema_to_tensorflow_metadata_json(selected_schema)
+        embedding_table = EmbeddingTable.from_config(config, table=self.table)
```
oliverholworthy, line 278: Passing the table here is so that we are creating a shallow copy with the same layer objects being referenced. instead of cloning the layer? Or a shared embedding table might then become independently trained which wouldn't be what we want when using this?

merlin/models/tf/core/tabular.py
```diff
@@ -597,6 +606,17 @@ def get_config(self):
 
         return config
 
+    def select_by_tag(self, tags: Tags) -> Optional["Filter"]:
+        if isinstance(self.feature_names, Tags):
+            schema = self.schema.select_by_tag(self.feature_names).select_by_tag(tags)
+        elif isinstance(self.feature_names, collections.Sequence):
+            schema = self.schema.select_by_name(self.feature_names).select_by_tag(tags)
+        else:
+            schema = self.schema.select_by_tag(tags)
```
oliverholworthy, line 615: in what scenario would this else be reached?

merlin/models/tf/core/tabular.py
```diff
@@ -140,6 +141,10 @@ def __init__(
         if schema:
             self.set_schema(schema)
 
+    def select_by_tag(self, tags: Tags) -> Optional["TabularBlock"]:
```
oliverholworthy, line 144: Is this method required if the `Block` has `select_by_tag` implemented? Or do we need to do something different here?

merlin/models/tf/inputs/embedding.py
edknv: Yes, thank you!

merlin/models/tf/core/tabular.py
edknv: Never in theory :) Replaced it with raising an error.

merlin/models/tf/core/tabular.py
edknv: We need a new method here because need to sub-select the columns we want.

merlin/models/tf/inputs/embedding.py
edknv: Yes, I intentionally did a shallow copy. Otherwise, I think we would be creating two independent tables? For example, in a two-tower, if both item and user towers were created with `select_by_tag` and the user tower used the shared embedding table, the weights would be different with a deep copy.

merlin/models/tf/core/combinators.py
```diff
@@ -453,6 +453,56 @@ def select_by_names(self, names: List[str]) -> Optional[List[Block]]:
...
+            tags = [tags]
+
+        selected_branches = {}
+        selected_schemas = Schema()
+
+        for name, branch in self.parallel_dict.items():
+            branch_has_schema = getattr(branch, "has_schema", False)
+            if not branch_has_schema:
+                continue
+            selected_branch = branch.select_by_tag(tags)
```
marcromeyn, line 496: Wouldn’t it be safer to do `hasattr(branch, “select_by_tag”)`?

merlin/models/tf/core/combinators.py
```diff
@@ -453,6 +453,56 @@ def select_by_names(self, names: List[str]) -> Optional[List[Block]]:
...
+            if not selected_branch:
+                continue
+            selected_branches[name] = selected_branch
+            selected_schemas += selected_branch.schema
+
+        if not selected_branches:
+            return
```
marcromeyn, line 503: I am wondering what the most logical behaviour here is? I am inclinded to say that maybe raising an exception could be better. Thoughts @gabrielspmoreira @oliverholworthy @edknv?

merlin/models/tf/core/combinators.py
edknv: I think a branch being skipped silently because it does not have the method implemented (but has the right tags) is also a concern here, so maybe we should raise an error? (which is what is happening now anyway but the focus would be on a better error message)

merlin/models/tf/core/combinators.py
edknv: I returned `None` here because I was following the behavior of `select_by_name` which returns `None` when there is no match. I'm not sure which is better, but I think we should be consistent in the behavior of `select_by` methods.

edknv: @marcromeyn I tried raising an error here, and it looks like we will need `try... except` blocks and maybe even a custom exception because some blocks are themselves `ParallelBlock`s, e.g., when a branch is an `Embeddings` which itself is a `ParallelBlock`. From the user perspective, it's probably better to raise an exception, but from the implementation perspective, I think just returning nothing here is okay.

marcromeyn (approved):

PR #702, opened 2022-08-31
Check for `Prediction` namedtuple in train/test steps

oliverholworthy:
Based on by comments in #685

### Goals :soccer:

 Check for `Prediction` namedtuple in train/test steps when unpacking data.

Wrapping `unpack_x_y_sample_weight` to handle `Prediction` tuple as input

### Testing Details :mag:

updated test of negative sampling layer in BatchedDataset where the output is a prediction tuple

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -580,7 +580,7 @@ class BaseModel(tf.keras.Model):
         """Custom train step using the `compute_loss` method."""
 
         with tf.GradientTape() as tape:
-            x, y, sample_weight = unpack_x_y_sample_weight(data)
+            x, y, sample_weight = _unpack_x_y_sample_weight(data)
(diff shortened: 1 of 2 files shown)


gabrielspmoreira (commented):
@oliverholworthy  I like this approach, as our DataAugmentation call() methods can always return `Prediction`, regardless if they are used in `dataloader.map()` or as a model transformation (e..g `pre` or `post` of a block)

oliverholworthy:
This approach doesn't work with the keras data adapter implementation. 

A few options include (if we want to remove the awkward param `return_tuple`) and have a layer work both inside the model and in the data loader:
- conditionally change the return type in a data augmentation layer based on the `training/testing` context (#703)
- make the `BatchedDataset` aware of this potential return type and unpack the values after the transform is called.
- For a data augmentation layer that changes inputs/targets - we could instead return a 2-tuple of (transformed_inputs, transformed_targets). And update the model base implementation to handle this case.

PR #703, opened 2022-08-31
Return correct type from data augmentation layer depending on the context

oliverholworthy:
Related to https://github.com/NVIDIA-Merlin/models/issues/596 and https://github.com/NVIDIA-Merlin/models/issues/626

### Goals :soccer:

Enable the negative sampling (data augmentation) layer to be used both in the dataset (Loader.map / transform) and as part of the model without requiring a parameter to control the correct output type.

- removes `return_tuple` parameter from `InBatchNegatives`
- detect whether in training or testing and return a tuple or Prediction (namedtuple) depending on the context.

We can determine if the layer is being called from the Model (either `train_step` or `test_step` because these custom methods on our Model pass in the keyword arguments `training` and `testing` depending on the type of step.

This enables us to return the correct type from the call method depending on whether this is being run inside a model or in the a dataset map step.

### Testing Details :mag:

Updates existing tests to unpack the return value correctly.

### DIFF
--- a/merlin/models/tf/transforms/negative_sampling.py
+++ b/merlin/models/tf/transforms/negative_sampling.py
@@ -64,7 +64,7 @@ class InBatchNegatives(tf.keras.layers.Layer):
         """Extend batch of inputs and targets with negatives."""
 
         def get_tuple(x, y):
-            if training or testing:
+            if training or testing or y is None:
                 return Prediction(x, y)
             return (x, y)
 
--- a/merlin/models/tf/transforms/negative_sampling.py
+++ b/merlin/models/tf/transforms/negative_sampling.py
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-from typing import Optional
+from typing import Optional, Tuple, Union
 
 import tensorflow as tf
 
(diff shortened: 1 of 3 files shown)


gabrielspmoreira:
@oliverholworthy why is this approach of checking `training` or `testing` within `UniformNegativeSampling.call()` preferred than overriding `unpack_x_y_sample_weight` as you propose in #702 ?

merlin/models/tf/data_augmentation/negative_sampling.py
```diff
@@ -62,12 +57,19 @@ def __init__(
...
-        self.return_tuple = return_tuple
 
-    def call(self, inputs: TabularData, targets=None, testing=False, **kwargs) -> Prediction:
+    def call(
+        self, inputs: TabularData, targets=None, training=False, testing=False, **kwargs
+    ) -> Union[Prediction, Tuple]:
         """Extend batch of inputs and targets with negatives."""
+
+        def get_tuple(x, y):
+            if training or testing:
```
marcromeyn, line 67: If you call model.predict the model would be called without either of those flags, could that be an issue?

oliverholworthy: Updated this check to handle `training or testing or y is None` to handle predict here.

marcromeyn (approved):

Issue #698, opened 2022-08-30
[TASK] Add EncoderBlock


marcromeyn:
There’s a need to introduce a new concept: `Encoder` which is a block that encodes features into some representation. This class can be used for prediction but can’t be trained/evaluated in isolation, only as part of a `Model`.

```python
class Encoder(tf.keras.Model):
	# Block that can be used for prediction but not for train/test

	def __init__(
		self, 
		inputs: Union[Schema, Block],
		*blocks: Block,
		pre: Optional[tf.keras.layers.Layer] = None,
		post: Optional[tf.keras.layers.Layer] = None,
		**kwargs,
	):
		...

	@classmethod
	def for_tag(
		cls,
		schema,
		tag,
		*block,
	):
		...

	def encode(
		self, 
		dataset: merlin.io.Dataset,
		id_col=None,
		**kwargs
	) -> merlin.io.Dataset:
		...

	def batch_predict(
		self, 
		dataset: merlin.io.Dataset,
		output_schema: Optional[Schema] = None
		**kwargs
	) -> merlin.io.Dataset:
		...
	
	def predict(
		self,
		x,
		batch_size=None,
		verbose='auto',
		steps=None,
		callbacks=None,
		max_queue_size=10,
		workers=1,
		use_multiprocessing=False
	):
		...

	def save(...):
		# Save with input- & output-schema

	@property
	def schema(self) -> Schema:
		...
	
	@property
	def output_schema(self) -> Schema:
		...

	@property
	def inputs(self) -> Block:
		...
```

PR #705, opened 2022-09-01
Add EncoderBlock

marcromeyn:
Fixes:
-  #698

### Goals :soccer:
Introducing a new concept: Encoder which is a block that encodes features into some representation. This class can be used for prediction but can’t be trained/evaluated in isolation, only as part of a Model.

`EncoderBlock` extends a Keras-model but throws an exception when a user tries to train it since it can only be trained when it’s part of a Model. It extends the Keras-model in order to make it possible to save.


merlin/models/tf/core/encoder.py
```diff
@@ -0,0 +1,120 @@
...
+        combinators.build_sequentially(self, list(self.to_call), input_shape=input_shape)
+        if not hasattr(self.build, "_is_default"):
+            self._build_input_shape = input_shape
+
+    def compute_output_shape(self, input_shape):
+        return combinators.compute_output_shape_sequentially(list(self.to_call), input_shape)
+
+    def __call__(self, inputs, **kwargs):
+        if "features" in kwargs:
+            kwargs.pop("features")
```
sararb, line 51: why do we need to remove the `features`?

sararb (approved):
Looks good to me! I have just left two questions to better understand the EncoderBlock class :)

merlin/models/tf/core/encoder.py
```diff
@@ -0,0 +1,120 @@
...
+
+from merlin.models.tf.core import combinators
+from merlin.models.tf.inputs.base import InputBlockV2
+from merlin.models.tf.utils import tf_utils
+from merlin.schema import Schema
+
+
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class EncoderBlock(tf.keras.Model):
+    """Block that can be used for prediction but not for train/test"""
```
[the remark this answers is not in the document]
marcromeyn: Good catch, I should update these docstrings. We will be able to handle evaluation.

merlin/models/tf/core/encoder.py
marcromeyn: I will add some docs here. It’s because when we serve this model we won’t get the features param. If we leave it in the model that gets exported expects it.

PR #707, opened 2022-09-01
Separate the `dev` and `docs` requirements

karlhigley:
### Goals :soccer:
Running into trouble installing the custom sphinx packages from `git` in CPU CI, so this would help us get the `dev` dependencies without trying to install `docs` dependencies we don't need in that environment.

New `requirements/docs.txt` file

### Testing Details :mag:
The issue we're having crops up on [this Core PR](https://github.com/NVIDIA-Merlin/core/pull/129)


karlhigley:
Closed in favor of #708

PR #709, opened 2022-09-02
Rename BatchedDataset to Loader

marcromeyn:
### Goals :soccer:
To prepare for migrating to our new dataloaders package, this PR renames `BatchedDataset` to `Loader` since that will be the new name. 

This PR also exposes the Loader in our public AP in order to make data-augmentations easier.


marcromeyn:
This seems like a very trivial PR or am I missing something @benfred @jperez999? This PR is blocking a bunch of other PRs we’d like to get in for the release, so a review would be very much appreciated!

PR #710, opened 2022-09-02
Move transformation-layers to sub package

marcromeyn:
### Goals :soccer:
This PR groups all transforms into a sub-package. The goal after this is to make these transforms schema-aware and allow users to user these either in the data-loader or inside the model.


edknv:
rerun tests

merlin/models/tf/prediction_tasks/next_item.py
```diff
@@ -187,13 +220,9 @@ def NextItemPredictionTask(
         )
 
     else:
-        if weight_tying:
-            prediction_call = ItemsPredictionWeightTying(schema)
-
-        else:
-            prediction_call = ItemsPrediction(schema)
+        prediction_call = ItemsPrediction(schema)
```
edknv, line 223: Do we want to remove `if weight_trying is True: prediction_call = ItemsPredictionWeightTying(schema)`?

merlin/models/tf/__init__.py
```diff
@@ -130,6 +118,15 @@
 from merlin.models.tf.predictions.sampling.base import Items, ItemSamplerV2
 from merlin.models.tf.predictions.sampling.in_batch import InBatchSamplerV2
 from merlin.models.tf.predictions.sampling.popularity import PopularityBasedSamplerV2
+from merlin.models.tf.transforms.features import (
```
gabrielspmoreira, line 121: I would prefer the package name to be `transformation` instead of `transforms`

gabrielspmoreira (approved):

PR #715, opened 2022-09-05
Introducing Outputs-V2

marcromeyn:
### Goals :soccer:
This PR contains a few different improvements w.r.t. the model-outputs. This is currently called `PredictionBlock` but we found this naming confusing since it’s not just run in prediction-mode. `ModelOutput` is therefore a better name, this object contains the following:
- Layer to be called in the forward pass (for instance `Dense(1, activation=“sigmoid”)` for binary-classification)
- Default loss (used when no loss is provided in `model.compile`)
- Default metrics (used when no loss is provided in `model.compile`)

The model-output classes this PR provides are as follows:
- `RegressionOutput`
- `BinaryOutput` 
- `CategoricalOutput`. This is doing a full softmax over all the categories.
- `ContrastiveOutput`. This is framing a categorical prediction in a contrastive way for training/evaluation. This class can be used for models like MF/Two-Tower as well as models using a sampled-softmax output.


tests/unit/tf/outputs/test_contrastive.py
```diff
@@ -0,0 +1,238 @@
...
+    assert output.outputs.shape == (batch[1].shape[0], 21)
+
+    model.model_outputs[0].compile(
+        negative_sampling=["in-batch", PopularityBasedSamplerV2(max_id=51996, max_num_samples=20)],
+    )
+    output = model(batch[0], batch[1], training=True)
+    assert output.outputs.shape == (batch[1].shape[0], 71)
+
+
+# def test_contrastive_dot_product(ecommerce_data: Dataset):
```
oliverholworthy, line 128: are these tests intentionally commented out? If so perhaps a comment explaining why may be helpful if this is  intended as a reminder of something?

merlin/models/tf/models/base.py
```diff
@@ -311,15 +311,7 @@ def compile(
...
         else:
-            self.output_names = [block.full_name for block in self.prediction_blocks]
-            negative_sampling = kwargs.pop("negative_sampling", None)
-            if negative_sampling:
-                if not isinstance(self.prediction_blocks[0], ContrastivePredictionBlock):
-                    raise ValueError(
-                        "Negative sampling strategy can be used only with a"
-                        " `ContrastivePredictionBlock` prediction block"
-                    )
-                self.prediction_blocks[0].compile(negative_sampling=negative_sampling)
```
oliverholworthy, line 322: was this explicit compile unnecessary before?

merlin/models/tf/outputs/contrastive.py
```diff
@@ -0,0 +1,345 @@
...
+        name: Optional[str] = None,
+        default_loss: Union[str, tf.keras.losses.Loss] = "categorical_crossentropy",
+        default_metrics_fn: MetricsFn = default_categorical_prediction_metrics,
+        downscore_false_negatives=True,
+        false_negative_score: float = MIN_FLOAT,
+        query_name: str = "query",
+        candidate_name: str = "candidate",
+        **kwargs,
+    ):
+        _to_call = kwargs.pop("to_call", None)
```
edknv, line 109: I realize this is just existing code with the variables renamed, but I'm wondering if it is possible to have `to_call` as a keyword argument when the first argument is `to_call`?
```python
>>> def foo(to_call, **kwargs):
...     _to_call = kwargs.pop("to_call", None)
...     print(_to_call)
... 
>>> foo(None, to_call="bar")
  File "<stdin>", line 1, in <module>
TypeError: foo() got multiple values for argument 'to_call'
```

merlin/models/tf/models/base.py
marcromeyn: It was something we had before as well, but it wasn’t used anywhere in the tests.

merlin/models/tf/outputs/contrastive.py
marcromeyn: Good point

merlin/models/tf/outputs/base.py
```diff
@@ -27,19 +27,18 @@
...
 @tf.keras.utils.register_keras_serializable(package="merlin.models")
-class PredictionBlock(Layer):
+class ModelOutput(Layer):
     """Base-class for prediction blocks.
 
     Parameters
     ----------
-    prediction : Layer
-        The prediction layer
+    call : Layer
```
oliverholworthy, line 40: The argument is named `to_call` in the init method below.

tests/unit/tf/outputs/test_contrastive.py
```diff
@@ -0,0 +1,219 @@
...
+    assert output[1].shape == (batch[1].shape[0], 51)
+
+    model_out.compile(
+        negative_sampling=[PopularityBasedSamplerV2(max_id=51996, max_num_samples=20)],
+    )
+
+    output = model(batch[0], batch[1], training=True)
+    assert output.outputs.shape == (batch[1].shape[0], 21)
+
+    model.model_outputs[0].compile(
```
oliverholworthy, line 119: perhaps we can use `model_out` again for this compile step. If the purpose is to demonstrate that `model_out` here is the same as `model.model_outputs[0]`, that could be in another test perhaps.

tests/unit/tf/outputs/test_contrastive.py
```diff
@@ -0,0 +1,219 @@
...
+    dataloader, schema = _next_item_loader(sequence_testing_data, to_one_hot=False)
+    model_out = mm.ContrastiveOutput(schema["item_id_seq"], "in-batch")
+    model = mm.Model(mm.InputBlockV2(schema), mm.MLPBlock([32]), model_out)
+    model.compile(optimizer="adam")
+
+    batch = next(iter(dataloader))
+    output = model(batch[0], batch[1], training=True)
+    assert output[1].shape == (batch[1].shape[0], 51)
+
+    model_out.compile(
```
oliverholworthy, line 112: Is this `compile` a similar compilation sense to the `model.compile`? It looks like this was already here on the block that has been rename. However, based on what it looks like it's doing, perhaps a different name could be picked if it's a way to change the samplers being used in this layer. (`update_negative_samplers` ?)

oliverholworthy (approved):

marcromeyn:
rerun tests

jperez999:
rerun tests

marcromeyn:
rerun tests

PR #716, opened 2022-09-05
Example notebook of Wide&Deep model

### LINKED ISSUE #704: [Task] Add example of Wide&Deep ranking model
### Description
Add an example of Wide&Deep ranking model in the [03-Exploring-different-models.ipynb](https://github.com/NVIDIA-Merlin/models/blob/main/examples/03-Exploring-different-models.ipynb) notebook.

### LINKED ISSUE #725: [BUG]  WideAndDeepModel func should check if wide_preprocess is not None when wide_input_block is not provided
### Bug description

Currently, if `wide_input_block` is not provided, it is set to None by default and in that case it can be created via `  TabularBlock.from_schema(schema=wide_schema, pre=wide_preprocess),` in these lines   [WideAndDeepModel](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/tf/models/ranking.py#L409-L415) func. However, if `wide_preprocess` is also not provided, it is set to None, in that case no transformation block will be applied and this will result in weighted summing encoded categorical features without applying any transformation (e.g. HashedCross, etc.)

To fix this we should another check in the if statement below. something like:

```
    if not wide_input_block:
        if wide_schema is not None and len(wide_schema) > 0 and wide_preprocess is not None:
            wide_input_block = ParallelBlock(
                TabularBlock.from_schema(schema=wide_schema, pre=wide_preprocess),
                is_input=True,
                aggregation="concat",
            )
```

### Environment details
- Merlin version: 
- Platform:
- Python version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?):

gabrielspmoreira:
Fixes #704 
Fixes #725
This PR adds an example of the [Wide&Deep architecture](https://arxiv.org/abs/1606.07792) we have finished implementing for 22.09. 

### Implementation details
Instead of creating a new notebook, I have added it to the notebook [03-Exploring-different-models.ipynb](https://github.com/NVIDIA-Merlin/models/blob/main/examples/03-Exploring-different-models.ipynb) which also showcases other ranking models like DLRM and DCN.


gabrielspmoreira:
rerun tests

bschifferer:
I think the unit and integration tests needs to be adjusted.

https://github.com/NVIDIA-Merlin/models/blob/main/tests/unit/tf/examples/test_03_exploring_different_models.py
https://github.com/NVIDIA-Merlin/models/blob/main/tests/integration/tf/test_ci_03_exploring_different_models.py

bschifferer:
The example is great, but I feel it is much more complicated than our other Merlin Models examples. For DLRM, DCN, etc, we need to call one function. Is that not possible for Wide & Deep?

Can we not have the following part in mm.WideAndDeepModel 
```
cat_schema = schema.select_by_tag(Tags.CATEGORICAL)
one_hot_encoding = mm.SequentialBlock(
                        mm.Filter(cat_schema),
                        mm.CategoryEncoding(cat_schema, sparse=True, output_mode="one_hot"),
                    )
features_crossing = mm.SequentialBlock(
            mm.Filter(cat_schema),
            #ml.AsDenseFeatures(max_seq_length=5), #Uncomment this  line if you include multi-hot features as input for your feature interaction.
            mm.HashedCrossAll(
                cat_schema,
                num_bins=100, # The resulting features will be hashed to this number of bings
                max_level=2, # Performs 2nd feature interactions, typically maximum is 3rd level, as the higher the level the greater the combinatorial explosion
                output_mode="multi_hot",
                sparse=True,
                ignore_combinations=[["item_id", "item_category"], 
                                     ["item_id", "item_brand"]]
            ),
        )
 wide_preprocessing_blocks = mm.ParallelBlock(
    [
     one_hot_encoding,
     features_crossing
    ],
    aggregation="concat"
)                           
```

?

Why do we have a separate function `get_wide_and_deep_model():` -> can we just call the model itself?

Do the other model support `multi-hot categorical features`? In my understanding, DLRM, DCN doesnt use them, do they?
If not, then I would remove the references for it and we need another advanced example how to use multi-hot features?

gabrielspmoreira:
> The example is great, but I feel it is much more complicated than our other Merlin Models examples. For DLRM, DCN, etc, we need to call one function. Is that not possible for Wide & Deep?
@bschifferer, thanks for your review.

I think you are right, this example of Wide&Deep was very detailed. So I simplified the code quite a bit by supporting only one-hot features (not multi-hot) and also removing the multiple optimizers subsection. I moved out those additional examples to the docstrings (API doc) of `WideAndDeepModel()`, if that makes sense.
Let me know if the example needs further simplification.

gabrielspmoreira:
rerun tests

bschifferer:
Do we need to update the CI test: https://github.com/NVIDIA-Merlin/models/blob/main/tests/integration/tf/test_ci_03_exploring_different_models.py ?

bschifferer (approved):
It is great!

If we do not need to update the CI test, then we can merge it.

gabrielspmoreira:
Sure @bschifferer . That integration had indeed to be updated and I fixed it.

PR #717, opened 2022-09-06
Added InputBlockV2 support to DeepFMModel (refactored and fixed) and DCNModel

### LINKED ISSUE #604: [Task] InputBlockV2 - Usage in ranking models high-level API

### LINKED ISSUE #742: [BUG] DeepFM implementation is incorrect and misses tests
### Bug description
The current implementation of `DeepFMModel` does not match exactly the architecture proposed in the original paper. The FM and deep tower output should be 1d and summed. But instead those towers were outputing multiple dimensions (>1), which were projected to a single dim by an MLP layer, which is not part of the original architectures described in the papers.
```
[1] Huifeng, Guo, et al. "DeepFM: A Factorization-Machine based Neural Network for CTR Prediction". arXiv:1703.04247  (2017).
[2] Steffen, Rendle, "Factorization Machines" IEEE International Conference on Data Mining, 2010.
```
In addition:
    - In the wide part the one-hot categorical features is not sparse, which might cause OOM for high cardinality features.
    - The DeepFMModel supports only categorical features. If continuous features were passed the branch with FMPairwiseInteraction an error were raised

gabrielspmoreira:
Fixes #604 

### Goals :soccer:
- Sets `InputBlockV2` as the default input block for `DCNModel` and `DeepFMModel`
- Creates `ToSparseFeatures` and `ToDenseFeatures` to convert tensor types for selected or all features
- Refactors `DeepFMModel` to extract the `FMBlock` and fixes the implemenation to match the one described in the original papers: (Fixes #742 )
```
[1] Huifeng, Guo, et al. "DeepFM: A Factorization-Machine based Neural Network for CTR Prediction". arXiv:1703.04247  (2017).
[2] Steffen, Rendle, "Factorization Machines" IEEE International Conference on Data Mining, 2010.
```

- Previously the `InputBlock` was used for `DCNModel` and `DeepFMModel`
- For `DCNModel`, the user were allowed to set `embedding_options`. Now the user can provide its own `InputBlock`, or provide custom args for the `Embeddings()` via `InputBlockV2(..., **kwargs)` 
- The previous implementation of `DeepFMModel` had the following issues that are fixed here:
    - In the wide part the one-hot categorical features was not sparse, which might cause OOM for high cardinality features.
    - The DeepFMModel supported only categorical features. If continuous features were passed the branch with FMPairwiseInteraction an error were raised
    - The FM and deep tower output should be 1d and summed. But instead those towers were outputing multiple dimensions (>1), which were projected to a single dim by an MLP layer, which is not part of the original architectures described in the papers.

### Testing Details :mag:
- There was no test for `DeepFMModel`, so added some and fixed bugs

### DIFF
--- a/merlin/models/tf/blocks/interaction.py
+++ b/merlin/models/tf/blocks/interaction.py
@@ -304,7 +304,7 @@ def FMBlock(
 
     fm_input_block = fm_input_block or InputBlockV2(
         cat_schema,
-        embeddings=Embeddings(schema.select_by_tag(Tags.CATEGORICAL), dim=factors_dim),
+        embeddings=Embeddings(schema, dim=factors_dim),
         aggregation=None,
     )
     pairwise_interaction = SequentialBlock(
--- a/merlin/models/tf/blocks/interaction.py
+++ b/merlin/models/tf/blocks/interaction.py
@@ -304,7 +304,7 @@ def FMBlock(
 
     fm_input_block = fm_input_block or InputBlockV2(
         cat_schema,
-        embeddings=Embeddings(schema, dim=factors_dim),
+        categorical=Embeddings(schema, dim=factors_dim),
         aggregation=None,
     )
     pairwise_interaction = SequentialBlock(
--- a/merlin/models/tf/blocks/interaction.py
+++ b/merlin/models/tf/blocks/interaction.py
@@ -294,7 +294,7 @@ def FMBlock(
     wide_input_block = wide_input_block or ParallelBlock(
         {
             "categorical": CategoryEncoding(cat_schema, output_mode="multi_hot", sparse=True),
-            "continuous": Filter(cont_schema, post=ToSparse()),
+            "continuous": Filter(cont_schema).connect(ToSparse()),
         },
         aggregation="concat",
     )
@@ -304,7 +304,7 @@ def FMBlock(
 
     fm_input_block = fm_input_block or InputBlockV2(
         cat_schema,
-        categorical=Embeddings(schema, dim=factors_dim),
+        categorical=Embeddings(cat_schema, dim=factors_dim),
         aggregation=None,
     )
     pairwise_interaction = SequentialBlock(
--- a/merlin/models/tf/blocks/interaction.py
+++ b/merlin/models/tf/blocks/interaction.py
@@ -243,6 +243,7 @@ class FMPairwiseInteraction(Block):
 def FMBlock(
     schema: Schema,
     fm_input_block: Optional[Block] = None,
+    wide_input_block: Optional[Block] = None,
     factors_dim: Optional[int] = None,
     **kwargs,
 ) -> tf.Tensor:
@@ -267,6 +268,10 @@ def FMBlock(
         If not provided, an InputBlockV2 is instantiated based on schema.
         Note: All features (including continuous) are considered in the
         1st-order (wide) part which uses another input block.
+    wide_input_block: Optional[Block], by default None
+        The input for the wide block. If not provided,
+        creates a default block that encodes categorical features
+        with one-hot / multi-hot representation and also includes the continuous features.
     factors_dim : Optional[int], optional
         If fm_input_block is not provided, the factors_dim is used to define the
         embeddings dim to instantiate InputBlockV2, by default None
@@ -280,15 +285,15 @@ def FMBlock(
     cat_schema = schema.select_by_tag(Tags.CATEGORICAL)
     cont_schema = schema.select_by_tag(Tags.CONTINUOUS)
 
-    first_order_inputs = ParallelBlock(
+    wide_input_block = wide_input_block or ParallelBlock(
         {
-            "categorical": CategoryEncoding(cat_schema, output_mode="one_hot", sparse=True),
+            "categorical": CategoryEncoding(cat_schema, output_mode="multi_hot", sparse=True),
             "continuous": SequentialBlock(Filter(cont_schema), ToSparseFeatures()),
         },
         aggregation="concat",
     )
 
-    first_order = first_order_inputs.connect(
+    first_order = wide_input_block.connect(
         MLPBlock([1], activation="linear", use_bias=True, **kwargs)
     )
 
--- a/merlin/models/tf/blocks/interaction.py
+++ b/merlin/models/tf/blocks/interaction.py
@@ -240,7 +240,6 @@ class FMPairwiseInteraction(Block):
         return (input_shapes[0], input_shapes[2])
 
 
-@tf.keras.utils.register_keras_serializable(package="merlin.models")
 def FMBlock(
     schema: Schema,
     fm_input_block: Optional[Block] = None,
--- a/merlin/models/tf/blocks/interaction.py
+++ b/merlin/models/tf/blocks/interaction.py
@@ -24,6 +24,7 @@ from merlin.models.tf.core.base import Block
 from merlin.models.tf.core.combinators import MapValues, ParallelBlock, SequentialBlock
 from merlin.models.tf.core.tabular import Filter
 from merlin.models.tf.inputs.base import InputBlockV2
+from merlin.models.tf.inputs.embedding import Embeddings
 from merlin.models.tf.transforms.features import CategoryEncoding, ToSparse
 from merlin.schema import Schema, Tags
 
@@ -244,8 +245,9 @@ def FMBlock(
     schema: Schema,
     fm_input_block: Optional[Block] = None,
     wide_input_block: Optional[Block] = None,
+    wide_logit_block: Optional[Block] = None,
     factors_dim: Optional[int] = None,
-    **wide_output_kwargs,
+    **kwargs,
 ) -> tf.Tensor:
     """Implements the Factorization Machine, as introduced in [1].
     It consists in the sum of a wide component that weights each
@@ -272,6 +274,10 @@ def FMBlock(
         The input for the wide block. If not provided,
         creates a default block that encodes categorical features
         with one-hot / multi-hot representation and also includes the continuous features.
+    wide_logit_block: Optional[Block], by default None
+        The output layer of the wide input. The last dimension needs to be 1.
+        You might want to provide your own output logit block if you want to add
+        dropout or kernel regularization to the wide block.
     factors_dim : Optional[int], optional
         If fm_input_block is not provided, the factors_dim is used to define the
         embeddings dim to instantiate InputBlockV2, by default None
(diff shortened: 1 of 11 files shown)


merlin/models/tf/models/ranking.py
```diff
@@ -204,32 +189,36 @@ def DeepFMModel(
...
-        embedding_options=EmbeddingOptions(
-            embedding_dim_default=embedding_dim, **embedding_option_kwargs
-        ),
-        **kwargs,
-    )
+    input_block = input_block or InputBlockV2(schema, dim=embedding_dim, aggregation=None, **kwargs)
+
+    # TODO: Identify why we need to set the schema manually for `InputBlockV2`
+    # to avoid an error of ParallelBlock without schema?
+    input_block.set_schema(schema)
```
gabrielspmoreira, line 210: @marcromeyn I had to include this hack because the `InputBlockV2` returns a `ParallelBlock` without schema in this case.  
I suspect that is the case because the data for this test contains only categorical features and the `Filter()` for continuous features gets empty, so that not all branches within `ParallelBlock` contains a schema.
Is this related to your recent refactory that makes `ParallelBlock` schema-aware?

gabrielspmoreira: This hack is not needed anymore, as @marcromeyn implemented a fix for that on `InputBlockV2` in this [PR](https://github.com/NVIDIA-Merlin/models/pull/705)

merlin/models/tf/inputs/base.py
```diff
@@ -218,6 +218,7 @@ def InputBlockV2(
     pre: Optional[BlockType] = None,
     post: Optional[BlockType] = None,
     aggregation: Optional[TabularAggregationType] = "concat",
+    **kwargs,
```
oliverholworthy, line 221: if we're passing these kwargs to the `Embeddings`, maybe can rename this to `embeddings_kwargs`? 

And why would these `kwargs` on the input block be passed to the embeddings instead of the ParallelBlock for example? If the embeddings needs to be customized from the default that can already be achieved by passing in the `embeddings` param?

merlin/models/tf/models/ranking.py
```diff
@@ -202,44 +182,41 @@ def DeepFMModel(
...
-        schema,
-        embedding_options=EmbeddingOptions(
-            embedding_dim_default=embedding_dim, **embedding_option_kwargs
-        ),
-        **kwargs,
-    )
 
-    pairwise_block = FMPairwiseInteraction().prepare(aggregation=StackFeatures(axis=-1))
-    deep_block = deep_block.prepare(aggregation=ConcatFeatures())
+    input_block = input_block or InputBlockV2(schema, dim=embedding_dim, aggregation=None, **kwargs)
```
oliverholworthy, line 205: Perhaps it would be clearer to change the default `aggregation` in the InputBlock to be `None` and change the places where an aggregation is required to pass it?

merlin/models/tf/core/transformations.py
```diff
@@ -40,6 +40,104 @@
...
+            The schema with the columns that will be transformed, by default None
+    """
+
+    def __init__(self, schema: Optional[Schema] = None, **kwargs):
+        super().__init__(**kwargs)
+        self.column_names = None
+        if schema is not None:
+            self.column_names = schema.column_names
+
+    def call(self, inputs: TabularData, **kwargs) -> TabularData:
```
oliverholworthy, line 59: If this is intended to be a base class, is it safe to remove this call method?

tests/unit/tf/blocks/test_interactions.py
```diff
@@ -26,7 +28,66 @@ def test_fm_pairwise_interaction():
...
+            sequence_combiner="mean",
+        ),
+        aggregation=None,
+    )
+
+    wide_input_block = mm.ParallelBlock(
+        {
+            "categorical_ohe": mm.SequentialBlock(
+                mm.Filter(cat_schema_onehot),
+                mm.CategoryEncoding(cat_schema_onehot, sparse=True, output_mode="one_hot"),
```
marcromeyn, line 70: If `mm.CategoryEncoding` already takes in a schema, can’t we make that class also do the filtering?

merlin/models/tf/core/transformations.py
```diff
@@ -40,6 +40,104 @@
...
+
+        return output_shapes
+
+    def compute_call_output_shape(self, input_shapes):
+        return self.compute_output_shape(input_shapes)
+
+
+@Block.registry.register("to_sparse")
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class ToSparseFeatures(FeaturesTensorTypeConversion):
```
marcromeyn, line 93: This naming confuses me. In the moving transforms PR, I am proposing to rename our current layers to `ListToSparse` & `ListToRagged` ([source](https://github.com/NVIDIA-Merlin/models/blob/de8faf947cc7399b547067903439225244180b4d/merlin/models/tf/transforms/tensor.py#L33)).

I think it would make more sense to call this one `ToSparse` and make it take in an optional schema in the constructor. If a schema is provided, only the features in the schema will be converted.

merlin/models/tf/blocks/interaction.py
```diff
@@ -234,4 +235,78 @@ def call(self, inputs: tf.Tensor, **kwargs) -> tf.Tensor:
...
+        and 2nd-order interaction component of FM
+    """
+
+    cat_schema = schema.select_by_tag(Tags.CATEGORICAL)
+    cont_schema = schema.select_by_tag(Tags.CONTINUOUS)
+
+    wide_input_block = wide_input_block or ParallelBlock(
+        {
+            "categorical": CategoryEncoding(cat_schema, output_mode="multi_hot", sparse=True),
+            "continuous": SequentialBlock(Filter(cont_schema), ToSparseFeatures()),
```
marcromeyn, line 291: Could be replace with `ToSparse(cont_schema)` with my previous comment.

merlin/models/tf/blocks/interaction.py
```diff
@@ -234,4 +235,78 @@ def call(self, inputs: tf.Tensor, **kwargs) -> tf.Tensor:
...
+        },
+        aggregation="concat",
+    )
+
+    first_order = wide_input_block.connect(
+        MLPBlock([1], activation="linear", use_bias=True, **kwargs)
```
marcromeyn, line 297: Seems like `**kwargs` is being forwarded to a few different blocks, if possible we’d like to remove this as much as possible since it leads to more understandable code. If you want to allow for full customization is would be cleaner if a param like `wide_logit` be added to `FMBlock` and make `Dense(1, activation="linear", use_bias=True)` the default. MLPBlock is not needed since we only need one dense-layer.

merlin/models/tf/blocks/interaction.py
```diff
@@ -234,4 +235,78 @@ def call(self, inputs: tf.Tensor, **kwargs) -> tf.Tensor:
...
+    )
+
+    fm_input_block = fm_input_block or InputBlockV2(
+        cat_schema, dim=factors_dim, aggregation=None, **kwargs
```
marcromeyn, line 301: We already expose a `fm_input_block` so I feel like forwaring `**kwargs` could be removed. Also, `aggregation=None` is the default so we could remove that here.

merlin/models/tf/blocks/interaction.py
```diff
@@ -234,4 +235,78 @@ def call(self, inputs: tf.Tensor, **kwargs) -> tf.Tensor:
...
+    )
+    pairwise_interaction = SequentialBlock(
+        Filter(cat_schema),
```
marcromeyn, line 304: The `InputBlock` already does filtering so this could be removed.

merlin/models/tf/blocks/interaction.py
marcromeyn: Another maybe easier to read way could be:

```python
InputBlockV2(
  cont_schema, 
  CategoryEncoding(cat_schema, output_mode="multi_hot", sparse=True),
  post=ToSparseFeatures()
)

```

marcromeyn (changes requested):

merlin/models/tf/inputs/base.py
gabrielspmoreira: Makes sense @oliverholworthy . As the user can always provide the `embeddings` as an argment, the **kwargs can be forwarded to another block (`ParallelBlock`)

merlin/models/tf/models/ranking.py
gabrielspmoreira: The only places where currently `InputBlockV2(...., aggregation=None)` are for `DeepFM` and `FMBlock`. So I it seems that is the exception and not the rule.

merlin/models/tf/core/transformations.py
gabrielspmoreira: Sure, the base class `FeaturesTensorTypeConversion` should not have the `call()` method, as it need to be defined in child classes.

merlin/models/tf/core/transformations.py
gabrielspmoreira: Sure @marcromeyn. Renamed the new transformers to `ToSparse` and `ToDense`

merlin/models/tf/blocks/interaction.py
gabrielspmoreira: The `CategoryEncoding` needs the `sparse` argument for the usage of the right TF op for encoding. But I have converted the continuous branch from `SequentialBlock(Filter(cont_schema), ToSparseFeatures())` to `Filter(cont_schema, post=ToSparse())`

merlin/models/tf/blocks/interaction.py
gabrielspmoreira: Makes sense. I was forwarding those **kwargs to be able to add regularization (dropout, kernel reg) to the wide part. But have created the `wide_logit_block` argument you suggested to make it more explicit.

merlin/models/tf/blocks/interaction.py
gabrielspmoreira: I think we should keep that `Filter`, as the user might want to provide his own `fm_input_block`, whose schema might not be filtered to output only categ features.

tests/unit/tf/blocks/test_interactions.py
gabrielspmoreira: The thing is that `CategoryEncoding` can be used with both `output_mode="one_hot"` and `output_mode="multi_hot"`. 
I am not sure we should trust within `CategoryEncoding` that multi-hot features will be tagged as SEQUENCE / LIST. So I used the test as an example on how the use can ensure the right features are selected for each branch

marcromeyn (approved):

gabrielspmoreira:
@marcromeyn I have rebased this PR and did the necessary changes so that it uses your updated `InputBlockV2`

Issue #722, opened 2022-09-07
[BUG] Session-based example breaks in TF 2.10


marcromeyn:
### Bug description

More info can be found here: https://github.com/NVIDIA-Merlin/models/runs/8223278321?check_suite_focus=true

bschifferer:
I checked and reproduced the error. I think this is related to the data loader. The error is raised in `merlin/models/loader/backend.py`. (see below)

It is interesting, that it works, if I run one batch manually through the network by using `model_bi_lstm(next(iter(train_dl))[0])`.

Using this line works:
```
model_bi_lstm(next(iter(train_dl))[0])
history = model_bi_lstm.fit(
    train_dl,
    validation_data=valid,
    batch_size=BATCH_SIZE,
    epochs=EPOCHS,
    schema=schema_model,
)
```

```
---------------------------------------------------------------------------
StopIteration                             Traceback (most recent call last)
Input In [106], in <cell line: 1>()
----> 1 history = model_bi_lstm.fit(
      2     train_dl,
      3     validation_data=valid,
      4     batch_size=BATCH_SIZE,
      5     epochs=EPOCHS,
      6     schema=schema_model,
      7 )

File /usr/local/lib/python3.8/dist-packages/merlin/models/tf/models/base.py:725, in BaseModel.fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq, max_queue_size, workers, use_multiprocessing, train_metrics_steps, **kwargs)
    717 callbacks = self._add_metrics_callback(callbacks, train_metrics_steps)
    719 fit_kwargs = {
    720     k: v
    721     for k, v in locals().items()
    722     if k not in ["self", "kwargs", "train_metrics_steps", "__class__"]
    723 }
--> 725 return super().fit(**fit_kwargs)

File /usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py:70, in filter_traceback.<locals>.error_handler(*args, **kwargs)
     67     filtered_tb = _process_traceback_frames(e.__traceback__)
     68     # To get the full stack trace, call:
     69     # `tf.debugging.disable_traceback_filtering()`
---> 70     raise e.with_traceback(filtered_tb) from None
     71 finally:
     72     del filtered_tb

[...]

PR #724, opened 2022-09-08
Implement `on_epoch_end` method on `BatchedDataset`

oliverholworthy:
Relates to #722 (issue)

### Goals :soccer:

Support Tensorflow/Keras 2.10 with our `BatchedDataset`

- Implementing the `on_epoch_end` method on the `BatchedDataset`. Calling `DataLoader.stop`.
  - I think this is what was happening implicitly previously (a result of calling the `len` method being called - which also calls `DataLoader.stop`).

### Testing Details :mag:

Add a test where the same loader created with `BatchedDataset` is passed to `.fit` more than once.


oliverholworthy:
One other thing I've noticed is that `len` is now being called more frequently in 2.10 (more than once per epoch). So the resetting of the iterator will happen more frequently. 

It seems that the reason for calling `.stop` in `__len__` maybe was added as a work-around for the implementation of `__getitem__` which ignores the `idx` argument and makes the iterator progress, and when `shuffle=True` (the default). Resetting the iterator when `len` is called hides this issue.

https://github.com/NVIDIA-Merlin/models/blob/5b38605f9a68e7925cb7d283fff1f6427d2616a2/merlin/models/tf/dataset.py#L316-L322

oliverholworthy:
Opened another PR as an alternative to this implementation which should improve performance by removing the calls to `.stop` #727

marcromeyn (approved):

oliverholworthy:
rerun tests

oliverholworthy:
For some reason `tests/unit/tf/models/test_base.py::test_train_metrics_steps` is failing in Jenkins. I've been unable to reproduce this locally so far.

oliverholworthy:
Hmm, that test (`tests/unit/tf/models/test_base.py::test_train_metrics_steps`) is now passing. Seems like we may have a flaky test.

PR #727, opened 2022-09-08
Change `BatchedDataset` from Keras Sequence to Generator

oliverholworthy:
Relates to #722 (issue)
Alternative to #724

### Goals :soccer:

- Support Tensorflow/Keras 2.10
- Improve performance of dataloader by removing redundant calls to dataloader `stop` method

- The `BatchedDataset` is currently implementing the interface of [tf.keras.utils.Sequence](https://www.tensorflow.org/api_docs/python/tf/keras/utils/Sequence) 
  - However this class is not a good match for the current generator implementation of BatchedDataset. 
  - This is because the Sequence expects `__getitem__` to be able to return batches by index to support shuffling.
  - If we can't support shuffling like this then it seems natural to remove `Sequence` from the implementation. This will then be supported with the [`GeneratorDataAdapter`](https://github.com/keras-team/keras/blob/506e36a6c967f62342aa49e2828b28edd2a59bc8/keras/engine/data_adapter.py#L826)

### Testing Details :mag:

- Adds a test to check that we can call `.fit` on a model twice with the same dataset.

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -701,11 +701,11 @@ class BaseModel(tf.keras.Model):
         train_metrics_steps=1,
         **kwargs,
     ):
-        x = _maybe_convert_merlin_dataset(x, batch_size, **kwargs)
-
         # Bind schema from dataset to model in case we can't infer it from the inputs
-        if isinstance(x, Loader):
+        if isinstance(x, (Loader, merlin.io.Dataset)):
             self.schema = x.schema
+
+        x = _maybe_convert_merlin_dataset(x, batch_size, **kwargs)
 
         validation_data = _maybe_convert_merlin_dataset(
             validation_data, batch_size, shuffle=shuffle, **kwargs
@@ -1407,5 +1407,8 @@ def _maybe_convert_merlin_dataset(data, batch_size, shuffle=True, **kwargs):
 
         if not shuffle:
             kwargs.pop("shuffle", None)
+
+    if isinstance(data, Loader):
+        data = (batch for batch in data)
 
     return data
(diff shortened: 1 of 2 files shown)


merlin/models/tf/models/base.py
```diff
@@ -1408,4 +1408,7 @@ def _maybe_convert_merlin_dataset(data, batch_size, shuffle=True, **kwargs):
         if not shuffle:
             kwargs.pop("shuffle", None)
 
+    if isinstance(data, Loader):
+        data = (batch for batch in data)
```
oliverholworthy, line 1412: Converting `Loader` into a generator here to ensure it works correctly with the Keras [GeneratorDataAdapter](https://github.com/keras-team/keras/blob/v2.10.0/keras/engine/data_adapter.py#L828). When passing the `Loader` to `GeneratorDataAdapter` it results in incorrect number of batches due to this [peek_and_restore](https://github.com/keras-team/keras/blob/v2.10.0/keras/engine/data_adapter.py#L933-L936) method

oliverholworthy: An alternative or additional option could be to update the `Dataloader` class to be more like an iterator in the way the `next`/`StopIteration` work. 
 
Currently, the `Loader` after raising `StopIteration` once, will start to cycle through the batches again. Whereas a regular python generator will continue to raise `StopIteration` each time `next` is called on a generator with no items left.

viswa-nvidia:
@jperez999 , please track this development and  if there is another PR. that covers this work , please close this appropriately

rnyak:
closing since we have https://github.com/NVIDIA-Merlin/dataloader/pull/65 merged.

PR #740, opened 2022-09-12
Add loader transforms

### LINKED ISSUE #720: [Task] Add ToTarget transform
Right now it’s quite awkward to use certain columns from a schema as a target. Right now we assume that these targets are tagged as part of nvtabular which is not always the case. In order to make this easier we could have a `ToTarget` transform that can be used as part of our data-loader.

### Starting point

```python

class ToTarget(Layer):
	def __init__(
		self, 
		schema,
		*target: Union[Schema, ColSchema, str], 
		one_hot=False
	):
		...

	def call(self, inputs: TabularData, targets=None, testing=False, **kwargs) -> Prediction:
		...

	def compute_output_schema(self, input_schema) -> Schema:
		...
```

edknv:
Fixes #720

### Goals :soccer:
Introduce `Loader`. Introduce `ToTarget`.

- Rename `BatchedDataset` to `Loader`.
- Expose `Loader` in the user-facing API (i.e., `merlin/models/tf/__init__.py`)
- Add a `transform` parameter to `Loader`. This is meant to be an alternative to `Loader.map()` which is somewhat awkward to use.
- Adds a new `transform` op named `ToTarget`. More details in #720.

### Testing Details :mag:
- Changed unit tests that were using `Loader.map(func)` to use the new `Loader(..., transform=func)` interface.
- Added new unit tests for `ToTarget`.

### DIFF
--- a/merlin/models/tf/__init__.py
+++ b/merlin/models/tf/__init__.py
@@ -76,7 +76,6 @@ from merlin.models.tf.core.combinators import (
     SequentialBlock,
 )
 from merlin.models.tf.core.encoder import EncoderBlock
-from merlin.models.tf.dataset import sample_batch
 from merlin.models.tf.inputs.base import InputBlock, InputBlockV2
 from merlin.models.tf.inputs.continuous import Continuous, ContinuousFeatures, ContinuousProjection
 from merlin.models.tf.inputs.embedding import (
@@ -90,6 +89,7 @@ from merlin.models.tf.inputs.embedding import (
     SequenceEmbeddingFeatures,
     TableConfig,
 )
+from merlin.models.tf.loader import KerasSequenceValidator, Loader, sample_batch
 from merlin.models.tf.losses import LossType
 from merlin.models.tf.metrics.topk import (
     AvgPrecisionAt,
@@ -248,4 +248,6 @@ __all__ = [
     "LossType",
     "sample_batch",
     "TensorInitializer",
+    "Loader",
+    "KerasSequenceValidator",
 ]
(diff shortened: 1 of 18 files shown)


merlin/models/tf/loader.py
```diff
@@ -487,13 +506,44 @@ def _build_sparse_tensor(self, values, offsets, diff_offsets, num_rows, seq_limi
...
-class DatasetValidator(tf.keras.callbacks.Callback):
+    @property
+    def output_schema(self) -> Schema:
+        schema = self.input_schema
+
+        for to, transform in self._transforms:
+            if hasattr(transform, "compute_output_schema"):
+                _schema = transform.compute_output_schema(schema)
+            else:
+                raise ValueError(f"Couldn't infer schema from transform {transform}")
```
gabrielspmoreira, line 530: If the `compute_output_schema()` of a Transform block is not implemented, could we not just assume it does not change the schema, instead of raising an exception?

rnyak:
@edknv do you mind to provide details /context in your PR? thanks.

merlin/models/tf/__init__.py
```diff
@@ -97,6 +97,7 @@
     SequenceEmbeddingFeatures,
     TableConfig,
 )
+from merlin.models.tf.loader import KerasSequenceValidater, Loader, sample_batch
```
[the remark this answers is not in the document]
edknv: Did you mean the other way around? `KerasSequenceValidatEr` -> `KerasSequenceValidatOr`. I actually had to look this up, and it looks like `validator` is commonly accepted. I used `KerasSequenceValidatOr`.

merlin/models/tf/loader.py
edknv: My impression so far is proper schema propagation through the layers/blocks is tricky, so I think being explicit is better here than silently assuming the default, but that's just my first impression. Do you think we should assume that the schema is not changed if the method is not implemented?

edknv:
rerun tests

marcromeyn (approved):

PR #745, opened 2022-09-14
Transformations of sequence inputs for sequential/session-based recommendation

### LINKED ISSUE #540: [Task] Target generation for Youtube-DNN retrieval model
### Description
Implement a mapping function to apply on top of the data loader in order to extract targets from the input sequence.  The goal is to define different strategies for setting the target of the Item Prediction task. 

We particularly want to support 4 different strategies: 

image

- `Last`:  extract the last non-padded item from the input sequence  and set it as the target.

- `Sliding window`: Generate multiple targets at different positions based on a sliding-window and extend the input batch with the related truncated input sequences. 

- `Random`: Randomly select a cutting point and set the last item as the target. The input sequence is all interactions from the left context. 

-  `Random held-out`: Randomly select a cutting point as the label.  The input sequence is all interactions from both contexts, left and right. 

**Additional context**
In the Youtube-DNN [paper](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/45530.pdf), they tested the following label selection strategies: 
image

### LINKED ISSUE #691: [TASK] Implement PredictNext, PredictLast & PredictRandom
We need a bunch of augmentations to support session-based recsys. In the examples below we have a sequence: `ABCDE` that we augment in following ways:

#### Prediction-next

| **Inputs**  | A   | B   | C   | D   |
| ------- | --- | --- | --- | --- |
| **Targets** | B   | C   | D   | E   | 

```python

class PredictNext(DataAugmentation):
	def __init__(
		self, 
		schema: Schema
	):
		…

        def compute_output_schema(
		self, 
		input_schema: Schema
	):
		…

```

#### Predict-last

| **Inputs**  | A   | B   | C   | D   |
| ------- | --- | --- | --- | --- |
| **Targets** | E   |     |     |     |

```python
class PredictLast(DataAugmentation):
	def __init__(
		self, 
		schema: Schema
	):
		…

        def compute_output_schema(
		self, 
		input_schema: Schema
	):
		…

```

#### Predict-random
We grow the batch with extra samples where we sample a row_index, use the previous timesteps as features, the next as target.

| **Inputs**  | A   | B   | C   | D   |
| ------- | --- | --- | --- | --- |
| **Targets** | E   |     |     |     |
| **Inputs**  | A   |     |     |     |
| **Targets** | B   |     |     |     |
| **Inputs**  | A   | B   |     |     |
| **Targets** | C    |     |     |     |

```python
class PredictRandom(DataAugmentation):
	def __init__(
		self, 
		schema: Schema
		# causal will control whether or not the elements in the sequence after the radomly picket element will be added to the inputs
		causal=True,
		# When this is bigger than 1, we will grow the batch with more samples
		num_per_batch=1
	):
		…

        def compute_output_schema(
		self, 
		input_schema: Schema
	):
		…
```

gabrielspmoreira:
Fixes #691 , Fixes #540

### Goals :soccer:
Implements the following inputs and target transformations for sequential item prediction:

For RNNs and Causal Language Modeling with Transformers
- [x] SeqPredictNext

For YouTubeDNN (#540)
- [x] SeqPredictLast
- [x] SeqPredictRandom

- The transformations are intended to be used with the `Loader`, for example with  `Loader().map(?)` / `Loader(transform=?)`
- Those transformations are `TabularBlock`, which take in the constructor the schema of the sequential columns that belong to the same group (i.e. represent the same sequence and has the same shape) and the sequential input column (typically the item id) from which the target column will be extracted.

### Testing Details :mag:
- Have added calling the transformations directly and also using the `Loader`
- Also tested `compute_output_shape()` and `compute_output_schema()` methods


gabrielspmoreira:
rerun tests

merlin/models/tf/transforms/sequence.py
```diff
@@ -62,3 +71,281 @@ def call_outputs(
...
+                new_inputs[k] = v[:, :-1]
+            else:
+                new_inputs[k] = v
+
+        return (new_inputs, targets)
+
+
+@Block.registry.register_with_multiple_names("seq_predict_last")
+@tf.keras.utils.register_keras_serializable(package="merlin_models")
+class SeqPredictLast(SequenceTransform):
```
oliverholworthy, line 239: For consistency in the API, we might want to pick one of  `Seq` or `Sequence` across all the layers that refer to this concept. Perhaps the longer version?

merlin/models/tf/transforms/sequence.py
```diff
@@ -62,3 +71,281 @@ def call_outputs(
...
+    pre : Optional[BlockType], optional
+        A block that is called before this method call().
+        If not set, the ListToRagged() block is applied to convert
+        the tuple representation of sequential features to RaggedTensors,
+        so that the tensors sequences can be processed
+    """
+
+    def call(
+        self, inputs: TabularData, targets=None, training=False, testing=False, **kwargs
+    ) -> Union[Prediction, Tuple]:
```
oliverholworthy, line 259: It looks like this will only return `Tuple` in practice. If so we could remove the reference to `Prediction`? Was this here so that this layer could be used inside a model?

merlin/models/tf/transforms/sequence.py
gabrielspmoreira: Makes sense @oliverholworthy . Have renamed the prefix `Seq` to `Sequence`

merlin/models/tf/transforms/sequence.py
gabrielspmoreira: done

marcromeyn (approved):

PR #752, opened 2022-09-19
Filter features in TabularBlock

edknv:
Fixes # (issue)


edknv:
`HashedCross` is problematic for this approach of using the schema to filter the columns, because the columns in the inputs and outputs of `HashedCross` are different. `HashedCross` [adds a new column in the output](https://github.com/NVIDIA-Merlin/models/blob/dc8b5ae33d5e304fb8bbffd1ae936222bb0c6bb5/merlin/models/tf/transforms/features.py#L431). We might need something like `compute_output_schema`.

PR #754, opened 2022-09-20
Making InputBlockV2 more generic

marcromeyn:
### Goals :soccer:
This PR refactors `InputBlockV2` to make it more generic and configurable. It also adds a bunch of examples how to use this block for various use-cases in the doc-strings. 

In this implementation both `continuous` & `categorical` can take in a `Tags` or `Block`. When a column-selector is provided (a schema or tags), we instantiate either `Embeddings` or `Continuous` using the selected columns.

Some ways to use `InputBlockV2`:

#### Simple Usage
```python
 inputs = InputBlockV2(schema)
```

#### Custom Embeddings
```python
inputs = InputBlockV2(
    schema,
    categorical=Embeddings(schema, dim=32)
)
```

#### Sparse outputs for one-hot
```python
inputs = InputBlockV2(
    schema,
    categorical=CategoryEncoding(schema, sparse=True),
    post=ToSparse()
)
```

#### Add continuous projection
```python
inputs = InputBlockV2(
    schema,
    continuous=ContinuousProjection(continuous_schema, MLPBlock([32])),
)
```

#### Merge 2D and 3D (for session-based)
```python
inputs = InputBlockV2(
    context_schema + sequence_schema,
    post=BroadcastToSequence(context_schema, sequence_schema)
)
```


merlin/models/tf/inputs/base.py
```diff
@@ -229,7 +226,7 @@ def InputBlockV2(
         Schema of the input data. This Schema object will be automatically generated using
         [NVTabular](https://nvidia-merlin.github.io/NVTabular/main/Introduction.html).
         Next to this, it's also possible to construct it manually.
-    embeddings : Optional[Block], optional
+    categorical : Optional[Block], optional
```
oliverholworthy, line 229: docstring needs updating to match new arguments and types:
- new arg `continuous`
- typehint here for `categorical` arg. 
- removal of params `continuous_projection` and `continuous_column_selector`

merlin/models/tf/inputs/base.py
```diff
@@ -211,13 +208,13 @@ def InputBlock(
...
-    embeddings: Optional[Block] = None,
-    continuous_projection: Optional[Block] = None,
-    continuous_column_selector: Union[Tags, Schema] = Tags.CONTINUOUS,
+    schema: Optional[Schema] = None,
+    categorical: Union[Tags, Schema, Layer] = Tags.CATEGORICAL,
+    continuous: Union[Tags, Schema, Layer] = Tags.CONTINUOUS,
     pre: Optional[BlockType] = None,
     post: Optional[BlockType] = None,
     aggregation: Optional[TabularAggregationType] = "concat",
+    **branches,
```
oliverholworthy, line 217: how is this `branches` intended to be used?

marcromeyn: Good question. I was looking at [this](https://github.com/NVIDIA-Merlin/models/blob/48b7e4ade146052710796e64c8d7f6425fa53e38/tests/unit/tf/blocks/test_interactions.py#L66-L82) test, when we allow for custom branches in the `InputBlockV2`, that could become something like:

```python
InputBlock(
	schema, 
	categorical=CategoryEncoding(
		cat_schema_onehot, sparse=True, output_mode="one_hot"
	),
	multi_hot=CategoryEncoding.multi_hot(cat_schema_multihot, sparse=True)
)
```

merlin/models/tf/core/combinators.py
```diff
@@ -523,8 +523,18 @@ def build(self, input_shape):
...
                 to_prune.append(name)
 
-        for p in to_prune:
-            self.parallel_layers.pop(p)
+        if isinstance(self.parallel_layers, dict):
+            pruned = {}
+            for name, layer in self.parallel_layers.items():
+                if name not in to_prune:
+                    pruned[name] = layer
+            self.parallel_layers = pruned
```
edknv, line 531: Nit: It looks like one of lines 531 or 537 can be removed and the other un-indented to remove duplicate code?

merlin/models/tf/inputs/base.py
```diff
@@ -211,68 +208,112 @@ def InputBlock(
...
+def parse_branch(
+    branch: Optional[Union[Tags, Schema, Layer]],
+    default: Callable[[Schema], Layer],
+    schema: Optional[Schema] = None,
+) -> Optional[Layer]:
+    if isinstance(branch, Layer) or branch is None:
+        return branch
+
+    if isinstance(branch, Schema):
+        schema = branch
```
edknv, line 313: Nit: Is there an edge case here where both `branch` and `schema` are provided and `schema` is overwritten? Or maybe that's intended or is okay? It looks like this function is a helper function and not intended for use by customers, so this is just a super nitty comment. If it's a helper function for internal use, maybe name it `_parse_branch`?

edknv (approved):

merlin/models/tf/core/combinators.py
marcromeyn: Yeah, we need to rewrite this at some point. Right now it’s a bit weird that `self.parallel_layers` could be either of type `List[Layer]` or `Dict[str, Layer]`. Would be simpler to always use the same type, but might be better in a separate PR.

marcromeyn:
rerun tests

Issue #746, opened 2022-09-14
[Task] Support TensorFlow 2.10


oliverholworthy:
TensorFlow 2.10 was released at the start of September 2022. Test with this version and update for compatibility.

- Update test suite to run with 2.10
  https://github.com/NVIDIA-Merlin/models/pull/726
  https://github.com/NVIDIA-Merlin/models/pull/776
- Fix incompatibilities with new version of tensorflow
  - retrieval models (usage of  `_call_fn_args` no longer works)
    https://github.com/NVIDIA-Merlin/models/pull/726
  - dataloader bug: https://github.com/NVIDIA-Merlin/models/issues/722
    Dataloader doesn't reset correctly after running `.fit`
    - Quick fix:
    https://github.com/NVIDIA-Merlin/models/pull/724
    - Follow-up:
    https://github.com/NVIDIA-Merlin/models/pull/727

PR #726, opened 2022-09-08
Update tensorflow version test matrix version specifiers for clarity 

oliverholworthy:
### Goals :soccer:

- Make Tensorflow Versions in GitHub Actions Clearer

### Implementation Details :construction:

- Changes the GitHub workflow matrix specification for tensorflow versions to improve clarity. (Using the [compatible release clause](https://peps.python.org/pep-0440/#compatible-release))
- ~Update `ItemRetrievalScorer` to use `tf_inspect` to get call args instead of `_call_fn_args` which stopped working in 2.10~ (Update: this was moved to #756 )

### Testing Details :mag:

- Existing test coverage should pass with all versions in the testing matrix

PR #756, opened 2022-09-20
Update `ItemRetrievalScorer` to use tf_inspect to get call args

oliverholworthy:
Relates to #746 

### Goals :soccer:

Support for Tensorflow 2.10

Update `ItemRetrievalScorer` to use tf_inspect to get call args instead of using the now unsupported `_call_fn_args`

### Testing Details :mag:

No changes to tests


marcromeyn (approved):

oliverholworthy:
This was part of #726 however, the teats are failing for the model train step. Possibly something to do with the way the dataset iteration has changed in 2.10.

gabrielspmoreira (approved):
This fixes an error which is raised when training retrieval models with TF 2.10

PR #758, opened 2022-09-21
Bug when predicting with list-column

marcromeyn:
### Goals :soccer:
Right now calling `model.predict` with a batch of data fails when the list-columns are not represented as ragged-tensors. This PR makes sure we use the ragged-tensor representation in `model_test` when `reload_model=True`


oliverholworthy (approved):

marcromeyn:
rerun tests

PR #760, opened 2022-09-21
Update ToTarget to accept Tags and all columns in Schema as targets

edknv:
Followup to #740 

### Goals :soccer:
Add `Tags` as an acceptable type in `ToTarget`. Update it to make all columns in the `target` schema as target columns.

- In the previous implementation, it was necessary for the user to have the Schema columns already tagged as `Tags.TARGET`. This PR changes the logic to assume that the user wants all columns in the provided target schema to be tagged as `Tags.TARGET`.
- The previous change makes the intended usage a bit verbose, e.g., `ToTarget(schema, schema.select_by_tag(Tags.ITEM)`. This PR also adds `Tags` as an acceptable type so that the user could simply use `ToTarget(schema, Tags.ITEM)`. 

### Testing Details :mag:
Unit tests added.


merlin/models/tf/transforms/features.py
```diff
@@ -726,10 +726,13 @@ def _target_column_schemas(self):
                 target_columns[t] = self.schema.select_by_name(t).first
             elif isinstance(t, ColumnSchema):
                 target_columns[t.name] = t
-            elif isinstance(t, Schema):
-                selected_schema = t.select_by_tag(Tags.TARGET)
+            elif isinstance(t, Tags):
+                selected_schema = self.schema.select_by_tag(t)
                 for col_name, col_schema in selected_schema.column_schemas.items():
                     target_columns[col_name] = col_schema
+            elif isinstance(t, Schema):
```
marcromeyn, line 733: Do think we should add the tag `TARGET` to the target-columns here?

One use-case I could think of where that might come in handy is:

```python
loader = mm.Loader(dataset, batch_siz=1000, transform=ToTarget(schema, Tags.ITEM_ID))
schema = loader.schema
model = mm.Model(mm.Encoder(schema, mm.MLPBlock([128, 64]), mm.OutputBlock(schema))
```

marcromeyn: Or would it be better to have `loader.target_schema`?

edknv: We add the `TARGET` tag in [compute_output_schema](https://github.com/edknv/models/blob/fix_to_target_1/merlin/models/tf/transforms/features.py#L771) instead, so with the latest changes in `Loader` for computing the output schema (#740), we should be able to do
```
schema = loader.output_schema
```
and the target columns in `loader.output_schema` will have `Tags.TARGET` added. @marcromeyn That should cover your example use case?

marcromeyn (approved):

PR #761, opened 2022-09-22
Introducing RetrievalModelV2 to unify two-tower & session-based enhancement

marcromeyn:
### Fixes
- #729

### Goals :soccer:
Introduce a new `RetrievalModelV2` that works for mf/two-tower as well as YouTube-dnn + session-based.

CleanShot 2022-10-05 at 12 59 13

Using the `RetrievalModelV2` to create:

#### MF
```python
user = schema.select_by_tag(Tags.USER_ID)
item = schema.select_by_tag(Tags.ITEM_ID)
model = mm.RetrievalModelV2(
        query=mm.EmbeddingEncoder(user, dim=8),
        candidate=mm.EmbeddingEncoder(item, dim=8),
        output=mm.ContrastiveOutput(candidate, "in-batch"),
 )
```

#### Two-tower
```python
user = schema.select_by_tag(Tags.USER)
item = schema.select_by_tag(Tags.ITEM)
model = mm.RetrievalModelV2(
        query=mm.Encoder(user, dim=8),
        candidate=mm.Encoder(item, dim=8),
        output=mm.ContrastiveOutput(item, "in-batch"),
 )
```

#### Session-based style
```python
user = schema.select_by_tag(Tags.USER)
item = schema.select_by_tag(Tags.ITEM_ID)
model = mm.RetrievalModelV2(
        query=mm.EmbeddingEncoder(user, dim=8),
        output=mm.ContrastiveOutput(item, "in-batch"),
)
```

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -16,7 +16,6 @@ from tensorflow.keras.utils import unpack_x_y_sample_weight
 import merlin.io
 from merlin.models.tf.core.base import Block, ModelContext, PredictionOutput, is_input_block
 from merlin.models.tf.core.combinators import ParallelBlock, SequentialBlock
-from merlin.models.tf.core.encoder import Encoder
 from merlin.models.tf.core.prediction import Prediction, PredictionContext
 from merlin.models.tf.core.tabular import TabularBlock
 from merlin.models.tf.inputs.base import InputBlock
@@ -39,6 +38,7 @@ from merlin.models.utils.dataset import unique_rows_by_features
 from merlin.schema import ColumnSchema, Schema, Tags
 
 if TYPE_CHECKING:
+    from merlin.models.tf.core.encoder import Encoder
     from merlin.models.tf.core.index import TopKIndexBlock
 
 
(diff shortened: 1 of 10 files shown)


marcromeyn:
rerun tests

merlin/models/tf/models/base.py
```diff
@@ -1397,6 +1399,146 @@ def to_top_k_recommender(
...
+        pre: Optional[tf.keras.layers.Layer] = None,
+        post: Optional[tf.keras.layers.Layer] = None,
+        **kwargs,
+    ):
+        if isinstance(output, ContrastiveOutput):
+            query_name = output.query_name
+            candidate_name = output.candidate_name
+
+        if query and candidate:
+            encoder = ParallelBlock({query_name: query, candidate_name: candidate})
```
oliverholworthy, line 1421: If I'm understanding this correctly, does this need to be a ParallelBlock, or could it be a dictionary? 

If they don't need to be paired together in that way it might be clearer to follow the the implementation if these were assigned to different properties (`self._query_encoder`, `self._candidate_encoder`)

marcromeyn: It needs to be a Block, but we could implement a custom block. Something lilke:

```python
class BiEncoder(ParallelBlock):
  …

  @property
  def query_encoder(self):
    return self[“query”]

  @property
  def candidate_encoder(self):
    return self[“candidate"]
```


Do you think that’s cleaner?

oliverholworthy: That may help clarify layer names if you were to look at the model representation in that way. 

In this case, I missed the connection to the `__init__` method below where this is the first block of the model.

sararb (approved):

PR #762, opened 2022-09-22
Revert "XGBoost - Use DaskDMatrix for evals data to ensure metrics in logs match result of evaluate"

oliverholworthy:
Reverts NVIDIA-Merlin/models#682

We're getting an error in our container builds from `test_gpu_hist_dmatrix`. 

I don't know for sure if this PR caused it and don't have an explanation for how this change relates to the error, but based on the timing it could be. And can't think of any other things to try to fix it for now apart from reverting this change. 

```
tests/unit/xgb/test_xgboost.py ........FF..........                      [100%]

=================================== FAILURES ===================================
_________ test_gpu_hist_dmatrix[fit_kwargs0-DaskDeviceQuantileDMatrix] _________

mock_train = <MagicMock name='train' id='140317824645680'>, fit_kwargs = {}
expected_dtrain_cls = <class 'xgboost.dask.DaskDeviceQuantileDMatrix'>
dask_client = <Client: 'tcp://127.0.0.1:39981' processes=5 threads=10, memory=20.00 GiB>

    @pytest.mark.skipif(not HAS_GPU, reason="No GPU available")
    @pytest.mark.parametrize(
        ["fit_kwargs", "expected_dtrain_cls"],
        [
            ({}, xgboost.dask.DaskDeviceQuantileDMatrix),
            ({"use_quantile": False}, xgboost.dask.DaskDMatrix),
        ],
    )
    @patch("xgboost.dask.train", side_effect=xgboost.dask.train)
    def test_gpu_hist_dmatrix(mock_train, fit_kwargs, expected_dtrain_cls, dask_client):
        train, valid = generate_data("music-streaming", num_rows=100, set_sizes=(0.5, 0.5))
        schema = train.schema
        model = XGBoost(schema, objective="reg:logistic", tree_method="gpu_hist")
>       model.fit(train, evals=[(valid, "valid")], **fit_kwargs)

tests/unit/xgb/test_xgboost.py:128: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
merlin/models/xgb/__init__.py:180: in fit
    train_res = xgb.dask.train(
/usr/lib/python3.8/unittest/mock.py:1081: in __call__
    return self._mock_call(*args, **kwargs)
/usr/lib/python3.8/unittest/mock.py:1085: in _mock_call
    return self._execute_mock_call(*args, **kwargs)
[...]

PR #763, opened 2022-09-22
Remove `cluster_type` from `dask_client` test fixture

oliverholworthy:
Removes `cluster_type` from `dask_client` test fixture.

This was initially set to be CPU only because the tests would not work in a CPU-only environment due to the way the cpu checks were working in the merlin core module (context [here](https://github.com/NVIDIA-Merlin/models/pull/466#discussion_r883895804) ).

This  is no longer required, and may be causing issues with the GPU version of the tests when  `tree_method="gpu_hist"` (see: #762)

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -64,7 +64,7 @@ def testing_data() -> Dataset:
 
 @pytest.fixture(scope="module")
 def dask_client() -> distributed.Client:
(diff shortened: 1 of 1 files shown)


oliverholworthy:
rerun tests

marcromeyn (approved):

PR #767, opened 2022-09-26
Skip running test for traditional ML notebook unless GPU available

oliverholworthy:
Supports https://github.com/NVIDIA-Merlin/core/pull/131

### Goals :soccer:

Run test for traditional ML notebook only when requirements satisfied.

### Testing Details :mag:

The notebook being tested requires GPU and supporting libraries (`cupy`)  to be installed to function correctly because it uses the `gpu_hist` tree method with XGBoost. 

Adding this skip based on the GPU check ensures we don't try to run the test unless the GPU requirements are satisfied


marcromeyn:
So does this mean that for libraries like xgboost, we’re GPU only?

oliverholworthy:
> So does this mean that for libraries like xgboost, we’re GPU only?

No XGBoost works on CPU too. The reason for this PR is that in this notebook we're activating a GPU-only codepath (with the `gpu_hist` tree method). And without this skip, it will attempt to run and fail in an evironment where the library dependencies are available (e.g. https://github.com/NVIDIA-Merlin/core/pull/131 )

An alternative would be to update the notebook to put some logic in that allows it to run on both CPU and GPU.

marcromeyn:
Am I correct in saying that as of now except `07-Train-traditional-ML-models-using-the-Merlin-Models-API` all notebooks are for both CPU + GPU? If that’s the case, I think it would make sense to update this notebook to make it work in both cases as well. What do you think @bschifferer @rnyak @radekosmulski?

marcromeyn (approved):

PR #768, opened 2022-09-26
Adding transformers as backend

marcromeyn:
### Goals :soccer:
This PR adds Huggingface transformers as a optional backend.

### Implementation Details :construction:
Also adds test-markers to give users the ability to only run the huggingface-related tests.

PR #771, opened 2022-09-27
Transformer Block (for HuggingFace support)

### LINKED ISSUE #721: Port TransformerBlock

sararb:
Fixes #721 

### Blocked by
- #768

### Goals :soccer:
- [x] Port TransformerBlock from transformers4rec
- [x] Implement specific HF transformer blocks: Bert, Albert, Roberta, XLNet, and GPT2

- Port the TransformerBlock from T4Rec and refactor it based on the latest HuggingFace code
- Add `pre` layer to support preparing additional inputs of the transformer model such as `attention_mask` or `head_mask`. This layer has access to the sequence of input embeddings and raw features. 
- Add a post with an `output_fn` function to process the [TFBaseModelOutputWithPoolingAndCrossAttentions](https://github.com/huggingface/transformers/blob/0b567aa430e9b1b56254de4775db5868104d34e5/src/transformers/modeling_tf_outputs.py#L125) return by HF layer and return the desired tensor(s). 
- Add custom transformers blocks (bert, albert..) with the `build_transformer_layer` method that prepares the HF configuration class with unified names of the required parameters `(d_model, n_head, n_layer,max_seq_length)`

### Testing Details :mag:
Add unit tests of the transformer block: 

- Test the TransformerBlock as a sequence encoder
- Test pooling the sequence of hidden states returned by the transformer model to a 2-D tensor
- Test different transformer encoding models: Bert, Albert, XLnet, RoBerta, GPT2.
- Test a TransformerBlock in a classification model
- Test save and load back classification model with transformer encoder block
- Test prepare additional inputs to the TransformerBlock

### Open questions/remarks 

- HuggingFace does not support ragged representation, so the input batch of interaction embeddings is converted to a dense tensor within the TransformerBlock call. 
- In case of masking is being used, How to access the boolean mask tensor?


merlin/models/tf/transformers/transformer_block.py
```diff
@@ -0,0 +1,548 @@
+#
```
marcromeyn, line 1: Renaming this file to `block.py` is probably more in-line with the rest of our naming

merlin/models/tf/transformers/transformer_block.py
```diff
@@ -0,0 +1,548 @@
...
+    Extract serializable custom keras layer `TF*MainLayer` from the HuggingFace model
+    """
+    main_layer = [v for _, v in hf_model.__dict__.items() if isinstance(v, tf.keras.layers.Layer)][
+        0
+    ]
+    return main_layer
+
+
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class PrepareTransformerInputs(tf.keras.layers.Layer):
```
marcromeyn, line 53: Would it make sense to move this to `transforms.py`?

merlin/models/tf/transformers/block.py
```diff
@@ -0,0 +1,557 @@
...
+
+
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class TransformerBlock(Block):
+    """
+    Class to support HF Transformers for session-based and sequential-based recommendation models.
+    Parameters
+    ----------
+    transformer: TransformerBody
+        The T4RecConfig, The pre-trained HF model or the custom keras layer TF*MainLayer,
```
oliverholworthy, line 60: Is `T4RecConfig` valid here? Or can that part be removed

merlin/models/tf/__init__.py
```diff
@@ -122,6 +122,20 @@
 from merlin.models.tf.prediction_tasks.multi import PredictionTasks
 from merlin.models.tf.prediction_tasks.regression import RegressionTask
 from merlin.models.tf.prediction_tasks.retrieval import ItemRetrievalTask
+from merlin.models.tf.transformers.block import (
```
oliverholworthy, line 125: If the idea is to make transformers an optional dependency. We may not be able to add these imports to this top-level module in this way?

marcromeyn: Good catch! We should wrap this in a try/catch to handle the case where `transformers` is not installed.

tests/unit/tf/transformers/test_block.py
```diff
@@ -0,0 +1,132 @@
...
+
+    schema = sequence_testing_data.schema.select_by_name(
+        ["item_id_seq", "categories", "user_country"]
+    )
+    schema["user_country"] = schema["user_country"].with_tags(
+        schema["user_country"].tags + "target"
+    )
+    sequence_testing_data.schema = schema
+    dataloader = mm.Loader(sequence_testing_data, batch_size=50)
+    dataloader = dataloader.map(_target_to_onehot)
```
edknv, line 131: We added an optional parameter `transform` to `Loader` which essentially replaces `.map()` so these two lines can be shortened to
```
dataloader = mm.Loader(sequence_testing_data, batch_size=50, transform=_target_to_onehot)
```

merlin/models/tf/transformers/block.py
sararb: I removed the mention of `T4RecConfig`.

tests/unit/tf/transformers/test_block.py
sararb: Ah thank you, I didn't know that option! I have updated the test with your suggestion

merlin/models/tf/__init__.py
sararb: I added a dependency check `is_transformers_available()` to import the blocks only when `transformers` is installed.

oliverholworthy (approved):

PR #774, opened 2022-09-28
Extend ListToDense to support a single tensor as input

sararb:
### Goals :soccer:
Currently, ListToDense only supports TabularData (dict of tensors) as inputs. However, in session-based, we want to use this operator to convert the ragged sequence representation to a dense tensor with a specific `max_sequence_length` before feeding it to a sequential module such as `Transformer`.  

- [x] Support single 2-D or 3-D ragged tensor as input
- [x] Support a tuple of values/row_lengths as input

- Add the private method `_convert_tensor_to_dense` that converts a ragged tensor or a tuple of values/row_lengths to a dense tensor. All tensors are padded and/or truncated to `max_seq_length` 
- Add the private method `_get_output_tensor_shape` to compute the right output shape 

### Testing Details :mag:
- Add a unit test `test_list_to_dense()` to `test_tensor.py`

### DIFF
--- a/merlin/models/tf/transforms/tensor.py
+++ b/merlin/models/tf/transforms/tensor.py
@@ -122,12 +122,13 @@ class ListToDense(TabularBlock):
     def _get_output_tensor_shape(self, val_shape):
         if isinstance(val_shape, tuple) and isinstance(val_shape[1], tf.TensorShape):
             val_shape = val_shape[1]
+            return tf.TensorShape([val_shape[0], self.max_seq_length])
         if val_shape.rank > 1 and val_shape[-1] > 1:
             shapes = val_shape.as_list()
             if self.max_seq_length:
                 shapes[1] = self.max_seq_length
             return tf.TensorShape(shapes)
-        return tf.TensorShape(val_shape.as_list()[0])
+        return tf.TensorShape((val_shape[0]))
 
     def _convert_tensor_to_dense(self, val):
         if isinstance(val, tuple):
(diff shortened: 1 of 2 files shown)


merlin/models/tf/transforms/tensor.py
```diff
@@ -101,33 +101,48 @@ def __init__(self, max_seq_length: Optional[int] = None, **kwargs):
...
+    def _convert_tensor_to_dense(self, val):
+        if isinstance(val, tuple):
+            val = list_col_to_ragged(val)
+        if isinstance(val, tf.RaggedTensor):
+            if self.max_seq_length:
+                shape = [None] * val.shape.rank
+                shape[1] = self.max_seq_length
+                val = val.to_tensor(shape=shape)
+            else:
+                val = tf.squeeze(val.to_tensor())
```
oliverholworthy, line 142: It's not clear to me why calling squeeze is related to conversion from ragged to dense. However, this was already here so maybe something to review another time.

oliverholworthy (approved):

merlin/models/tf/transforms/tensor.py
sararb: Right, I suspect it is needed to remove any trailing dimension =`1` we can get from the data loader's multi-hot outputs but I couldn't reproduce this specific case. I also removed `tf.squeeze` and re-run the tests, all of them passed.

sararb:
rerun tests

PR #775, opened 2022-09-29
Implemented Masked Language Modeling (alternative #1)

### LINKED ISSUE #692: [TASK] Implement PredictMasked (BERT-like masking)
# BERT-like masking

Let’s say we have a sequence `ABCDE`, BERT-like masking would result in the following:
| **Inputs**  | A        | `MASKED` | C        | `MASKED` | E        |
| ------- | -------- | -------- | -------- | -------- | -------- |
| **Targets** | `MASKED` | B        | `MASKED` | D        | `MASKED` |

_Note, the number of target items might differ for different samples in the batch. We should ensure that we have at least one target and at most `len(seq) - 1`_

The class that’s responsible for masking could roughly look like:

```python
class PredictMasked(DataAugmentation):
	def __init__(
		self, 
		schema: Schema
		prediction_block=None,
		mask_selection_rate,
		mask_selection_length,
		unselectable_token_ids=[0],
		mask_token_rate=0.8,
		random_token_rate=0.1
	):
		...

	def compute_mask(self, ...):
		...
```

We want to make use of standard Keras functionality w.r.t. masking. Some useful links:
- [Mask propagation](https://keras.io/guides/understanding_masking_and_padding/)
- [MLMMaskGenerator](https://github.com/keras-team/keras-nlp/blob/master/keras_nlp/layers/mlm_mask_generator.py#L20) in [keras_nlp](https://github.com/keras-team/keras-nlp)
	- Works on normal + `tf.RaggedTensor`s
	- Uses `tensorflow_text`
- [mask_language_model](https://github.com/tensorflow/text/blob/master/tensorflow_text/python/ops/masking_ops.py#L91) in [tensorflow_text](https://github.com/tensorflow/text)

gabrielspmoreira:
Fixes #692 

### Goals :soccer:
Implements the masked language modeling technique for training Transformers, like originally proposed for BERT (NLP) and later adapted to RecSys by [BERT4Rec](https://arxiv.org/abs/1904.06690).

The technique was implemented as a transformation class (`SequenceRandomTargetMasking`).
It is intended to be used as a `pre` of `TransformerBlock` (introduced in PR #771) for architectures that support masked language modeling  
For example:

```python
mlm = SequenceRandomTargetMasking(schema, masking_prob=0.15)
transformer = mm.TransformerBlock(
        transformer=BertConfig(hidden_size=EMBED_DIM, num_attention_heads=16),
        pre=mlm
)
```

The block randomly selects (masks) item ids from the sequence to be targets and replaces the input embeddings at the masked positions by a trainable dummy embedding, so that we don't leak targets information.
It returns a `result = Prediction(inputs_replaced, targets)`.   
The targets mask can be accessed via `result.targets._keras_mask` or `mlm.target_mask`. The input masks are the opposite of the target masks and can be accessed via `results.outputs._keras_mask`.

### Testing Details :mag:
I have implemented a handful of tests that checks the correct behaviour of MLM and also asserts the exceptions.

### DIFF
--- a/merlin/models/tf/__init__.py
+++ b/merlin/models/tf/__init__.py
@@ -122,6 +122,7 @@ from merlin.models.tf.prediction_tasks.classification import (
 from merlin.models.tf.prediction_tasks.multi import PredictionTasks
 from merlin.models.tf.prediction_tasks.regression import RegressionTask
 from merlin.models.tf.prediction_tasks.retrieval import ItemRetrievalTask
+from merlin.models.tf.transformers.blocks import SequenceRandomTargetMasking
 from merlin.models.tf.transforms.features import (
     CategoryEncoding,
     HashedCross,
@@ -252,4 +253,5 @@ __all__ = [
     "BroadcastToSequence",
     "Loader",
     "KerasSequenceValidator",
+    "SequenceRandomTargetMasking",
 ]
--- a/merlin/models/tf/transformers/blocks.py
+++ b/merlin/models/tf/transformers/blocks.py
@@ -160,7 +160,6 @@ class SequenceRandomTargetMasking(Block):
             target_mask = tf.where(
                 tf.expand_dims(replacement_cond, -1), one_target_mask, target_mask_by_prob
             )
-
             padding_mask = tf.sequence_mask(row_lengths)
             target_mask_ragged = tf.ragged.boolean_mask(target_mask, padding_mask)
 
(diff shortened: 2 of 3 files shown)


gabrielspmoreira:
I have implemented an alternative implementation in #780 according to some suggestions from @marcromeyn . Only one of these PRs will be merged

gabrielspmoreira:
Closing this draft PR, as a preferred design was implemented in #780

PR #778, opened 2022-09-29
[WIP] multi-gpu example/bug

bschifferer:
This PR shows how to use merlin models with multi-GPU - this is a simple demonstration. It require more work to provide data parallel support


bschifferer:
I close this PR, as this was only for demonstration of the functionality + bug

PR #779, opened 2022-09-29
Add test for `unique_rows_by_features` and update to propagate schema correctly

oliverholworthy:
### Goals :soccer:

- Ensure that helper function `unique_rows_by_features` works correctly.
  For the case where you would like to generate embeddings from an `Encoder` (see #761) provided a training dataset that contains duplicate rows of user or item features.

- Move implementation to new function `unique_by_tag` and update `unique_rows_by_features` to be alias with deprecation warning.
- propagate the original schema through to the resulting dataset returned
- enables `Tags.ID` to be used as the `grouping_tag` by ensuring we perform the selection on the subset of features selected.

### Testing Details :mag:

- Checks that rows with the same ID spread across different partitions are handled correctly.
- Checks that the correct schema is returned on the resulting dataset.


merlin/models/utils/dataset.py
```diff
@@ -52,16 +52,34 @@ def dataset_to_coo(dataset: Dataset):
 
 def unique_rows_by_features(
     dataset: Dataset, features_tag: Union[str, Tags], grouping_tag: Union[str, Tags]
```
marcromeyn, line 54: Would it make sense to make `Tags.ID` the default for `grouping_tag`?

oliverholworthy: yes, I think that would be good if that's the main use-case we have for this helper.

merlin/models/utils/dataset.py
```diff
@@ -50,18 +51,66 @@ def dataset_to_coo(dataset: Dataset):
...
     if columns:
-        id_col = dataset.schema.select_by_tag(grouping_tag).first.name
+        id_col = features_schema.select_by_tag(grouping_tag).first.name
         ddf = ddf[columns].drop_duplicates(id_col, keep="first")
 
-    return Dataset(ddf)
+    return Dataset(ddf, schema=features_schema)
+
+
+def unique_features_by_row(
```
marcromeyn, line 87: Shouldn’t this be `unique_rows_by_features`?

oliverholworthy: Yes it should be

marcromeyn (approved):

merlin/models/utils/dataset.py
oliverholworthy: Updated

PR #780, opened 2022-09-29
Introduces Masked Language Modeling for Transformers

### LINKED ISSUE #692: [TASK] Implement PredictMasked (BERT-like masking)

gabrielspmoreira:
Fixes #692 

### Goals :soccer:
Implements the masked language modeling (MLM) technique for training Transformers, like originally proposed for BERT (NLP) and later adapted to RecSys by [BERT4Rec](https://arxiv.org/abs/1904.06690).

**Note**: This PR is the alternative implementation to #775 (which won't be merged). 

The MLM training with Transformers is implemented in two steps:
1 - Extract the targets from the sequential input item ids and selects (masks) the targets: `SequenceMaskRandom`
It should be used as a `pre` of `model.fit(..., pre=SequenceMaskRandom())`  

2 - Replaces the input embeddings at masked positions by a dummy embedding: `ReplaceMaskedEmbeddings`. It is meant to be used as a `pre` of `TransformerBlock`.

####  Notes
**Note #1**: The `SequenceMaskRandom` sets the mask in both inputs and targets. The inputs mask might be later lost because of the usage of ops that dont support Keras Masking, or that are mask producers (e.g. the tf.keras.layers.Embedding).
 The `targets` mask is forwarded to the following layers, and can be used by `ReplaceMaskedEmbeddings` (to mask input embeddings) and also the `OutputBlock` (to mask the loss)
**Note #2**: The `SequenceMaskRandom` was originally planned to be used  as a transform of `Loader` like `SequencePredictNext`. But the `._keras_mask` set on target and inputs tensors is lost when the tensors are provided to the model in graph mode. So we have to use `SequenceMaskRandom` as a  `pre` of `model.fit(..., pre=SequenceMaskRandom())` .
Example usage:

```python
seq_schema = sequence_testing_data.schema.select_by_tag(Tags.SEQUENCE).select_by_tag(
        Tags.CATEGORICAL
    )
target = sequence_testing_data.schema.select_by_tag(Tags.ITEM_ID).column_names[0]
loader = Loader(sequence_testing_data, batch_size=128, shuffle=False)
model = mm.Model(
    mm.InputBlockV2(
        seq_schema,
        embeddings=mm.Embeddings(
            seq_schema.select_by_tag(Tags.CATEGORICAL), sequence_combiner=None
        ),
    ),
    BertBlock(
        d_model=48,
        n_head=4,
        n_layer=2,
        pre=mm.ReplaceMaskedEmbeddings(),
    ),
    mm.CategoricalOutput(
        seq_schema.select_by_name(target), default_loss="categorical_crossentropy"
    ),
)
)

model.compile(run_eagerly=False, optimizer="adam")

# Trains on randomly masked positions
seq_mask_random = mm.SequenceMaskRandom(schema=seq_schema, target=target, masking_prob=0.3)
model.fit(loader, epochs=epochs, steps_per_epoch=1, pre=seq_mask_random)

# Evaluates to predict the last position (to mimic next-item prediction)
seq_mask_last = mm.SequenceMaskLast(schema=seq_schema, target=target)
metrics = model.evaluate(
        loader, steps=1, return_dict=True, pre=seq_mask_last
    )
```

#### Masked losses and metrics
This PR also adds support of Keras masking and 3D predictions to losses and metrics.  
Before computing the loss and metrics, the ragged targets are converted to dense, and mask is copied  from the target to the predictions tensor, because Keras considers the mask of predictions tensor when computing the loss (ignoring False values). Finally, the target is automatically converted to one-hot encoding if it is not, as the sparse labels is a requirement from many losses and metrics

### Testing Details :mag:
I have implemented a handful of tests that checks the correct behaviour of MLM and also asserts the exceptions.

### DIFF
--- a/merlin/models/tf/transforms/sequence.py
+++ b/merlin/models/tf/transforms/sequence.py
@@ -16,9 +16,11 @@
 from typing import Optional, Tuple, Union
 
 import tensorflow as tf
+from tensorflow.keras.backend import random_bernoulli
 
 from merlin.models.tf.core.base import Block, BlockType, PredictionOutput
 from merlin.models.tf.core.combinators import TabularBlock
+from merlin.models.tf.core.prediction import Prediction
 from merlin.models.tf.transforms.tensor import ListToRagged
 from merlin.models.tf.typing import TabularData
 from merlin.models.tf.utils import tf_utils
@@ -77,7 +79,6 @@ class SequenceTransform(TabularBlock):
     """Prepares sequential inputs and targets for next-item prediction.
     The target is extracted from the shifted sequence of item ids and
     the sequential input features are truncated in the last position.
-
     Parameters
     ----------
     schema : Schema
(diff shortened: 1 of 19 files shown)


merlin/models/tf/transforms/sequence.py
```diff
@@ -348,3 +349,341 @@ def call(
...
+
+    def build(self, input_shape):
+        if len(input_shape) != 3:
+            raise ValueError("The inputs must be a 3D tensor (batch_size, seq_length, vector_dim)")
+        self.hidden_size = input_shape[-1]
+        if self.hidden_size is None:
+            raise ValueError("The last dim of inputs cannot be None")
+        # Create a trainable embedding to replace masked interactions
+        initializer = tf.random_normal_initializer(mean=0.0, stddev=0.001)
+        self.masked_item_embedding = tf.Variable(
```
marcromeyn, line 560: `item` is a bit of an overloaded term in Recsys. Would something like `masked_timestep_embedding` be a better name?

marcromeyn:
rerun tests

marcromeyn:
rerun tests

marcromeyn (approved):

PR #783, opened 2022-10-03
[WIP] Add horovod for distributed training

edknv:
WIP. Opening PR to run the CI.

Fixes # (issue)

### DIFF
--- a/merlin/models/tf/distributed/backend.py
+++ b/merlin/models/tf/distributed/backend.py
@@ -1,0 +1,10 @@
+hvd = None
+
+try:
+    import horovod.tensorflow.keras as hvd  # noqa: F401
+except ImportError:
+    pass
+
+
+if hvd:
+    hvd.init()
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -18,6 +18,7 @@ from merlin.models.tf.core.base import Block, ModelContext, PredictionOutput, is
 from merlin.models.tf.core.combinators import SequentialBlock
 from merlin.models.tf.core.prediction import Prediction, PredictionContext
 from merlin.models.tf.core.tabular import TabularBlock
+from merlin.models.tf.distributed.backend import hvd
 from merlin.models.tf.inputs.base import InputBlock
 from merlin.models.tf.loader import Loader
 from merlin.models.tf.losses.base import loss_registry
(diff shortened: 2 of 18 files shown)


merlin/models/tf/models/base.py
```diff
@@ -329,6 +337,27 @@ def compile(
...
+            if isinstance(optimizer, merlin.models.tf.DistributedOptimizer):
+                # TODO
+                pass
+            if isinstance(optimizer, merlin.models.tf.MultiOptimizer):
+                # TODO
+                pass
+            else:
+                if isinstance(optimizer, str):
+                    optimizer = tf.keras.optimizers.get(optimizer)
+                scaled_lr = optimizer.learning_rate * hvd.size()
```
bschifferer, line 351: I think we need to do this optional / adding a scaling parameter? In that way, it is enforced by our library. I dont know if we should do this?

merlin/models/tf/distributed/backend.py
```diff
@@ -0,0 +1,27 @@
...
+from merlin.core.dispatch import HAS_GPU
+
+hvd = None
+multi_gpu = False
+
+if HAS_GPU:
+    try:
+        import horovod.tensorflow.keras as hvd
+
+        multi_gpu = True
```
bschifferer, line 12: It seems that we will do always multi_gpu training, is that correct? I can see a case, where we have multiple GPUs on 1 machine but we want to do single GPU training. E.g. I have 4x GPUs, but I want to run 4x independent experiments in parallel

merlin/models/tf/models/base.py
```diff
@@ -121,6 +122,13 @@ def fit(
             validation_data, batch_size, shuffle=shuffle, **kwargs
         )
         callbacks = self._add_metrics_callback(callbacks, train_metrics_steps)
+        if multi_gpu:
+            callbacks = self._add_distributed_callbacks(callbacks)
+        # Horovod: write logs on worker 0.
+        verbose = verbose if hvd.rank() == 0 else 0
+        # Horovod: adjust number of steps based on number of GPUs.
+        if steps_per_epoch:
+            steps_per_epoch = steps_per_epoch // hvd.size()
```
bschifferer, line 131: We should add a warning, if we drop some batches/examples due to //hvd.size() ?

merlin/models/tf/distributed/backend.py
edknv: In this draft, I was thinking users could just set the global variable to False, e.g., `distributed.multi_gpu = False`, if they wanted to do single GPU training. But now I think we should drop this entirely. Users would be running the script with `horovodrun -np N script.py`, so if they want to do multi-gpu training, they can control the number of processes with `-np`. Or, if they simply do not use `horovodrun` or `mpirun`, it would just use a single GPU and `hvd.size()` would be `1`. Some of my thoughts here: https://github.com/NVIDIA-Merlin/Merlin/issues/651#issuecomment-1272785284

PR #790, opened 2022-10-06
Retrieval models definition using new RetrievalModelV2, InputBlockV2 and ModelOutput classes

### LINKED ISSUE #611: [Task] Update the high-level API  of retrieval models with new core API
# 🚀 Feature request

The objective is to update retrieval models with the new core API classes once they are completed and tested. The feature includes updating the code base and adding related unit tests.  
Note: The old ItemRetrievalTask and ItemRetrievalScorer will get removed. 

This will fix #498

### LINKED ISSUE #805: Update the retrieval integration tests on CI to use the refactored MF and TwoTower

sararb:
Fixes #611 
Fixes #805

### Goals :soccer:
- The objective of this PR is to implement a new version of high-level retrieval models API (`MatrixFactorizationModelV2`, `TwoTowerModelV2`, and `YoutubeDNNRetrievalModelV2`) using the new core API (such as RetrievalModelV2, InputBlockV2, and ModelOutput)

- The tested functionalities for the three models are: 
         - Saving/loading retrieval models
         - Saving /loading Top-k encoders defined from the trained retrieval models.  
         - Computing top-k evaluation of retrieval models 
         - Exporting item embeddings and the query tower.

- [x] Introduce new retrieval methods: MatrixFactorizationModelV2, TwoTowerModelV2, and YoutubeDNNRetrievalModelV2
- [x] Ensure embeddings export is working for MatrixFactorizationModelV2 and TwoTowerModelV2
- [x] Link Top-K encoder to retrieval models. 

### Testing Details :mag:
- [x] Update all unit tests of retrieval models with the new API. 
- [x] Test export towers and embeddings works for MatrixFactorizationModelV2 and TwoTowerModelV2. 
- [x] Test top-k recommendation and evaluation of MatrixFactorizationModelV2 and TwoTowerModelV2. 
- [x] Test top-k recommendation and evaluation of YoutubeDNNRetrievalModelV2. 
- [x] Test MatrixFactorizationV2 and TwoTowerModelV2 with Binary and Regression tasks. 
- [x] Add integration tests for MatrixFactorizationV2 and TwoTowerModelV2.

**Notes:** 
- This PR does not refactor the old retrieval API 

- While exporting `candidates embeddings` is working, exporting the query embeddings from `YoutubeDNNRetrievalModelV2` raises an error related to the `batch-predict` step. The error (more details can be found in the issue #801) is related to the apply method of dask-df does not support a list column with an unspecified dtype. 

- Integration tests showed a mismatch between the evaluation scores of the old and new API. After debugging the API, I found three differences that needed to be fixed in the new API (embedding regularization (fixed [here](https://github.com/NVIDIA-Merlin/models/pull/812)), log-q correction (fixed [here](https://github.com/NVIDIA-Merlin/models/pull/811)), and logits scaling (fixed [here](https://github.com/NVIDIA-Merlin/models/pull/815)) )

### DIFF
--- a/merlin/models/tf/models/retrieval.py
+++ b/merlin/models/tf/models/retrieval.py
@@ -1,14 +1,23 @@
+import warnings
 from typing import Any, Callable, Dict, List, Optional, Sequence, Union
+
+import tensorflow as tf
 
 from merlin.models.tf.blocks.mlp import MLPBlock
 from merlin.models.tf.blocks.retrieval.matrix_factorization import QueryItemIdsEmbeddingsBlock
 from merlin.models.tf.blocks.retrieval.two_tower import TwoTowerBlock
 from merlin.models.tf.blocks.sampling.base import ItemSampler
 from merlin.models.tf.core.base import Block, BlockType
-from merlin.models.tf.inputs.base import InputBlock
+from merlin.models.tf.core.combinators import ParallelBlock
+from merlin.models.tf.core.encoder import EmbeddingEncoder, Encoder
+from merlin.models.tf.inputs.base import InputBlock, InputBlockV2
 from merlin.models.tf.inputs.embedding import EmbeddingOptions
-from merlin.models.tf.models.base import Model, RetrievalModel
+from merlin.models.tf.models.base import Model, RetrievalModel, RetrievalModelV2
 from merlin.models.tf.models.utils import parse_prediction_tasks
+from merlin.models.tf.outputs.base import DotProduct, ModelOutput
+from merlin.models.tf.outputs.contrastive import ContrastiveOutput
+from merlin.models.tf.outputs.sampling.base import ItemSamplersType
+from merlin.models.tf.outputs.sampling.popularity import PopularityBasedSamplerV2
 from merlin.models.tf.prediction_tasks.base import ParallelPredictionBlock, PredictionTask
 from merlin.models.tf.prediction_tasks.next_item import NextItemPredictionTask
 from merlin.models.tf.prediction_tasks.retrieval import ItemRetrievalTask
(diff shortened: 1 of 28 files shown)


merlin/models/tf/models/retrieval.py
```diff
@@ -300,3 +309,307 @@ def YoutubeDNNRetrievalModel(
...
+        By default 100
+    min_sampled_id: int, optional
+        The minimum id value to be sampled with sampled softmax.
+        Useful to ignore the first categorical
+        encoded ids, which are usually reserved for <nulls>,
+        out-of-vocabulary or padding.
+        By default 0.
+    """
+    if not inputs:
+        inputs = InputBlockV2(schema, **kwargs)
```
oliverholworthy, line 591: Do we need this reference to InputBlockV2 if the automatic construction of the input block is happening in the `Encoder`?

oliverholworthy:
I don't think we can close this issue https://github.com/NVIDIA-Merlin/models/issues/498 (in the sense that the issue is referring to the specific version of `TwoTowerModel` that will continue to have the loading issue even with this new version?). So that issue might exist until we've removed the old `TwoTowerModel` and migrated all the examples to use the new one?

sararb:
> I don't think we can close this issue #498 (in the sense that the issue is referring to the specific version of `TwoTowerModel` that will continue to have the loading error even with this new version?). So that issue might exist until we've removed the old `TwoTowerModel` and migrated all the examples to use the new one?

That's a good point! The focus of this PR is to implement a new version of retrieval models, such as `TwoTowerModelV2`. Removing the old API and updating the examples should happen in a separate PR. I removed the issue from the PR description and just add a comment about using TwoTowerModelV2 as an alternative solution.

merlin/models/tf/models/retrieval.py
```diff
@@ -300,3 +309,277 @@ def YoutubeDNNRetrievalModel(
...
+        By default 0.
+    """
+    if not inputs:
+        inputs = schema
+
+    candidate = schema.select_by_tag(candidate_id_tag)
+    if not candidate:
+        raise ValueError(f"The schema should contain a feature tagged as `{candidate_id_tag}`")
+    candidate = candidate.first
+    num_classes = candidate.int_domain.max + 1
```
gabrielspmoreira, line 567: I think we should try to avoid the access to custom properties of the ColumnSchema (like `int_domain.max + 1`) as we might want to change those properties in the feature and there will be many places to update (which is error-prone( and instead use functions to get info like feature cardinality.
We could use instead the `schema_utils.categorical_cardinalities()` instead.

merlin/models/tf/models/retrieval.py
```diff
@@ -300,3 +309,277 @@ def YoutubeDNNRetrievalModel(
...
+
+    query = Encoder(inputs, top_block, post=post)
+
+    if not outputs:
+        warnings.warn(
+            "By default, the YoutubeDNN model is trained using popularity-bases sampled softmax."
```
gabrielspmoreira, line 573: I think don't need the warning in this case. The docstrings is enough for interested users.

merlin/models/tf/models/retrieval.py
```diff
@@ -300,3 +309,277 @@ def YoutubeDNNRetrievalModel(
...
+        )
+        candidate_table = query.first["categorical"][candidate.name]
```
gabrielspmoreira, line 575: I understand that `.first()` here is a default keras method for (Encoder) Model that returns the first layer (block) of the model. There can be a case where the InputBlock wont' be the first layer?
Some scenarios potential scenarios for that could be:
- Adding a pre to InputBlock
- Providing a pre for the model.fit(), which is currently used for `SequenceMaskRandom`

Would `model.get_blocks_by_name()` be a safer way to retrieve the input block?

merlin/models/tf/models/retrieval.py
```diff
@@ -300,3 +309,277 @@ def YoutubeDNNRetrievalModel(
...
+        "The query tower should be an instance of `Encoder` class"
+    )
+    assert isinstance(candidate_tower, Encoder), ValueError(
+        "The query tower should be an instance of `Encoder` class"
+    )
+
+    if not outputs:
+        if not negative_samplers:
+            warnings.warn(
+                "By default, the TwoTower model is trained using constrastive learning."
```
gabrielspmoreira, line 468: I think this should not be a warning. For interested uses the docstrings should be enough.

merlin/models/tf/models/retrieval.py
sararb: Here first is a property of the Encoder class that returns the first block using `self.blocks[0]`. The blocks of the encoder are defined in this [line](https://github.com/NVIDIA-Merlin/models/blob/637c5691820312d6f82d891c742ae3a81a7b6db8/merlin/models/tf/core/encoder.py#L54).  Based on this definition,  I believe `self.blocks[0]` will always be related to the whole input_block (including the `pre` defined inside the input block)

tests/unit/tf/models/test_retrieval.py
```diff
@@ -368,3 +746,147 @@ def last_interaction_as_target(inputs, targets):
...
+    to_remove = (
+        sequence_testing_data.schema.select_by_tag(Tags.SEQUENCE)
+        .select_by_tag(Tags.CONTINUOUS)
+        .column_names
+    )
+    sequence_testing_data.schema = sequence_testing_data.schema.excluding_by_name(to_remove)
+
+    seq_schema = sequence_testing_data.schema.select_by_tag(Tags.SEQUENCE)
+    target = sequence_testing_data.schema.select_by_tag(Tags.ITEM_ID).column_names[0]
+    predict_next = mm.SequencePredictLast(schema=seq_schema, target=target)
```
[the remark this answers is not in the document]
sararb: I added `SequencePredictRandom` to the unit test

sararb:
rerun tests

merlin/models/tf/models/retrieval.py
gabrielspmoreira: ok. Makes sense to me @sararb

marcromeyn (approved):

sararb:
rerun tests

marcromeyn:
rerun tests

PR #794, opened 2022-10-10
Change output format of ToTarget to tuple

edknv:
Follow-up to #740 

### Goals :soccer:
This PR changes the output format of `ToTarget` from `PredictionsOutput` to a tuple `(x, y)`.

Since `ToTarget` was intended as a transformation operation (to be mainly used in `Loader(transform=ToTarget(...))`) that transforms the schema and the input *before* feeding the input to a keras model, the output should be something that keras recognizes (i.e., tuple). When the output format is `PredictionsOutput`, keras throws an error:
```
ValueError: Data is expected to be in format `x`, `(x,)`, `(x, y)`, or `(x, y, sample_weight)`, found: PredictionOutput...
```

### Testing Details :mag:
Two existing unit tests in `core/test_encoder.py` and `models/test_base.py` that were using a custom function to transform inputs without a target were modified to use `ToTarget`.

### DIFF
--- a/merlin/models/tf/transforms/features.py
+++ b/merlin/models/tf/transforms/features.py
@@ -15,7 +15,7 @@
 #
 import warnings
 from itertools import combinations
-from typing import Dict, Optional, Sequence, Union
+from typing import Dict, Optional, Sequence, Tuple, Union
 
 import tensorflow as tf
 from keras.layers.preprocessing import preprocessing_utils as p_utils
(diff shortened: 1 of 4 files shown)


merlin/models/tf/transforms/features.py
```diff
@@ -741,10 +741,7 @@ def call(
         self, inputs: TabularData, targets=None, testing=False, **kwargs
     ) -> "PredictionOutput":
         if testing:
```
gabrielspmoreira, line 743: Are `training` and `testing` flags provided when a transform block is used with the Loader?
Or do we also want to support the usage of `ToTarget()` inside the model?

edknv: Thanks, I now think it's better to drop this parameter entirely because I can't think of an exact scenario where this would be used. Maybe we can add it back if needed.

marcromeyn (approved):

PR #795, opened 2022-10-10
Use pytest-xdist with tensorflow tests to reduce running time of tests

oliverholworthy:
### Goals :soccer:

Reduce running time of tensorflow tests

Using [pytest-xdist](https://github.com/pytest-dev/pytest-xdist) to run tensorflow tests in parallel when more then one process is available.

### Testing Details :mag:

Based on the logs in the unittest there are 2 processes available.
Existing tests completed in ~10mins. 
Which appears to be faster than the current usual completion time of ~15-20mins

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ tests:
 	coverage html --include 'merlin/models/*'
 
 tests-tf:
-	coverage run -m pytest --durations=100 -rsx tests -m "tensorflow and not (integration or example)" || exit 1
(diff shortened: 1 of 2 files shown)


oliverholworthy:
Changed back to draft status. On the second run it appears that maybe the 2 processes may not bring as much benefit as the first suggested (16,16,19 mins) vs first (10, 10, 10 mins). This may indicate that the 2 virtual CPUs that are visible are not completely independent of each other (and of other workflows) running at the same time. 

If we want to be more confident that this approach is helping, perhaps we need some more data-points on the running time. and compare with other other builds over the next few days

marcromeyn (approved):

PR #772, opened 2022-09-27
Introduces OutputBlock and metrics fixes/improvements for multi-task learning

marcromeyn:
### Goals :soccer:
This PR introduces a `OutputBlock` which is the equivalent to the `InputBlock`. It loops over all columns in the schema and based on the tags it adds the right `ModelOutput`. It is a replacement (V2) of the previous block for prediction based `PredictionTask` (V1).

It also contains related tasks:
- [x] Closes #914 
- [x] Created `ColumnBasedSampleWeight` block to set task-specific sample weights base on other features/targets of the dataset

- Refactored metrics creation to:
  - [x] share code for both `metrics` and `weighted_metrics` (`BaseModel.create_metrics()`)
  - [x] accept string metrics (parsed by `metrics_registry`) 
  - [x] standardize metrics when using `OutputBlock` (V2) and `PredictionTask` (v1) when using multi-task learning
  - [x] fix behavior that was adding "_1" to metrics when default task metrics were used (instead of manually defined ones) wiht multi-task learning
  - [x] fix a bug that allowed a mismatch between tasks and metrics when a dict of `PredictionTask` or `ModelOutput` was provided with keys in an order not sorted alphabetically (which is what Keras expects)

- [x] Closes https://github.com/NVIDIA-Merlin/models/issues/942

### Implementation Details :construction:
If there’s only one target column, `OutputBlock` will output a single `ModelOutput`, otherwise multiple outputs will be wrapped in

PR #798, opened 2022-10-11
Tf/dep prediction tasks

marcromeyn:
Fixes # (issue)


gabrielspmoreira:
@marcromeyn I think we could close this draft PR (not without merging it), as it became outdated with the changes from #772 .
What do you think?

PR #811, opened 2022-10-18
Add LogQ correction support to the new ModelOutput API

sararb:
### Context
- We observed an accuracy drop when running the retrieval research script with the new API defined in #790. One of the main issues was that the logic of log-Q correction is defined inside the `call_outputs` method, which is not used in the new ModelOutput API. 

- The serialization of logQ is also not working because we missed `get_config`/`from_config` methods to serialize the required schema object. After adding the methods to the class, a new issue was raised: 

    ==>  If loading the class as a standalone using get_config/from_config  methods, the  tf Variable `candidate_probs`  defined [here](https://github.com/NVIDIA-Merlin/models/blob/8f1a73a07fc4586a374d9d7c793111946726517b/merlin/models/tf/transforms/bias.py#L129) is not loaded:  
             **Note:** This was fixed by adding `candidate_tag_id` and `item_freq_probs` to the config as recommended by @oliverholworthy 

    ==>  Loading a whole model containing logQ block using `tf.keras.load()` is correctly loading the variable `candidate_probs` .
             **Note:** I needed to save `item_freq_probs` as a NumPy array to avoid the serialization error raised by tf.Keras.save_model 

### Goals :soccer:

- [x] Add the logic of log-q correction to the `call` method
- [x] Fix the serialization of LogQ correction

### Testing Details :mag:
- Update tests in `test_bias.py` to ensure logq correction is applied in the new ModelOutputAPI.
- Add a test in `test_contrastive.py` to test ContrastiveOutput with the logq correction as a post block.

### DIFF
--- a/merlin/models/tf/prediction_tasks/classification.py
+++ b/merlin/models/tf/prediction_tasks/classification.py
@@ -226,3 +226,6 @@ class MultiClassClassificationTask(PredictionTask):
 
     def call(self, inputs, training=False, **kwargs):
         return inputs
+
+    def compute_output_shape(self, input_shape):
+        self.pre.compute_output_shape(input_shape)
--- a/merlin/models/tf/prediction_tasks/classification.py
+++ b/merlin/models/tf/prediction_tasks/classification.py
@@ -226,6 +226,3 @@ class MultiClassClassificationTask(PredictionTask):
 
     def call(self, inputs, training=False, **kwargs):
         return inputs
-
-    def compute_output_shape(self, input_shape):
-        self.pre.compute_output_shape(input_shape)
--- a/merlin/models/tf/transforms/bias.py
+++ b/merlin/models/tf/transforms/bias.py
@@ -19,8 +19,10 @@ import tensorflow as tf
 
 from merlin.models.config.schema import requires_schema
 from merlin.models.tf.core.base import Block, PredictionOutput
+from merlin.models.tf.core.prediction import Prediction
 from merlin.models.tf.utils import tf_utils
 from merlin.models.utils import schema_utils
+from merlin.models.utils.schema_utils import schema_to_tensorflow_metadata_json
 from merlin.schema import Schema, Tags
 
 
@@ -100,6 +102,8 @@ class PopularityLogitsCorrection(Block):
     schema: Schema, optional
         The `Schema` with input features,
         by default None
+    candidate_tag_id: Tag, optional
+        The tag of the candidate id
     """
 
     def __init__(
@@ -108,6 +112,7 @@ class PopularityLogitsCorrection(Block):
         is_prob_distribution: bool = False,
         reg_factor: float = 1.0,
         schema: Schema = None,
+        candidate_tag_id=Tags.ITEM_ID,
         **kwargs,
     ):
         super().__init__(**kwargs)
@@ -115,6 +120,7 @@ class PopularityLogitsCorrection(Block):
             self.set_schema(schema)
 
         self.reg_factor = reg_factor
+        self.candidate_id_name = self.schema.select_by_tag(candidate_tag_id).first.name
 
         if item_freq_probs is not None:
             self._check_items_cardinality(item_freq_probs)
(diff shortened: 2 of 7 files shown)


merlin/models/tf/transforms/bias.py
```diff
@@ -206,6 +212,39 @@ def update(
...
+                    tf.tile(tf.squeeze(negative_probs), tf.shape(positive_item_ids)[0:1]),
+                    (-1, tf.shape(negative_item_ids)[0]),
+                )
+                positive_probs = tf.concat(
+                    [tf.expand_dims(positive_probs, -1), negative_probs], axis=1
+                )
+
+            # Applies the logQ correction
+            epsilon = 1e-16
+            predictions = outputs.predictions - (
```
oliverholworthy, line 241: we could move this shared logic into a separate method since it's now being used by both `call` and `call_outputs`? (everything between `postiive_probs = ...` and this line)

then this would become something like:

```python
predictions = self.compute_log_q_correction(predictions, positive_item_ids, negative_item_ids)
```

merlin/models/tf/transforms/bias.py
```diff
@@ -206,6 +212,39 @@ def update(
     def compute_output_shape(self, input_shape):
         return input_shape
 
+    def call(
+        self, outputs: Union[Prediction, PredictionOutput], features, training=False, **kwargs
+    ):
+        if training:
+            if not isinstance(outputs, Prediction):
```
oliverholworthy, line 219: this may be clearer if we combine the instance check to the line above (`if training and isinstance(outputs, Prediction)`)

merlin/models/tf/transforms/bias.py
```diff
@@ -243,3 +282,21 @@ def _check_items_cardinality(self, item_freq_probs):
                 f"(expected {cardinalities[item_id_feature_name]}"
                 f", got {tf.shape(item_freq_probs)[0]})"
             )
+
+    def compute_call_output_shape(self, input_shapes):
+        return self.compute_output_shape(input_shapes)
+
+    def get_config(self):
```
oliverholworthy, line 289: > After adding the methods to the class, a new issue was raised:
> 
> ==>` If loading the class as a standalone using get_config/from_config methods, the tf Variable candidate_probs defined [here](https://github.com/NVIDIA-Merlin/models/blob/8f1a73a07fc4586a374d9d7c793111946726517b/merlin/models/tf/transforms/bias.py#L129) is not loaded.
==> Loading a whole model containing logQ block using tf.keras.load() is correctly loading the variable candidate_probs .

Is the above issue resolved? And if so, what changed here to fix this?

tests/unit/tf/outputs/test_contrastive.py
```diff
@@ -71,6 +71,36 @@ def test_two_tower_constrastive(ecommerce_data: Dataset):
     testing_utils.model_test(model, ecommerce_data)
 
 
+def test_two_tower_constrastive_with_logq_correction(ecommerce_data: Dataset):
+    from merlin.models.tf.transforms.bias import PopularityLogitsCorrection
```
oliverholworthy, line 75: Could these imports be moved to the top of the file? My understanding is that the main reason for an inline import inside a function to avoid a circular dependency issue. Otherwise, I think we're following the style of putting all imports at the top of the file so that the imports can be managed by our pre-commit checks.

merlin/models/tf/prediction_tasks/classification.py
```diff
@@ -226,3 +226,6 @@ def from_schema(
 
     def call(self, inputs, training=False, **kwargs):
         return inputs
+
+    def compute_output_shape(self, input_shape):
+        self.pre.compute_output_shape(input_shape)
```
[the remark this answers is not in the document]
sararb: This was related to the unit test `test_two_tower_advanced_options`. Adding a call method the logQ block raises an issue of `missing compute_output_shape` in the classification task when training a retrieval model (two-tower or MF) with an ItemRetrievalTask and LogQ correction. 

Thank you for the suggestion, I moved the implementation of the compute output shape to the `PredictionTask` and made sure we check if `task_block` or `pre` are specified to call their `compute_output_shape` methods, otherwise we return the original `input_shape`.

tests/unit/tf/outputs/test_contrastive.py
sararb: Thank you for clarifying that! I always assumed that if an import is used in one single unit test, we should call it inside the function :)

oliverholworthy (approved):

merlin/models/tf/transforms/bias.py
oliverholworthy: I think maybe the issue is that we also need to save some more of the params in this `get_config` - the `item_freq_probs` and `candidate_tag_id`

PR #812, opened 2022-10-18
Add L2-batch regularization to EmbeddingTable

sararb:
### Goals :soccer:
We observed an accuracy drop when running the retrieval research script with the new API defined in https://github.com/NVIDIA-Merlin/models/pull/790. This figure shows that the regularization loss of the new API was higher during the first iterations, compared to the regularization loss of the old API. Which prevents the model from learning useful information.  

image

After some investigation, it seems the issue was related to the embedding table regularization.  In fact, in the old API, the l2 regularization was computed only for the embeddings in the current batch using this custom [code](https://github.com/NVIDIA-Merlin/models/blob/093cf2bc2942e54b8e56b1709af0ea17ca476676/merlin/models/tf/inputs/embedding.py#L852). While  the new Embeddings API uses base Keras regularizers ([here](https://github.com/NVIDIA-Merlin/models/blob/093cf2bc2942e54b8e56b1709af0ea17ca476676/merlin/models/tf/inputs/embedding.py#L168)) that computes the regularization loss using the whole embeddings table weights. 

- Add the support of embeddings l2 regularization using only the current batch weights. 

### Testing Details :mag:
- Add a unit test to `embeddings.py` that compares the three scenarios: no regularization, batch regularization and embedding table regularization.

### DIFF
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -198,6 +198,9 @@ class EmbeddingTable(EmbeddingTableBase):
        for example, or generally for any layer that manipulates tensors
        using Python control flow. If `False`, we assume that the layer can
        safely be used to generate a static computation graph.
+    l2_batch_regularization_factor: float, optional
+        Factor for L2 regularization of the embeddings vectors (from the current batch only)
+        by default 0.0
     **kwargs: Forwarded Keras Layer parameters
     """
 
@@ -217,6 +220,7 @@ class EmbeddingTable(EmbeddingTableBase):
         dtype=None,
         dynamic=False,
         table=None,
+        l2_batch_regularization_factor=0.0,
         **kwargs,
     ):
         """Create an EmbeddingTable."""
@@ -249,6 +253,7 @@ class EmbeddingTable(EmbeddingTableBase):
             )
         self.sequence_combiner = sequence_combiner
         self.supports_masking = True
+        self.l2_batch_regularization_factor = l2_batch_regularization_factor
 
     def select_by_tag(self, tags: Union[Tags, Sequence[Tags]]) -> Optional["EmbeddingTable"]:
         """Select features in EmbeddingTable by tags.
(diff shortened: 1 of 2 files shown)


merlin/models/tf/inputs/embedding.py
```diff
@@ -380,6 +385,14 @@ def call(
         else:
             out = self._call_table(inputs, **kwargs)
 
+        if self.l2_batch_regularization_factor > 0:
+            if isinstance(out, dict):
```
oliverholworthy, line 389: In this case where we have a dict with potentially more than one feature. Do we want to call `add_loss` more than once? Currently it's going to apply this only to the last feature.

If we want to call it for each feature, then either need to loop through the feature names again here or could put this code inside the `_call_table` method?

merlin/models/tf/inputs/embedding.py
```diff
@@ -483,6 +496,7 @@ def Embeddings(
     post: Optional[BlockType] = None,
     aggregation: Optional[TabularAggregationType] = None,
     block_name: str = "embeddings",
+    l2_batch_regularization_factor: float = 0.0,
```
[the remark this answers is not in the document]
oliverholworthy: And if we made that change, would also need to update the part in the body of this function to match the other kwargs:

```python
if l2_batch_regularization_factor:
    kwargs["l2_batch_regularization_factor"] = l2_batch_regularization_factor
```

merlin/models/tf/inputs/embedding.py
```diff
@@ -524,7 +533,7 @@ def Embeddings(
         Transformation block to apply after the embeddings lookup, by default None
     aggregation: Optional[TabularAggregationType], optional
         Transformation block to apply for aggregating the inputs, by default None
-    block_name: str, optional
+    block_name: Optional[float, Dict[str, float]] = 0.0
```
oliverholworthy, line 536: missing the name and description of the new parameter here `l2_batch_regularization_factor`.

oliverholworthy (approved):

PR #817, opened 2022-10-21
Fix `EmbeddingFeatures` serialization/deserialization when a feature is named 'name'

oliverholworthy:
### Goals :soccer:

- Fix bug in `EmbeddingFeatures` serialization when a features with the name `name` is included.
  - When a feature with the name `"name"` is included alongside others. Those that get processed after will be assigned to share the same embedding table. 
    - In the case where the embedding sizes of these features are the same, this wouldn't result in an exception, but would silently produce different results when a model is re-loaded. 
    - If the embeddings were of different sizes, then an exception would be raised along the lines of `Input layer "dense_X" is incompatible: expected axis -1 of input shape to have value 212, but received input with shape (2, 158)`

  - If a feature with `name` is included, then when the class is deserialized, it will ignore and features that are loaded after `name`. Which means the model re-load will not result in the same embedding tables. Resulting in errors when trying to run inference on a reloaded model.
  - Updates and simplifies `from_config` method of `EmbeddingFeatures` to remove the line that caused this error
### Testing Details :mag:

- added a  test for `EmbeddingFeatures` serialization using `get_config`/`from_config`

### DIFF
--- a/merlin/models/tf/inputs/embedding.py
+++ b/merlin/models/tf/inputs/embedding.py
@@ -1030,24 +1030,19 @@ class EmbeddingFeatures(TabularBlock):
             feature_configs[key] = feature_config_dict
 
         config["feature_config"] = feature_configs
+        config["l2_reg"] = self.l2_reg
 
         return config
 
     @classmethod
     def from_config(cls, config):
         # Deserialize feature_config
-        feature_configs, table_configs = {}, {}
+        feature_configs = {}
         for key, val in config["feature_config"].items():
-            feature_params = deepcopy(val)
-            table_params = feature_params["table"]
-            if "name" in table_configs:
-                feature_params["table"] = table_configs["name"]
-            else:
-                table = deserialize_table_config(table_params)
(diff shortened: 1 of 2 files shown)


merlin/models/tf/inputs/embedding.py
```diff
@@ -1030,24 +1030,19 @@ def get_config(self):
...
 
     @classmethod
     def from_config(cls, config):
         # Deserialize feature_config
-        feature_configs, table_configs = {}, {}
+        feature_configs = {}
         for key, val in config["feature_config"].items():
-            feature_params = deepcopy(val)
-            table_params = feature_params["table"]
-            if "name" in table_configs:
```
oliverholworthy, line 1043: This is the line that was causing issues. Once a features called `"name"` is processed. It will skip all remaining features and assign them to share the name feature. Resulting in the embeddings for the rest of the features having the same embedding as the `name` feature.

marcromeyn (approved):

PR #820, opened 2022-10-24
Reduce running time of unit tests

oliverholworthy:
### Goals :soccer:

- Improve iteration speed by reducing the running time of unit tests


oliverholworthy:
Will re-visit this in future work targetted the slowest tests first. The approch in this PR of reducing the cardinality of some of the example data requires more careful work, since some of the unit tests depend on these values.

PR #825, opened 2022-10-26
Add horovod for distributed training

edknv:
Fixes https://github.com/NVIDIA-Merlin/Merlin/issues/651

### Goals :soccer:
- Use horovod to enable data-parallel distributed training

Horovod documentation lists the following steps for enabling horovod ([link](https://horovod.readthedocs.io/en/stable/keras.html)):
1. Run `hvd.init()`.
2. Pin each GPU to a single process.
3. Scale the learning rate by the number of workers.
4. Wrap the optimizer in `hvd.DistributedOptimizer`.
5. Add `hvd.callbacks.BroadcastGlobalVariablesCallback(0)` to broadcast initial variable states from rank 0 to all other processes.

The high-level goal/design of this PR is to perform the above steps for the user automatically behind the Models API.
1. `hvd.init()` is run in [merlin/models/tf/distributed/backend.py](https://github.com/NVIDIA-Merlin/models/pull/825/files#diff-b2a05b848647531e0814a2e4474db3f75d89f27578686eac4138dcc0817f80fb).
2. We pin GPU by using [examples/usecases/multi-gpu/hvd_wrapper.sh](https://github.com/NVIDIA-Merlin/models/pull/825/files#diff-889c765e52b23d82e96237cb779aef318161b39a86888ee421c8797048a166f5). (Note: Horovod examples recommend using `tf.config.experimental.set_visible_devices()` but this sometimes doesn't work, e.g., on NGC.)
3. Learning rates are scaled automatically both for Keras optimizers and `MultiOptimizer`.
4. If users use a Keras optimizer or MultiOptimizer, they are automatically converted to `hvd.DistributedOptimizer` when hvd is installed there are more than one GPUs.
5. `hvd.callbacks.BroadcastGlobalVariablesCallback(0)` is automatically added to callbacks if hvd is installed and there are more than one GPUs.

From the user's perspective, the training script itself looks identical whether the script is for a single GPU or multi-GPU. There is likely no need for the user to import the upstream horovod library at all. When they run the script normally with python, e.g.,
`python -m pytest tests/unit/tf -m horovod`, it will run in a single process even if they have multiple GPUs. If the user wants to use multiple GPUs, the script has to be executed with `horovodrun`: `horovodrun -np 2 sh hvd_wrapper.sh python -m pytest tests/unit/tf -m horovod`.

### Sanity checks

- [ ] Does the throughput scale as we go from 1 GPU to 2 GPUs, 4 GPUs, etc.
- [ ] Does the performance stay the same (or increase) as we go from 1 GPU to 2 GPUs, 4 GPUs, etc.

### Future work

- Can we get rid of `hvd_wrapper.sh` by absorbing it either into `horovodrun` or Models?
- Can we use [ray](https://github.com/horovod/horovod/blob/master/examples/ray/tensorflow2_mnist_ray.py) to make the API cleaner?

### Testing Details :mag:
Unit tests on CPU (Github Actions, `python`) and on multiple GPUs (Jenkins, `horovodrun`).
Sanity checks listed above on NGC.


tests/unit/tf/horovod/test_horovod.py
```diff
@@ -0,0 +1,79 @@
...
+
+import merlin.models.tf as mm
+from merlin.datasets.advertising.criteo.dataset import default_criteo_transform
+from merlin.io.dataset import Dataset
+from merlin.models.tf.distributed.backend import hvd
+from merlin.models.utils.example_utils import workflow_fit_transform
+from merlin.schema.tags import Tags
+
+
+def test_import():
```
jperez999, line 14: Why is this test written this way? Doesnt the init file pytest.importorskip cover this? In what case does this assert fail?

tests/unit/tf/horovod/test_horovod.py
```diff
@@ -0,0 +1,79 @@
...
+    from merlin.models.tf.distributed.backend import hvd
+
+    horovod_found = importlib.util.find_spec("horovod")
+
+    assert (horovod_found and hvd is not None) or (not horovod_found and hvd is None)
+
+
+def test_horovod_multigpu(criteo_data, tmpdir, batch_size=11, learning_rate=0.03):
```
jperez999, line 22: This test is not able to run inside the context of a regular testrun... is there no way to reference the test using a subprocess call and adding a parameter that allows you to ignore the subprocess command if you are already in that context? Or maybe put this subprocess command in the init file... and keep this as a function without the test_ in the front of the name so it is ignored by pytest.

merlin/models/tf/loader.py
```diff
@@ -618,3 +623,42 @@ def sample_batch(
...
+def get_default_hvd_seed_fn(seed=None):
+    """
+    Generate consistent dataloader shuffle seeds across workers
+    Reseeds each worker's dataloader each epoch to get fresh a shuffle
+    that's consistent across workers.
+    """
+    if HAS_GPU:
+        try:
+            import cupy
+        except ImportError:
```
jperez999, line 637: In what case is it possible to have a gpu but not cupy? And is the import error message correct? I see that you only need cupy if the GPU is available... Why cant we default to numpy in the case where GPU is available but not cupy?

tests/unit/tf/horovod/test_horovod.py
jperez999: This may not work... but have we tried to run it in the same context as other tests... we run other subprocesses in many other tests (not mpi based but still for example notebooks). If it was tried and didnt work, all good, just want to make sure we tried.

tests/conftest.py
```diff
@@ -106,3 +106,5 @@ def pytest_collection_modifyitems(items):
             item.add_marker(pytest.mark.datasets)
         if "/transformers/" in path:
             item.add_marker(pytest.mark.transformers)
+        if "/horovod/" in path:
+            item.add_marker(pytest.mark.horovod)
```
jperez999, line 110: What does this pytest marker do? (just curious)

ci/test_unit.sh
```diff
@@ -17,4 +17,7 @@
 #!/bin/bash
 set -e
 
-TF_GPU_ALLOCATOR=cuda_malloc_async python -m pytest -rxs tests/unit
\ No newline at end of file
+TF_GPU_ALLOCATOR=cuda_malloc_async python -m pytest -rxs tests/unit
+
+# Only run tests marked with `horovod` with `horovodrun` command
+horovodrun -np 2 sh ./examples/usecases/multi-gpu/hvd_wrapper.sh python -m pytest -rxs tests/unit -m horovod
```
jperez999, line 23: If my comment below is possible, would it be better to run this inside the context of the regular test/unit call in line 20?

edknv:
rerun tests

examples/usecases/multi-gpu/hvd_wrapper.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
marcromeyn, line 1: Could we have our own version of `horovodrun` that includes this code? That would simplify our examples as well

tests/unit/tf/horovod/test_horovod.py
edknv: It was a silly test carried over from template code. Removed.

tests/unit/tf/horovod/test_horovod.py
edknv: I tried running it with `subprocess` and wasn't able to find a way to make it work in subprocess. When tests are executed with `horovod -np 2 hvd_wrapper.sh python -m pytest tests/unit`, the tests pass and I can see from logs it's using both GPUs. These tests should work both on CPU (by _not_ using `horovodrun`) and on GPU (by using `horovodrun`).

merlin/models/tf/loader.py
edknv: Good point. Updated.

tests/conftest.py
edknv: It marks all the tests inside `/horovod` directory with the `horovod` marker so we can selectively run only the tests inside this directory with `horovodrun ... pytest -m horovod`.

ci/test_unit.sh
edknv: Tried the `subprocess` approach but was unsuccessful as mentioned in my response above.

examples/usecases/multi-gpu/hvd_wrapper.sh
edknv: I agree, that wold be great, but I couldn't a way to do this. Setting of these environment variables has to come after `horovodrun` and seems to depend heavily on the order of imports so it has come before everything else (likely tensorflow). I wasn't able to find a solution in an afternoon, and I'm not sure if it's worth digging more. I think we will have to pretty much solve the import order problem, which seems difficult to analyze and debug.

marcromeyn: Makes sense. In that case we can leave that task for something to work on after this PR.

tests/unit/tf/horovod/test_horovod.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    # As a workaround for nvtabular producing different numbers of batches in
+    # workers, we repartition the dataset in multiples of hvd.size().
+    # https://github.com/NVIDIA-Merlin/models/issues/765
+    train = Dataset(os.path.join(tmpdir, "processed", "train", "*.parquet"))
+    ddf = train.to_ddf().repartition(npartitions=2)
+    train = Dataset(ddf, schema=train.schema)
+
+    train_loader = mm.Loader(
```
bschifferer, line 46: Is there a way that we do not initialize a loader?

bschifferer:
@marcromeyn  @edknv  I am started to develop an example and have follow up questions:

1. We require to initialize a separate dataloader - we do not require it in any other example. Is it required we will do it in this case?
2. model.fit runs each worker on the full dataset. Therefore, each epoch is 2x larger than before.

```
train = Dataset(ddf, schema=train.schema)

    train_loader = mm.Loader(
        train,
        schema=train.schema,
        batch_size=batch_size,
        shuffle=True,
        drop_last=True,
    )

    target_column = train.schema.select_by_tag(Tags.TARGET).column_names[0]

    model = mm.DLRMModel(
        train.schema,
        embedding_dim=16,
        bottom_block=mm.MLPBlock([32, 16]),
        top_block=mm.MLPBlock([32, 16]),
        prediction_tasks=mm.BinaryClassificationTask(target_column),
    )

    opt = tf.keras.optimizers.Adagrad(learning_rate=learning_rate)
    model.compile(optimizer=opt, run_eagerly=False, metrics=[tf.keras.metrics.AUC()])

    losses = model.fit(
        train_loader,
        batch_size=batch_size,
    )
```

edknv:
> 1. We require to initialize a separate dataloader - we do not require it in any other example. Is it required we will do it in this case?

@bschifferer Dataloader is used here to drop the last batch, i.e., `drop_last=True`, so we can drop the last batch if it's not full. It's not strictly required but I think in a distributed setting it makes sense to drop the last batch which might be small.

> 2. model.fit runs each worker on the full dataset. Therefore, each epoch is Nx larger than single GPU (where N is number of GPUs).

When I test this, each worker seems to have 1/Nth steps and the epochs are shorter. On a single GPU:
```
Epoch 1/2
2136/2136 [==============================] - 44s 18ms/step - loss: 0.6853 - auc: 0.5708 - regularization_loss: 0.0000e+00
Epoch 2/2
2136/2136 [==============================] - 39s 18ms/step - loss: 0.6846 - auc: 0.5738 - regularization_loss: 0.0000e+00
```
On 2 GPUs using `horovodrun -np 2`:
```
[1,0]<stdout>:Epoch 1/2
1068/1068 [==============================] - 31s 21ms/step - loss: 0.6850 - auc: 0.5719 - regularization_loss: 0.0000e+00stdout>::
[1,0]<stdout>:Epoch 2/2
1068/1068 [==============================] - 23s 21ms/step - loss: 0.6845 - auc: 0.5739 - regularization_loss: 0.0000e+00stdout>::>:
```
I added an assertion for testing the step sizes in the unit tests in [6013913](https://github.com/NVIDIA-Merlin/models/pull/825/commits/601391386295c15777656f23057aaf5603671f4c).

tox.ini
```diff
@@ -24,7 +23,11 @@ deps =
...
 setenv =
     TF_GPU_ALLOCATOR=cuda_malloc_async
+    HOROVOD_GPU_OPERATIONS=NCCL
+allowlist_externals =
+    horovodrun
 commands =
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/nvtabular.git
     python -m pytest --cov-report term --cov merlin -rxs tests/unit
+    horovodrun -np 2 sh examples/usecases/multi-gpu/hvd_wrapper.sh python -m pytest -m horovod -rxs tests/unit
```
marcromeyn, line 33: Would it make sense to add this command to the `Makefile`?

tests/unit/tf/horovod/test_horovod.py
```diff
@@ -0,0 +1,160 @@
...
```
jperez999, line 54: Does this loader need to have a global_size and global_rank supplied? (they can both be one I think...) we would want to test the multi-gpu part right? Same for next test.

tox.ini
edknv: Github actions uses the tests in the `Makefile` right? I was focusing on the GPU case so didn't think about adding it to CPU tests. I think we need to add a new workflow to the test matrix because of the horovod dependency so maybe I should follow up with another PR that runs tests on github actions.

tests/unit/tf/horovod/test_horovod.py
edknv: We set it [here](https://github.com/edknv/models/blob/6f9561059fa7bd060739641ab31baed21d3cd836/merlin/models/tf/loader.py#L299-L300) in the loader unless the user overrides it:
```python
            global_size = global_size or hvd.size()
            global_rank = global_rank or hvd.rank()
```
This PR is currently running tests on the Jenkins instance for the multi-part. I added a test run in [tox.ini](https://github.com/edknv/models/blob/6f9561059fa7bd060739641ab31baed21d3cd836/tox.ini#L33):
```
    horovodrun -np 2 sh examples/usecases/multi-gpu/hvd_wrapper.sh python -m pytest -m horovod -rxs tests/unit
```

edknv:
One test on Jenkins which was added very recently in https://github.com/NVIDIA-Merlin/models/pull/841 is failing. It seems like introducing horovod in the dependency caused dependency conflict with dask, causing an error somewhere in nvtabular + dask, unrelated to horovod. I see:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
feast 0.19.4 requires dask<2022.02.0,>=2021.*, but you have dask 2022.5.1 which is incompatible.
```
which I don't see in other CI logs. Need to resolve the dependency conflict somehow.

jperez999:
> One test on Jenkins which was added very recently in #841 is failing. It seems like introducing horovod in the dependency or something else caused a dependency conflict with dask, causing an error somewhere in nvtabular + dask, unrelated to horovod. I see:
> 
> ```
> ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
> feast 0.19.4 requires dask<2022.02.0,>=2021.*, but you have dask 2022.5.1 which is incompatible.
> ```
> 
> which I don't see in other CI logs. Need to resolve the dependency conflict somehow, but I can't seem to reproduce the error locally.

This not a problem as far as we have seen. Its just that the older version of feast does not know that newer versions of dask exist. It has yet to cause problems in our merlin end 2 end level testing. However it is on our radar to update that version of feast.

tox.ini
marcromeyn: I was more thinking that it might be useful to be able to trigger the mutli-gpu test locally (using for instance the Makefile) But we can leave that for a future PR as well.

marcromeyn (approved):

Issue #824, opened 2022-10-25
[BUG] transformer_post="pooler_output" does not work for GPT2Block and XLNetBlock


rnyak:
### Bug description
I get the following error when I use `transformer_post="pooler_output"` with GPT2Block or XLNet Block.
```
/usr/local/lib/python3.8/dist-packages/keras/initializers/initializers_v2.py:120: UserWarning: The initializer TruncatedNormal is unseeded and being called multiple times, which will return identical values  each time (even if the initializer is unseeded). Please update your code to provide a seed to the initializer, or avoid using the same initalizer instance more than once.
  warnings.warn(
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In [7], line 2
      1 model.compile(run_eagerly=False, optimizer='adam')
----> 2 model.fit(loader, epochs=1)

File /usr/local/lib/python3.8/dist-packages/merlin/models/tf/models/base.py:831, in BaseModel.fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq, max_queue_size, workers, use_multiprocessing, train_metrics_steps, pre, **kwargs)
    828     self._reset_compile_cache()
    829     self.train_pre = pre
--> 831 out = super().fit(**fit_kwargs)
    833 if pre:
    834     del self.train_pre

File /usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py:70, in filter_traceback.<locals>.error_handler(*args, **kwargs)
     67     filtered_tb = _process_traceback_frames(e.__traceback__)
     68     # To get the full stack trace, call:
     69     # `tf.debugging.disable_traceback_filtering()`
---> 70     raise e.with_traceback(filtered_tb) from None
     71 finally:
     72     del filtered_tb

File /usr/local/lib/python3.8/dist-packages/merlin/models/tf/models/base.py:1006, in Model._maybe_build(self, inputs)
   1004             child._feature_shapes = feature_shapes
   1005             child._feature_dtypes = feature_dtypes
-> 1006 super()._maybe_build(inputs)

File /usr/local/lib/python3.8/dist-packages/merlin/models/tf/models/base.py:1024, in Model.build(self, input_shape)
   1022 for layer in self.blocks:
   1023     try:
-> 1024         layer.build(input_shape)
   1025     except TypeError:
[...]

PR #828, opened 2022-10-26
Adding SequenceSummary transforms for transformers

marcromeyn:
Fixes: 
- #824

### Goals :soccer:
In T4Rec we have a `summary_type` that can be passed to transformer-blocks. This PR adds the same transforms but integrates it into the `post`. This allows users to do things like: `BertBlock(..., post="sequence_mean")`.

### Testing Details :mag:
Added a test to ensure correct behaviour.

### DIFF
--- a/merlin/models/tf/transformers/transforms.py
+++ b/merlin/models/tf/transformers/transforms.py
@@ -13,11 +13,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
+from types import SimpleNamespace
 from typing import Dict
 
 import tensorflow as tf
 from tensorflow.keras.layers import Layer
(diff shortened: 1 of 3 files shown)


rnyak:
@marcromeyn can we add a test in [here](https://github.com/NVIDIA-Merlin/models/blob/eee1b6b71a7292df664305c70012443cd2212eea/tests/unit/tf/transformers/test_block.py#L90-L113) to test that GPT2Block and XLNetBlock works with different transformation like `post = sequence_mean` that's the critical one.

marcromeyn:
I don't think there's a need for that test. We know that the `last_hidden_state` is a 3-D tensor (HF has tests to ensure that) & we have a test to make sure all sequence-aggregations work from a 3D tensor.

marcromeyn:
rerun tests

PR #830, opened 2022-10-26
Ensure metrics passed to `model.compile()` are always reset

### LINKED ISSUE #144: [BUG] Inconsistent eval metric results
I am training a DLRM model with my custom dataset. I get the following metric print outs during training:
```
To silence this warning, decorate the function with @tf.autograph.experimental.do_not_convert
1315/1315 [==============================] - 60s 40ms/step - y/binary_classification_task/precision: 0.0108 - y/binary_classification_task/recall: 0.0038 - y/binary_classification_task/binary_accuracy: 0.9879 - y/binary_classification_task/auc: 0.6896 - loss: 0.0575 - regularization_loss: 0.0000e+00 - total_loss: 0.0575
********************
Start evaluation
329/329 [==============================] - 15s 38ms/step - y/binary_classification_task/precision: 0.0000e+00 - y/binary_classification_task/recall: 0.0000e+00 - y/binary_classification_task/binary_accuracy: 0.9910 - y/binary_classification_task/auc: 0.7320 - loss: 0.0480 - regularization_loss: 0.0000e+00 - total_loss: 0.0480
********************

```

The AUC metric printed out for validation dataset is `0.7320` but the precision and recall are still ` 0.0000e+00`. 

I build the DLRM model as follows

```
train['ypred'] = 0
lr =0.0005

for fold in range(5):
    ind_train = train.fold != fold
    ind_valid = train.fold == fold
    
    TRAIN = train.loc[ind_train].reset_index(drop=True)

    VALID = train.loc[ind_valid].reset_index(drop=True)

    print(TRAIN.shape, VALID.shape)    

    dlrm_body = ml.DLRMBlock(schema,
        embedding_dim=16,
        bottom_block=ml.MLPBlock([64, 16]),
        top_block=ml.MLPBlock([64, 32]),
    )
    model = dlrm_body.connect(ml.BinaryClassificationTask("y"))
    optimizer = tf.keras.optimizers.Adam(lr)
    model.compile(optimizer=optimizer, run_eagerly=False)
    
    train_loader = get_dataloader(nvt.Dataset(TRAIN)) 
    losses = model.fit(train_loader, epochs=1)
    model.save("./tmp/my_model")
    model.reset_metrics()

    # Evaluate
    print('*'*20)
    print("Start evaluation")
    eval_loader = get_dataloader(nvt.Dataset(VALID)) 
    eval_metrics = model.evaluate(eval_loader, return_dict=True)

    train.loc[ind_valid, 'ypred'] = model.predict(eval_loader) 
    print("AUC SCORE:")
    print(roc_auc_score(train.loc[ind_valid, 'y'], train.loc[ind_valid, 'ypred']) )
```

sararb:
Fixes #144 

### Goals :soccer:

- Debugging the source of issue #144, I noticed that Keras `.evaluate()` method does not reset metrics defined in the `compile()` methods. 

- While this is ok for almost all use cases where metrics are set for the first time in the `compile`, this is an issue in the case of top-k evaluation where the metrics passed to the `recommender.compile()` were already used in the training of the retrieval model and already contains training results. 

- In fact, our current logic of top-k evaluation is : 
         - Train a retrieval model with top-k metrics.
         - Convert the retrieval model to a top-k recommender.
         - Compile the top-k recommender with the same metrics and run .evaluate() to score over the whole catalog instead of in-batch items. 

- The source of the issue related to Keras not resetting the metrics at the first call of `fit` or `evaluate` that happens after the `compile()` is related to: 
    - When model.compile is called : The [metrics are wrapped](https://github.com/keras-team/keras/blob/v2.10.0/keras/engine/training.py#L729-L734) in a [MetricsContainer](https://github.com/keras-team/keras/blob/v2.10.0/keras/engine/compile_utils.py#L369) saved as self.compiled_metrics
    - The [model.metrics property](https://github.com/keras-team/keras/blob/v2.10.0/keras/engine/training.py#L792-L840) extracts the metrics from the MetricsContainer from `self.compiled_metrics.metrics` 
    - Reset_metrics[ is called prior to evaluation on the dataset](https://github.com/keras-team/keras/blob/v2.10.0/keras/engine/training.py#L1940) which loops through the values returned by self.metrics and calls reset_state
    - However, the first time model.evaluate is called, self.metrics returns [] because the [MetricsContainer doesn't return the metrics since it's not built](https://github.com/keras-team/keras/blob/v2.10.0/keras/engine/compile_utils.py#L438-L443) (according to `MetricsContainer._built` set by default to False).
   - This is why the metrics are not reset the first call to model.evaluate()
   - The [build](https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/engine/compile_utils.py#L508) method of  `MetricsContainer` is responsible for building the metrics and set the flag `._built` to True. 
  - So maybe, building the metrics before calling `reset_states` in .evaluate() could fix this issue.

- In the meantime of creating a reproducible code and opening a feature request in Keras for our specific use case, I added a logic inside the base model class of merlin to ensure metrics passed to compile are always reset. 

- Add logic of reset metrics states in `_create_metrics` and `_create_weighted_metrics`
- We need to check that the metric isn't `None` and is `built` ([default value](https://github.com/keras-team/keras/blob/b80dd12da9c0bc3f569eca3455e77762cf2ee8ef/keras/metrics/base_metric.py#L120) of keras metrics) before reset_states

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -375,8 +375,8 @@ class BaseModel(tf.keras.Model):
             out = metrics
 
         for metric in tf.nest.flatten(out):
-            # We need to ensure metrics passed to `compile()` are reset
-            if metric and metric.built:
+            # We ensure metrics passed to `compile()` are reset
+            if metric:
                 metric.reset_state()
         return out
 
@@ -403,8 +403,7 @@ class BaseModel(tf.keras.Model):
                         out[block.full_name] = weighted_metrics[i]
 
         for metric in tf.nest.flatten(out):
-            # We need to ensure metrics passed to `compile()` are reset
-            if metric and metric.built:
+            if metric:
                 metric.reset_state()
 
         return out
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -374,6 +374,10 @@ class BaseModel(tf.keras.Model):
         else:
             out = metrics
 
+        for metric in tf.nest.flatten(out):
+            # We need to ensure metrics passed to `compile()` are reset
+            if metric and metric.built:
+                metric.reset_states()
         return out
 
     def _create_weighted_metrics(self, weighted_metrics=None):
@@ -397,6 +401,11 @@ class BaseModel(tf.keras.Model):
                 else:
                     for i, block in enumerate(self.model_outputs):
                         out[block.full_name] = weighted_metrics[i]
+
+        for metric in tf.nest.flatten(out):
+            # We need to ensure metrics passed to `compile()` are reset
+            if metric and metric.built:
+                metric.reset_states()
 
         return out
 
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -377,7 +377,7 @@ class BaseModel(tf.keras.Model):
         for metric in tf.nest.flatten(out):
             # We need to ensure metrics passed to `compile()` are reset
             if metric and metric.built:
-                metric.reset_states()
+                metric.reset_state()
         return out
 
     def _create_weighted_metrics(self, weighted_metrics=None):
@@ -405,7 +405,7 @@ class BaseModel(tf.keras.Model):
         for metric in tf.nest.flatten(out):
             # We need to ensure metrics passed to `compile()` are reset
             if metric and metric.built:
-                metric.reset_states()
+                metric.reset_state()
 
         return out
 
--- a/tests/unit/tf/models/test_base.py
+++ b/tests/unit/tf/models/test_base.py
@@ -140,7 +140,8 @@ class UpdateCountMetric(tf.keras.metrics.Metric):
         return self.call_count[0]
 
     def reset_state(self):
-        self.call_count.assign(tf.constant([0.0]))
+        if self.built:
+            self.call_count.assign(tf.constant([0.0]))
 
 
 @pytest.mark.parametrize(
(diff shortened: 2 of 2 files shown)


merlin/models/tf/models/base.py
```diff
@@ -398,6 +402,11 @@ def _create_weighted_metrics(self, weighted_metrics=None):
                     for i, block in enumerate(self.model_outputs):
                         out[block.full_name] = weighted_metrics[i]
 
+        for metric in tf.nest.flatten(out):
+            # We need to ensure metrics passed to `compile()` are reset
+            if metric and metric.built:
```
sararb, line 407: We need to check if metric is built to ensure related states variables were created (like in this [unit-test](class UpdateCountMetric(tf.keras.metrics.Metric):))

merlin/models/tf/models/base.py
```diff
@@ -374,6 +374,10 @@ def _create_metrics(self, metrics=None):
         else:
             out = metrics
 
+        for metric in tf.nest.flatten(out):
+            # We need to ensure metrics passed to `compile()` are reset
+            if metric and metric.built:
+                metric.reset_states()
```
oliverholworthy, line 380: Based on [this code](https://github.com/keras-team/keras/blob/v2.10.0/keras/metrics/base_metric.py#L419-L424), looks like `reset_states` has been deprecated. Replaced by `reset_state`. This works, since it's an alias for backwards compatibility, but maybe for consistency with the latest API, can be renamed here.

merlin/models/tf/models/base.py
oliverholworthy: Maybe we could remove this check for `metric.built` since in the keras code reset_metrics doesn't check this. And instead update the `UpdateCountMetric` `reset_state` to handle the absence of the `call_count` attribute.

oliverholworthy (approved):

PR #833, opened 2022-10-27
Add test for retrieval model with transformer block

oliverholworthy:
Fixes # (issue)

### Goals :soccer:

- Add unit test for retreival model with transformer encoder

### DIFF
--- a/merlin/models/tf/outputs/contrastive.py
+++ b/merlin/models/tf/outputs/contrastive.py
@@ -314,9 +314,7 @@ class ContrastiveOutput(ModelOutput):
         return negatives
 
     def embedding_lookup(self, ids: tf.Tensor):
-        if ids.shape.rank == 2:
-            ids = tf.squeeze(ids, axis=1)
-        return self.to_call.embedding_lookup(ids)
+        return self.to_call.embedding_lookup(tf.squeeze(ids))
 
     def to_dataset(self, gpu=None) -> merlin.io.Dataset:
         return merlin.io.Dataset(tf_utils.tensor_to_df(self.to_call.embeddings, gpu=gpu))
--- a/merlin/models/tf/outputs/contrastive.py
+++ b/merlin/models/tf/outputs/contrastive.py
@@ -197,9 +197,6 @@ class ContrastiveOutput(ModelOutput):
             positive_id = features[self.col_schema.name]
             positive_embedding = inputs[self.candidate_name]
 
-        if isinstance(positive_id, tf.RaggedTensor):
-            positive_id = positive_id.to_tensor()
-
         positive = Candidate(positive_id, {**features}).with_embedding(positive_embedding)
         negative = self.sample_negatives(positive, features, training=training, testing=testing)
         if self.has_candidate_weights and (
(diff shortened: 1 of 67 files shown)


oliverholworthy:
rerun tests

oliverholworthy:
rerun tests

merlin/models/tf/outputs/contrastive.py
```diff
@@ -314,6 +317,8 @@ def sample_negatives(
         return negatives
 
     def embedding_lookup(self, ids: tf.Tensor):
+        if isinstance(ids, tf.RaggedTensor):
+            ids = ids.to_tensor()
```
marcromeyn, line 321: The `Embedding` layer from Keras support ragged-tensors, right? What's the reasoning behind always converting to dense?

merlin/models/tf/transforms/sequence.py
```diff
@@ -219,7 +219,7 @@ def call(
         self._check_seq_inputs_targets(inputs)
 
         # Shifts the target column to be the next item of corresponding input column
-        new_target = inputs[self.target_name][:, 1:]
+        new_target = inputs[self.target_name][:, -1:]
```
sararb, line 222: The tranform `SequencePredictNext` aims to perform a sliding window prediction of size 1 (we referred to it as Causal Language modeling in T4Rec to align with the NLP domain). More specifically, the session-based model will use the hidden representation at position N to predict the target at position N+1. So each input row in the batch will be linked to multiple targets (from position 2 up to last non-padded position)

If you want to predict only the last item of the sequence you should use `SequencePredictLast` indead.

oliverholworthy: Thanks for the explanation. I think I got confused here between predict next and predict last.

tests/unit/tf/transformers/test_block.py
```diff
@@ -25,6 +25,59 @@ def test_import():
...
+    output_schema = seq_schema.select_by_name(target)
+
+    query_encoder = mm.Encoder(
+        mm.InputBlockV2(
+            query_schema,
+            embeddings=mm.Embeddings(
+                query_schema.select_by_tag(Tags.CATEGORICAL), sequence_combiner=None
+            ),
+        ),
+        GPT2Block(d_model=48, n_head=4, n_layer=2, pre=mm.ReplaceMaskedEmbeddings()),
```
gabrielspmoreira, line 49: The `ReplaceMaskedEmbeddings()` is only needed when `SequenceMaskRandom()` is used (Masked Language Modeling. As we are using `SequencePredictNext` here for Causal Language Modeling, there will be no masked items to have embeddings replaced

oliverholworthy: Thanks @gabrielspmoreira that makes sense. I was copying this from another test that used GPT2Block and didn't make the connection with the `SequenceMaskRandom`.

I guess including it in a case like this where there are no masks doesn't affet the model? (in other words does it act as a no-op). Or would it be worth adding something to `ReplaceMaskedEmbeddings` to raise an exception if no masks are found?

marcromeyn (approved):

PR #839, opened 2022-10-31
Trying to add Bokeh to dev-requirements to fix failing xgb-tests

marcromeyn:
### Goals :soccer:
Quick fix to handle issues we are seeing in the XGB-tests that seem to be causes by a new version of `distributed`

`bokeh` is added to dev-requirements.

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,3 +1,4 @@
+bokeh
 check-manifest
 pytest>=5
 pytest-cov>=2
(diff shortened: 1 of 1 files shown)


requirements/dev.txt
```diff
@@ -1,3 +1,4 @@
+bokeh
```
oliverholworthy, line 1: maybe worth adding a comment above this one to remind us why this was required

oliverholworthy (approved):

PR #840, opened 2022-10-31
Return Ragged Tensor instead of tuple from dataloader

oliverholworthy:
### Goals :soccer:

Simplify input features to model and layers by updating dataloader to return ragged tensors for list-columns instead of tuple of (values, row_lengths).

Support TensorFlow 2.10 by avoiding bugs related to list-to-ragged conversions.

- Removes `ListToRagged`
- Renames `ListToDense`/`ListToSparse` to `RaggedToDense`/`RaggedToSparse` and drop support for tuple-like inputs.

### DIFF
--- a/merlin/models/loader/backend.py
+++ b/merlin/models/loader/backend.py
@@ -633,8 +633,14 @@ class DataLoader:
                     )
                 X[column_name] = self._to_sparse_tensor(X[column_name], column_name)
 
+        for col_name in X:
+            if isinstance(X[col_name], tuple):
+                values, offsets, diff_offsets, num_rows = self._pull_values_offsets(X[col_name])
+                X[col_name] = self._get_ragged_tensor(values, offsets, diff_offsets, num_rows)
+
         # TODO: use dict for labels as well?
         # would require output layers to match naming
         if len(self.label_names) > 1:
             labels = self._tensor_split(labels, len(self.label_names), axis=1)
+
         return X, labels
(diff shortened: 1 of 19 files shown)


gabrielspmoreira (commented):
@oliverholworthy  Indeed, the code would be much more cleaner if the Loader provides a RaggedTensor to sequence features.
How would that work with Triton? Would the model accept ragged tensors (or lists of lists) as inputs in the signature?

marcromeyn:
This is part of a bigger design discussion, closing this issue for when we have a holistic solution

oliverholworthy:
Some addtional context. I created this PR while trying to debug errors I was getting with Ragged Tensors working on #833 

It turns out that moving the ragged coercion into the dataloader didn't fix the issue. One issue with having ragged tensors as inputs to the model is that the input signature seems to lose information about the names of the features.

```
The given SavedModel SignatureDef contains the following input(s):
  inputs['args_0'] tensor_info:
      dtype: DT_INT32
      shape: (-1, 2)
      name: serving_default_args_0:0
  inputs['args_0_1'] tensor_info:
      dtype: DT_INT32
      shape: (-1)
      name: serving_default_args_0_1:0
  inputs['args_0_2'] tensor_info:
      dtype: DT_INT64
      shape: (-1)
      name: serving_default_args_0_2:0
  inputs['args_0_3'] tensor_info:
      dtype: DT_INT64
      shape: (-1, 2)
      name: serving_default_args_0_3:0
```

The issue with Ragged Tensors and TensorFlow 2.10 being tracked in this issue #858

PR #841, opened 2022-10-31
Add test for Two Tower Model with Shared Embeddings

oliverholworthy:
- Depends on #851 

### Goals :soccer:

Add a test for Two Tower Model with Shared Embeddings. To serve as an example and check that the shared embeddings functionality works.

### Testing Details :mag:

Add test for Two Tower Model with Shared Embeddings

### DIFF
--- a/tests/unit/tf/models/test_retrieval.py
+++ b/tests/unit/tf/models/test_retrieval.py
@@ -1,9 +1,11 @@
 from pathlib import Path
 
+import nvtabular as nvt
 import pytest
 import tensorflow as tf
 
 import merlin.models.tf as mm
+from merlin.core.dispatch import make_df
 from merlin.io import Dataset
 from merlin.models.tf.metrics.topk import (
     AvgPrecisionAt,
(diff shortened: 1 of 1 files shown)


tests/unit/tf/models/test_retrieval.py
```diff
@@ -18,6 +20,64 @@
...
+    outputs = user_id + item_id + user_feat
+
+    train_dataset = Dataset(train)
+
+    workflow = nvt.Workflow(outputs)
+    workflow.fit(train_dataset)
+    train = workflow.transform(train_dataset)
+    schema = train.schema
+
+    embeddings = mm.Embeddings(schema)
```
marcromeyn, line 61: Wouldn't it be easier to create a single `InputBlockV2` and use `select_by_tag` on that to get the `user_inputs` & `item_inputs`?

oliverholworthy: I've changed this to use the inputblock select_by_tag and it turns out it didn't work. Fixes to make this work in #851

marcromeyn (approved):

PR #843, opened 2022-11-01
Adding a first property to ParallelBlock

marcromeyn:
### Goals :soccer:
Tiny PR that adds a `.first` property to a `ParallelBlock`. This is useful for things like weight-tying where you could do:

```python
import merlin.models.tf as mm

inputs = mm.InputBlockV2(schema)
outputs = mm.CategoricalOutput(inputs.select_by_tag(Tags.ITEM_ID).first)
```


rnyak:
rerun tests

oliverholworthy (approved):

PR #845, opened 2022-11-02
Use merlin-dataloader package

edknv:
Fixes https://github.com/NVIDIA-Merlin/dataloader/issues/14

### Goals :soccer:
Use the new merlin-dataloader package under the hood.

- `merlin.Loader` inherits from the merlin-dataloader's `tensorflow.Loader`, mainly to make `cat_names`, `cont_names` and `label_names` available.
- `merlin.DataLoader` (the parent class of `merlin.Loader`) is completely removed.
- The `transform` in `merlin.Loader(transform=...)` was previously implemented for `PredictMaksed` in session-based models, so that we could have `mm.Loader(..., transform=PredictMasked(...))' but we didn't end up using this approach. Thus, this PR proposes to remove `transform` and revert back to using `Loader.map()`.
- ~~One notebook did not work with the new dataloader. Previously the Models dataloader worked even if a ListColumn did not have `value_count` in the schema, but the new dataloader will raise an error if a list doesn't have or infer `value_count`. he obvious solution is to the `ValueCount` op after `ListSlice`  but padding in `ListSlice` also had to be removed. See related issue: https://github.com/NVIDIA-Merlin/NVTabular/issues/1700.~~
- The files related to Models version of Torch dataloader have been simply removed. When Torch dataloader is needed in Models in the future, it should use `merlin.loader.torch.Loader` instead.

### Testing Details :mag:
All existing unit tests. Some tests are updated to work with the new format.
Updated `examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb` to work with the new dataloader.


merlin/models/tf/loader.py
```diff
@@ -289,288 +276,78 @@ def __init__(
...
+        )
+        label_names = label_names or (
+            dataset.schema.select_by_tag(Tags.TARGET).column_names if _get_schema(dataset) else []
+        )
+
         cat_names, cont_names, label_names = _validate_schema(
             feature_columns, cat_names, cont_names, label_names, schema=dataset.schema
         )
 
+        dataset.schema = _augment_schema(
```
oliverholworthy, line 300: What is the use-case for this `_augment_schema` ? Is there something that depends on this functionality?

jperez999:
Just wanted to put this here... The naming convention in the dataloader package has been changed to reduce confusion. https://github.com/NVIDIA-Merlin/dataloader/pull/57

edknv:
rerun tests

edknv:
rerun tests

examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb
```diff
@@ -1546,9 +1546,13 @@
     }
    ],
    "source": [
+    "def as_ragged(inputs, targets):\n",
+    "    _as_ragged = mm.ListToRagged()\n",
+    "    return _as_ragged(inputs), targets\n",
+    "\n",
     "history = model_bi_lstm.fit(\n",
-    "    loader,\n",
```
oliverholworthy, line 1550: why is this as_ragged conversion required here?

merlin/models/tf/loader.py
```diff
@@ -289,293 +276,85 @@ def __init__(
...
-    def __len__(self):
-        """
-        recreating since otherwise Keras yells at you
-        """
-        # TODO: what's a better way to do this inheritance
-        # of the appropriate methods? A Metaclass?
-        DataLoader.stop(self)
-        return DataLoader.__len__(self)
 
     def on_epoch_end(self):
```
oliverholworthy, line 329: This method can be removed too, now that the loader doesn't inherid from the keras Sequence class

merlin/models/tf/utils/testing_utils.py
```diff
@@ -76,7 +76,7 @@ def assert_model_is_retrainable(
 
 def model_test(
     model: Model,
-    dataset: Union[merlin.io.Dataset, Loader],
+    dataloader: Union[merlin.io.Dataset, Loader],
```
oliverholworthy, line 79: maybe a name like `dataset_or_loader` would help make this clear what the value is expected to be?

tests/unit/datasets/test_synthetic.py
```diff
@@ -50,7 +50,8 @@ def test_tf_tensors_generation_cpu():
     assert tensors["user_id"].shape == (100, 1)
     assert tensors["user_age"].dtype == tf.float64
     for name, val in filter_dict_by_schema(tensors, schema.select_by_tag(Tags.LIST)).items():
-        assert len(val) == 2
+        assert isinstance(val, tf.SparseTensor)
```
oliverholworthy, line 53: why is the dataloader now returning sparse tensors here. Previously every list feature, ragged or not was being returned by the loader as a tuple of (values, row_lengths).

examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb
edknv: Not needed. Removed.

tests/unit/tf/horovod/test_horovod.py
```diff
@@ -81,10 +81,14 @@ def test_horovod_multigpu_dlrm(
...
-        train_loader,
-        batch_size=batch_size,
-    )
+    steps = 0
+    try:
+        while True:
+            inputs, target = next(train_loader)
+            losses = model.fit(inputs, target)
+            steps += 1
+    except StopIteration:
```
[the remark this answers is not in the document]
edknv: We want to count the total number of steps in each epoch here, so `.peek` won't work. Added some comments in [1a89c9f](https://github.com/NVIDIA-Merlin/models/pull/845/commits/1a89c9f250f914bcedc3db701674c750d82821c5).

tests/unit/datasets/test_synthetic.py
edknv: Thanks for catching this. It turns out this is due to the difference in default behaviors between the old merlin.models.loader and dataloader. The new dataloader uses the dataset schema and will create sparse tensors out of all list columns by default. In the old merlin.models.loader, the `sparse_*` parameters were not used by default (and it didn't use the dataset schema) and these were never used so we almost never expected sparse tensors. I think it's safe to keep the old behavior for now. I changed sparse tensors back to optional in [02a54d8](https://github.com/NVIDIA-Merlin/models/pull/845/commits/02a54d8a45ab95a4056b43157f53a44114763672).

edknv:
rerun tests

tests/unit/tf/horovod/test_horovod.py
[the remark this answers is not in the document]
oliverholworthy: is this equivalent to looping through the batches like this without a try/except?

```python
for x, y in train_loader:
    losses = model.fit(x, y)
    steps += 1
```

tests/unit/tf/core/test_prediction.py
```diff
@@ -70,5 +70,5 @@ def flip_target(target):
     features, targets = mm.sample_batch(ecommerce_data, batch_size=100)
     outputs, context = model(features, targets=targets, training=True, output_context=True)
 
-    flipped = np.logical_not(targets["click"].numpy()).astype(np.int)
```
oliverholworthy, line 73: Why is targets no longer a dict here? This feels like it should be unrelated to the dataloader change.

tests/unit/datasets/test_synthetic.py
oliverholworthy: I've checked training a model with sparse tensors from the loader and results in the same issue described in #889 (the input signature loses it's names) if this did happen.

tests/unit/tf/horovod/test_horovod.py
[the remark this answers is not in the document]
edknv: Yes, thanks! Updated in [d6de79d](https://github.com/NVIDIA-Merlin/models/pull/845/commits/d6de79dc9e8579b8c59d4f589d243f0e3801a463).

tests/unit/tf/core/test_prediction.py
edknv: Dataloader [no longer returns a dictionary for targets](https://github.com/NVIDIA-Merlin/dataloader/blob/498543d54629216dc09a5854dca4dbeeabcab356/merlin/dataloader/loader_base.py#L626-L631), and will pick one column and use it if there are multiple target columns. I'm not sure if this is what we want, but it is how dataloader behaves for now.

edknv: Following team discussion, it looks like having dataloader output a tensor, not a dictionary, might not work for Models. We might need to change the dataloader's behavior upstream.

tox.ini
```diff
@@ -27,6 +27,8 @@ setenv =
     TF_GPU_ALLOCATOR=cuda_malloc_async
 commands =
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git
+;    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git
+    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@multiple_labels
```
edknv, line 31: Testing with https://github.com/NVIDIA-Merlin/dataloader/pull/80. Should change to upstream before merging.

edknv:
Note: Dataloader was changed to be a generator previously but the change was reverted in https://github.com/NVIDIA-Merlin/dataloader/pull/79 so dataloader is a keras sequence again. Reverted some changes that had to be made due to dataloader being a generator in [705dc6b](https://github.com/NVIDIA-Merlin/models/pull/845/commits/705dc6b573a4d79727602044517fa66b85efd161).

oliverholworthy (approved):

Issue #844, opened 2022-11-02
[BUG]  one_hot arg is not working in  ToTarget(Block) class


rnyak:
### Steps/Code to reproduce bug

I am getting the following error when I set `one_hot=True` in the ToTarget as `mm.Loader(train, batch_size=128, transform=mm.ToTarget(schema, "user_country", one_hot=True))`

```
---------------------------------------------------------------------------
InvalidArgumentError                      Traceback (most recent call last)
Cell In [11], line 2
      1 model.compile(run_eagerly=False, optimizer='adam')
----> 2 model.fit(loader, epochs=1)

File /usr/local/lib/python3.8/dist-packages/merlin/models/tf/models/base.py:856, in BaseModel.fit(self, x, y, batch_size, epochs, verbose, callbacks, validation_split, validation_data, shuffle, class_weight, sample_weight, initial_epoch, steps_per_epoch, validation_steps, validation_batch_size, validation_freq, max_queue_size, workers, use_multiprocessing, train_metrics_steps, pre, **kwargs)
    853     self._reset_compile_cache()
    854     self.train_pre = pre
--> 856 out = super().fit(**fit_kwargs)
    858 if pre:
    859     del self.train_pre

File /usr/local/lib/python3.8/dist-packages/keras/utils/traceback_utils.py:67, in filter_traceback.<locals>.error_handler(*args, **kwargs)
     65 except Exception as e:  # pylint: disable=broad-except
     66   filtered_tb = _process_traceback_frames(e.__traceback__)
---> 67   raise e.with_traceback(filtered_tb) from None
     68 finally:
     69   del filtered_tb

File /usr/local/lib/python3.8/dist-packages/tensorflow/python/eager/execute.py:54, in quick_execute(op_name, num_outputs, inputs, attrs, ctx, name)
     52 try:
     53   ctx.ensure_initialized()
---> 54   tensors = pywrap_tfe.TFE_Py_Execute(ctx._handle, device_name, op_name,
     55                                       inputs, attrs, num_outputs)
     56 except core._NotOkStatusException as e:
     57   if name is not None:

InvalidArgumentError: Graph execution error:

Detected at node 'Reshape' defined at (most recent call last):
    File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
      return _run_code(code, main_globals, None,
    File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
      exec(code, run_globals)
[...]

PR #846, opened 2022-11-02
Fixing one_hot flag in ToTarget

marcromeyn:
Fixes 
- #844

### DIFF
--- a/merlin/models/tf/transforms/features.py
+++ b/merlin/models/tf/transforms/features.py
@@ -761,7 +761,7 @@ class ToTarget(Block):
         return outputs, targets
 
     def _to_one_hot(self, name, target):
-        num_classes = self.schema[name].int_domain.max + 1
+        num_classes = schema_utils.categorical_cardinalities(self.schema)[name]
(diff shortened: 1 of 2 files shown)


merlin/models/tf/transforms/features.py
```diff
@@ -743,17 +743,29 @@ def call(
...
-                targets = inputs[name]
+                _target = inputs[name]
+                if self.one_hot:
+                    _target = self._to_one_hot(name, _target)
+                targets = _target
 
         return outputs, targets
 
+    def _to_one_hot(self, name, target):
+        num_classes = self.schema[name].int_domain.max + 1
```
sararb, line 764: Can we use the `schema_utils.categorical_cardinalities()` method instead? Just to standardize how we access the custom properties of the ColumnSchema (and avoid modifying the code in different places if these properties change in merlin-core)

oliverholworthy (approved):

marcromeyn:
rerun tests

sararb (approved):

PR #851, opened 2022-11-02
Update `ParallelBlock.select_by_tag` to handle base case and kwargs

oliverholworthy:
### Goals :soccer:

- Update `ParallelBlock.select_by_tag` to handle base case and kwargs.
  - Motivated by test in  #841

- When `ParallelBlock.select_by_tag` is used, return a new ParallelBlock with the same optional args (e.g. pre/post/aggregation/is_input)
  - This is useful when taking a sub-selection of an input_block
- Handle base case where the schema selection results in the same schema as the block itself, so that not all leaf branches need to have a `select_by_tag` implenmentation with a schema

### Testing Details :mag:

- Adds a test for ParallelBlock with kwargs and sub-selection

### DIFF
--- a/merlin/models/tf/core/combinators.py
+++ b/merlin/models/tf/core/combinators.py
@@ -483,6 +483,10 @@ class ParallelBlock(TabularBlock):
         -------
         ParallelBlock
         """
+
+        if self.schema is not None and self.schema == self.schema.select_by_tag(tags):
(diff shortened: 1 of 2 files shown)


marcromeyn (approved):

marcromeyn:
rerun tests

PR #859, opened 2022-11-03
Replaces GPT2Block by BertBlock in the masked language modeling tests

gabrielspmoreira:
### Goals :soccer:
Quick fix that replaces in the tests of masked language modeling (auto-encoding) the `GPT2Block` by `BertBlock`, as the former is originally designed for causal language modeling (auto-regressive).
Tests are typically used as reference for examples and by the users, so it is important to have them consistent on how the users should use the blocks in practice.


oliverholworthy (approved):

gabrielspmoreira:
rerun tests

Issue #857, opened 2022-11-03
[BUG] Numerical instability in loss curve over epochs when training on A100 with TF32 precision


rnyak:
### Bug description
We have been reported by a user that when they train a TT model on P4 instance on AWS they observe the unstable loss curve as below. And this is mostly because of `TF32` precision on Ampere GPU. The blue line below is `TF32` and green is `FP32` on the same machine.

![Screen Shot 2022-10-28 at 9 01 38 AM](https://user-images.githubusercontent.com/16246900/199797549-9491dde2-254c-47c3-84ab-a9f47ba9388e.png)

Currently this is being debugged and we run experiments  to diagnose the issue.

viswa-nvidia:
@rnyak , please triage this bug.

gabrielspmoreira:
@rnyak As you know, I performed a number of experiments with Two-Tower models (V1) on both V100 and A100 trying to reproduce the issue with a public dataset. I was able to find some situations where the loss became unstable, but not much due to the usage of Ampere (A100 x V100) or to the usage of the new TensorFloat32 (TF32) support in Ampere compared to FP32 ops, but to other factors like high Learning Rate.

One thing to notice is that the `loss` metric reported for each step by Keras is the cumulative average of the loss within the epoch. The plot provided above is the default logging with `TensorboardCallback`, which logs the (avg per-epoch) `loss` in the end of each epoch. To be able to better investigate the loss instability, I introduce in PR #862 the `loss_batch` metric, which reports the actual loss per-batch and not the cumulative per-epoch average `loss`. 

In the following plot for A100 runs, we have the `loss_batch` on the left for `FP32`, in the middle for TF32, and on the right the cumulative per-epoch avg. `loss` for both. 
P.s. Those peaks at the avg `loss` in the beginning of each epoch happen because the cumulative per-epoch average is more sensitive to the peaks (high values) in the first training steps of each epoch.
We can notice that with `TP32` the peaks are slightly higher than with `FP32` but the general trend of avg loss is the same.
![image](https://user-images.githubusercontent.com/504752/200353559-6f6ee68b-8eb7-499a-8832-d214318714ad.png).

The most important factor we could find related to loss instability was the usage of high LR. For example, below we plot in red the `loss_batch` with LR `1e-3` and in purple with LR `1e-4` (both with Adam optimizer on A100 with TP32). With the smaller `LR` we don't observe such instability.
[...]

PR #862, opened 2022-11-04
Adding per-batch loss metric (loss_batch)

gabrielspmoreira:
### Goals :soccer:
By default, TF Keras loss metric is the epoch cumulative average loss, which is reset every epoch as the other metric.
When the loss is unstable (i.e., does not monotonically decrease over time), like reported in #857 , it is hard to figure out in which steps we have loss spikes from the averaged loss metric.

### Goals :soccer:
- Introduce the `loss_batch` metric, that computes the actual per-batch loss (differently from the per-epoch averaged loss from Keras), as can be seen in below plot.

image

- Changed our overriden `Model` `train_step()` and `test_step()` to log the `loss_batch` metric.

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -720,8 +720,11 @@ class BaseModel(tf.keras.Model):
 
         metrics = self.train_compute_metrics(outputs, self.compiled_metrics)
 
-        # Adding regularization loss to metrics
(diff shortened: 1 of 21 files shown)


marcromeyn (approved):

gabrielspmoreira:
rerun tests

PR #863, opened 2022-11-07
Introduce SOKEmbedding using Sparse Operation Kit

WonderingWJ:
Fixes # (issue)

### Goals :soccer:
Integrate SOK into merlin-models

New a SOK Embedding class to be as the backend of InputBlock and Embeddings

### Testing Details :mag:
Currently only test sok dynamic variables


merlin/models/tf/inputs/embedding.py
```diff
@@ -150,6 +150,182 @@ def from_config(cls, config):
 
         return cls(dim, *schema, **config)
 
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
```
marcromeyn, line 153: Might be better to move this to another file, or only include this class when the user has the sok installed. This because we don't want the sok to be a required dependency.

merlin/models/tf/inputs/embedding.py
```diff
@@ -150,6 +150,182 @@ def from_config(cls, config):
...
+class SOKEmbedding(EmbeddingTableBase):
+    """
+    dim: int The last dimension of the variable
+    vocab_sizes: list, rows of the variable list 
+    initializer: string, a list of dict {"indices":numpy.array, "value": numpy.array} = "uniform"
```
marcromeyn, line 158: Would be great to enumerate which strings we accept here.

merlin/models/tf/inputs/embedding.py
```diff
@@ -150,6 +150,182 @@ def from_config(cls, config):
...
+            When it's string, it specifies the initializer used to generate initial values.
+            When it's list of numpy.array, its shape must be [vocab_size[i], embedding_vec_size],
+            and will be used as the initial indices and value.
+    use_dynamic_variable: bool = "False" use sok.DynamicVariable or sok.Variable
```
marcromeyn, line 162: Would be great to elaborate a bit more about what the difference is between the 2 types of variables.

merlin/models/tf/inputs/embedding.py
```diff
@@ -150,6 +150,182 @@ def from_config(cls, config):
 
         return cls(dim, *schema, **config)
 
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class SOKEmbedding(EmbeddingTableBase):
+    """
```
marcromeyn, line 155: A one-line summary of the class would be great here.

merlin/models/tf/inputs/embedding.py
```diff
@@ -150,6 +150,182 @@ def from_config(cls, config):
...
+                        v = sok.Variable(
+                            initializer[i],
+                            mode="localized:%d" % self._localized[i],
+                        )
+                else:
+                    raise ValueError("Wrong Configuration!!!")
+                self._trainable_weights.append(v)
+                self._vars.append(v)
+
+    def call(self, inputs, training=True):
```
marcromeyn, line 237: Should we throw in exception when a dense-tensor is provided here?

merlin/models/tf/inputs/embedding.py
WonderingWJ: > OK, any other comments

WonderingWJ: Do you have any suggestions?

merlin/models/tf/inputs/embedding.py
WonderingWJ: sok.lookup_sparse can accept Sparse and dense-tensor, but both of them should have a combiner
292 def lookup_sparse(params, sp_ids, combiners):
293     """
294     Abbreviated as ``sok.experiment.lookup_sparse``.
295
296     Peform fused sparse lookup on the given embedding ``params``. This function
297     is similar to the ``tf.nn.embedding_lookup_sparse``, but with two differences:
298
299         - It can do distributed lookup.
300         - It can accept multiple params and multiple sp_ids to do fused lookup at once,
301           which brings performance benifits.
302
303     Parameters
304     ----------
305     params: list, tuple
306             a list or tuple of trainable *sok.Variable*.
307     sp_ids: list, tuple
308             a list or tuple of tf.SparseTensor or tf.RaggedTensor.
309     combiners: list, tuple
310             a list or tuple of string to specify the combiner of each lookup.

merlin/models/tf/inputs/embedding.py
edknv: Moved to a separate file in [7555c6f](https://github.com/NVIDIA-Merlin/models/pull/863/commits/7555c6f8fbc42db7def23309b757a6c9e7fc7208).

merlin/models/tf/distributed/embedding.py
```diff
@@ -0,0 +1,202 @@
+from typing import Union
+
+import sparse_operation_kit as sok
```
marcromeyn, line 3: Maybe we should wrap this in a try/catch so we can throw an better error when it's not there

tests/unit/tf/horovod/test_embedding.py
```diff
@@ -0,0 +1,108 @@
+import numpy as np
+import pytest
+import tensorflow as tf
+
+import merlin.models.tf as mm
+from merlin.schema import ColumnSchema, Tags
+
+hvd = pytest.importorskip("horovod.tensorflow")
+sok = pytest.importorskip("sparse_operation_kit")
```
edknv, line 9: Since horovod and sok are not currently installed in the CI, the unit test here will not run automatically for this test and we have to run the test manually for now. Can we make sure that the test will succeed in the latest container `nvcr.io/nvidia/merlin/merlin-tensorflow:22.11`? (or `22.12` which will be released shortly.)

WonderingWJ: @marcromeyn

edknv (commented):
There are some trivial linting issues that need to resolved, but more importantly, I took a look at the unit tests, and I'm concerned that the proposed unit tests don't actually test the `SOKEmbedding` class that is implemented in this PR. Two tests `test_sok_dynamic_variables` and `test_distributed_variables` don't have any references to the new `SOKEmbedding` class added to Merlin Models. If I'm interpreting the code correctly, the tests are testing the correctness of `sok.DynamicVariable` and `sok.Variable` **not** the `SOKEmbedding` class. I think we should have at minimum something like:
```python
sok_embedding = SOKEmbedding(...)
outputs = sok_embedding(inputs)
assert np.isclose(outputs, expected_outputs)
```

Furthermore, why is `sok` coming from the `experiment` sub-module in `sparse_operation_kit`? Is this how end users are expected to use this class? What I'm mainly concerned about is the `experiment` namespace of the sub-module because it says `experiment`. The [documentation](https://nvidia-merlin.github.io/HugeCTR/sparse_operation_kit/master/api/init.html) says we are supposed to use something like:
```python
import horovod.tensorflow as hvd
import sparse_operation_kit as sok

hvd.init()

sok.Init(**kwargs)
```
Is this not correct anymore? Does the unit test correctly reflect the actual use case in the real world?

tests/unit/tf/horovod/test_embedding.py
```diff
@@ -0,0 +1,105 @@
...
+
+import merlin.models.tf as mm
+from merlin.schema import ColumnSchema, Tags
+
+hvd = pytest.importorskip("horovod.tensorflow")
+sparse_operation_kit = pytest.importorskip("sparse_operation_kit")
+from sparse_operation_kit import experiment as sok
+
+hvd.init()
+sok.init()
```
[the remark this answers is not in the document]
WonderingWJ: @edknv The reason that I only test `sok.DynamicVariable` and `sok.Variable` is that currently SOKEmbedding has no `load` and 'dump' interface. So after I implement the interface I will test `SOKEmbedding` directly.
Thank you for your advice. I've already modify the import code based on your suggetion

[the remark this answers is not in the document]
WonderingWJ: @edknv 
```
import horovod.tensorflow as hvd
import sparse_operation_kit as sok

hvd.init()

sok.Init(**kwargs)
```
This is the old way of `init`. We are developing the new generation of SOK `sok.experiment`. We will release the experiment code as `sok` soon.

edknv:
@WonderingWJ I added unit tests for `SOKEmbedding` in #980. Can you please review it and make sure that the way I'm using it is correct, and either merge it to this branch or use it as a reference to write similar tests? The tests failed due to missing variables when I tested locally. Please fix the errors.

edknv:
Does SOK work with CPUs? We have unit tests in the CI for horovod dependencies but it's currently set up only for the CPU. When I tried to install SOK in the CPU test, it seems to complain about a missing object file:
```
[1,0]<stdout>:E   tensorflow.python.framework.errors_impl.NotFoundError: /home/runner/work/models/models/.tox/py38-horovod-cpu/env/lib/python3.8/site-packages/sparse_operation_kit/lib/libsparse_operation_kit_compat_ops.so: undefined symbol: _ZNK10tensorflow8OpKernel11TraceStringB5cxx11ERKNS_15OpKernelContextEb
```
(https://github.com/NVIDIA-Merlin/models/actions/runs/4186557543/jobs/7255213965)
GPU tests are still a work in progress, so the new tests added in `tests/unit/tf/horovod/test_embedding.py` have to be run manually for now.

WonderingWJ:
> 

I thing the reason of the missing object is that different ABI of tensorflow.  Can you provide more info?

edknv:
@WonderingWJ I didn't mean the error in the CI. As I mentioned before, we are working on setting up the CI to run the tests for the ones that require horovod (#1009), but currently it won't run the tests that you write under `tests/unit/tf/horovod` yet, so you will have to run them manually for now. You can run them manually by using the production image `nvcr.io/nvidia/merlin/merlin-tensorflow:22.12` and by doing something like:
```shell
$ docker run --rm -it --gpus all nvcr.io/nvidia/merlin/merlin-tensorflow:22.12 bash
root@3ad510246a5d:/opt/tritonserver# cd /models/                                                                                                                                                                                            
root@3ad510246a5d:/models# git fetch origin fea-sok-integration-wj
root@3ad510246a5d:/models# git checkout fea-sok-integration-wj
root@3ad510246a5d:/models# CUDA_VISIBLE_DEVICES=0 pytest tests/unit/tf/horovod/test_embedding.py
```
or to run it on multiple GPUs using horovod:
```
root@3ad510246a5d:/models# horovodrun -np 2 sh examples/usecases/multi-gpu/hvd_wrapper.sh python -m pytest tests/unit/tf/horovod/test_embedding.py
```

edknv:
@WonderingWJ Sorry, it looks like I was indeed talking about a CI error in `horovod (cpu)`. Sorry for the confusion. I was trying to install sok in `horovod (cpu)` (which doesn't have GPUs) and was running into installation issues. But let's not worry about that now. I think there are other major issues to take care of, which you can see by running the tests locally. Or hopefully #1009 is merged soon so you can use the CI to see the errors.

WonderingWJ:
> tests

@edknv  Thank you! After the MR is merged, I will test it.

edknv:
@WonderingWJ #1009 has been merged. Do we need to do anything other than `pip install sparse_operation_kit` to install SOK? I pushed a commit for pip-installing SOK in b716adc and we are getting the following:
```
tensorflow.python.framework.errors_impl.NotFoundError: /home/gha-user/gha1/models/models/.tox/py38-multi-gpu/lib/python3.8/site-packages/sparse_operation_kit/lib/libsparse_operation_kit.so: undefined symbol: cusparseDestroy
```
([link to GHA logs](https://github.com/NVIDIA-Merlin/models/actions/runs/4367233076/jobs/7638278955))
Maybe it's an issue with tox and we need to pass some environment variable or something.

Also some errors due to missing variables:
```
[1,1]<stdout>:>       return cls(
[1,1]<stdout>:            dim,
[1,1]<stdout>:            col_schema,
[1,1]<stdout>:            name=name,
[1,1]<stdout>:            initializer=weights,
[1,1]<stdout>:            use_dynamic_variable=use_dynamic_variable,
[1,1]<stdout>:            localized=localized,
[1,1]<stdout>:            trainable=trainable,
[1,1]<stdout>:            **kwargs,
[1,1]<stdout>:        )
[1,1]<stdout>:E       TypeError: __init__() missing 1 required keyword-only argument: 'vocab_sizes'
[1,1]<stdout>:
[1,1]<stdout>:merlin/models/tf/distributed/embedding.py:179: TypeError
```
([link to another GHA logs](https://github.com/NVIDIA-Merlin/models/actions/runs/4366713042/jobs/7637082937))

edknv:
@WonderingWJ If it's not so straightforward to install sok in the CI, we could also manually test this PR as described in [my previous comment](https://github.com/NVIDIA-Merlin/models/pull/863#issuecomment-1456516130), merge this PR as soon as the manual testing passes, and tackle the CI in a separate PR.

WonderingWJ:
@edknv 
```
merlin/models/tf/distributed/embedding.py:224: in from_config
    return cls(dim, *schema, vocab_size, initializer, use_dynamic_variable, localized, **config)
merlin/models/tf/distributed/embedding.py:66: in __init__
    super(SOKEmbedding, self).__init__(
merlin/models/tf/inputs/embedding.py:80: in __init__
    self.add_feature(col_schema)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = SOKEmbedding(
  (features): Dict(
    (item_id): ColumnSchema(name='item_id', tags={<Tags.CATEGORICAL: 'categorical'>}...rties={'domain': {'min': 0, 'max': 10, 'name': 'item_id'}}, dtype=dtype('int32'), is_list=False, is_ragged=False)
  )
)
col_schema = List(
  (0): 10
)

    def add_feature(self, col_schema: ColumnSchema) -> None:
        """Add a feature to the table.

        Adding more than one feature enables the table to lookup and return embeddings
        for more than one feature when called with tabular data (Dict[str, TensorLike]).

        Additional column schemas must have an int domain that matches the existing ones.

        Parameters
        ----------
        col_schema : ColumnSchema
        """
>       if not col_schema.int_domain:
E       AttributeError: 'ListWrapper' object has no attribute 'int_domain'

merlin/models/tf/inputs/embedding.py:116: AttributeError
```
Can you help me with the error? I don't know what's wrong with the col_schema. Thank you!

edknv:
It looks to me like the wrong values are being passed to `col_schema` because of the variable-length argument `*schema` in
```
merlin/models/tf/distributed/embedding.py:224: in from_config
    return cls(dim, *schema, vocab_size, initializer, use_dynamic_variable, localized, **config)
```
I think the following in the log is incorrect:
```
col_schema = List(
  (0): 10
)
```
and `[10]` should actually be passed to `vocab_size`?

Maybe use the keyword arguments so that the variable-length argument `*schema` can be parsed correctly?
```
return cls(dim, *schema, vocab_size=vocab_size, initializer=initializer, use_dynamic_variable=use_dynamic_variable, localized=localized, **config)
```

WonderingWJ:
> length

Thank you for your reply, it doesn't help. I'm confused....

edknv (approved):

edknv:
@marcromeyn I'm okay with merging this if you are.

edknv:
The implementation is pretty much isolated and independent of the rest of the library, so it should be good to merge. I will go ahead and merge this. If there are any concerns, we can address them later.

PR #871, opened 2022-11-09
Update version of core for gpu tests to run from release-22.11

oliverholworthy:
Temporarily unblock merging of PRs to Merlin Models by update version of core for gpu tests to run from release-22.11


oliverholworthy:
This was a potential option to unblock a PR for the release. However, we decided that we'll leave it for the next one

PR #872, opened 2022-11-09
Pin fiddle to 0.2.2 so that tests run with pip install

oliverholworthy:
### Goals :soccer:

Enable tests to run in Python 3.8

Pins dev dependency `fiddle` to enable tests to run Python 3.8 environment.

This causes errors in Models tests in core PRs.

### Testing Details :mag:

n/a


oliverholworthy:
rerun tests

edknv (approved):

PR #875, opened 2022-11-10
Update tox.ini to not use sitepackages

edknv:


edknv:
Closing in favor of 872.

PR #876, opened 2022-11-10
revert tox usage in ci unit test script

jperez999:
Tox should only be used when you need to build an environment. CI scripts in the ci folder are meant to be leveraged by the container build process, we do NOT want to use tox to create a new environment. We are testing to make sure that the stuff in the containers has everything necessary to run the targeted commit for the project.


jperez999:
rerun tests

PR #886, opened 2022-11-15
Save output schema of model and add save method to `Encoder`

### LINKED ISSUE #677: [Task] Implement save method on Tensorflow implementation of models.

oliverholworthy:
Fixes #677

### Goals :soccer:

Save output schema of model alongside input schema and saved model.

Extracting output schema from the saved model signature. [In the same way that we have been doing in Merlin Systems](https://github.com/NVIDIA-Merlin/systems/blob/v0.7.0/merlin/systems/dag/ops/tensorflow.py#L168-L173)

### Testing Details :mag:

Updates save/load test to check that saved schema column names match input/output signatures


merlin/models/tf/models/base.py
```diff
@@ -883,7 +916,7 @@ def fit(
 
         # Bind schema from dataset to model in case we can't infer it from the inputs
         if isinstance(x, Loader):
-            self.schema = x.schema
+            self.schema = x.schema.excluding_by_tag(Tags.TARGET)
```
oliverholworthy, line 919: ensuring that `self.schema` only contains input features and not targets

sararb (approved):
Thank you for the PR! It looks good to me.

Issue #889, opened 2022-11-16
[BUG] Incorrect Input Signature of Saved Model with Ragged Features and Loader transform 


oliverholworthy:
### Bug description

When using a `SequenceTransform` as a Loader transformation as part of model training. The saved the model input signature names change to generic names [`arg_0`, `arg_0_1`, `args_0_2`, ...]

### Steps/Code to reproduce bug

1. Run test `test_transformer_with_causal_language_modeling` with a `model.save(tmpdir)`
2. Check the output signature

```
saved_model_cli show --tag_set serve --signature_def serving_default --dir <path-to-saved-model

The given SavedModel SignatureDef contains the following input(s):
  inputs['args_0'] tensor_info:
      dtype: DT_INT64
      shape: (-1)
      name: serving_default_args_0:0
  inputs['args_0_1'] tensor_info:
      dtype: DT_INT32
      shape: (-1)
      name: serving_default_args_0_1:0
  inputs['args_0_10'] tensor_info:
      dtype: DT_FLOAT
      shape: (-1)
      name: serving_default_args_0_10:0
  inputs['args_0_11'] tensor_info:
      dtype: DT_INT32
      shape: (-1)
      name: serving_default_args_0_11:0
  inputs['args_0_12'] tensor_info:
      dtype: DT_INT64
      shape: (-1)
      name: serving_default_args_0_12:0
  inputs['args_0_13'] tensor_info:
      dtype: DT_INT32
      shape: (-1)
      name: serving_default_args_0_13:0
  inputs['args_0_14'] tensor_info:
      dtype: DT_INT64
      shape: (-1, 1)
      name: serving_default_args_0_14:0

...

The given SavedModel SignatureDef contains the following output(s):
  outputs['item_id_seq/categorical_output'] tensor_info:
      dtype: DT_FLOAT
      shape: (-1, -1, 51997)
      name: StatefulPartitionedCall:0
Method name is: tensorflow/serving/predict
```

### Expected behavior

Names of inputs in saved model signature should match input feature names

### Environment details
- Merlin version: 22.10
- Python version: 3.8.10
- Tensorflow version (GPU): [`2.9.1+nv22.8`, `2.9.2`]

oliverholworthy:
The same issue occurs with SparseTensor's passed as inputs to the model.

PR #890, opened 2022-11-16
Support `tuple` return type from model `pre` and update test to use this

oliverholworthy:
Relates to #889 

### Goals :soccer:

Support `SequenceTransform` layers passed to `pre` arg of `model.evaluate` and `model.predict`.

- Handles tuple return type from `pre` layer in `model.evaluate` and `model.predict`
  - this is already supported in `model.fit` 

### Testing Details :mag:

- Update `test_transformer_with_causal_language_modeling` to use `pre` arg instead of loader transform.
  - while #889 remains an issue, this approach doesn't produce a model that is easy to serve.

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -742,6 +742,11 @@ class BaseModel(tf.keras.Model):
             out = call_layer(self.test_pre, x, targets=y, features=x, testing=True)
             if isinstance(out, Prediction):
                 x, y = out.outputs, out.targets
+            elif isinstance(out, tuple):
+                assert (
+                    len(out) == 2
+                ), "output of `pre` must be a 2-tuple of x, y or `Prediction` tuple"
+                x, y = out
             else:
                 x = out
 
@@ -771,6 +776,11 @@ class BaseModel(tf.keras.Model):
             out = call_layer(self.predict_pre, x, features=x, training=False)
             if isinstance(out, Prediction):
                 x = out.outputs
+            elif isinstance(out, tuple):
+                assert (
+                    len(out) == 2
+                ), "output of `pre` must be a 2-tuple of x, y or `Prediction` tuple"
+                x, y = out
             else:
                 x = out
 
--- a/tests/unit/tf/transformers/test_block.py
+++ b/tests/unit/tf/transformers/test_block.py
@@ -238,7 +238,7 @@ def test_transformer_with_causal_language_modeling(sequence_testing_data: Datase
     target = sequence_testing_data.schema.select_by_tag(Tags.ITEM_ID).column_names[0]
     predict_next = mm.SequencePredictNext(schema=seq_schema, target=target)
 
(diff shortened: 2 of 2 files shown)


oliverholworthy:
> Are we going to face the serving issue https://github.com/NVIDIA-Merlin/models/issues/889 for any Loader transform that returns a RaggedTensor? 

Yes we'll face this issue for any loader with ragged or sparse tensors in the input.

> If that is the case, what about raising an exception in the Loader when a RaggedTensor is returned by the Loader?

I've added a check to the `train_step` to raise an exception if the inputs contain ragged or sparse tensors. And updated all the tests that started failing as a result of this assertion.

oliverholworthy:
rerun tests

merlin/models/tf/models/base.py
```diff
@@ -736,6 +736,15 @@ def train_step(self, data):
         with tf.GradientTape() as tape:
             x, y, sample_weight = unpack_x_y_sample_weight(data)
 
+            # Ensure that we don't have any ragged or sparse tensors passed at training time.
+            if isinstance(x, dict):
+                for k in x:
+                    if isinstance(x[k], (tf.RaggedTensor, tf.SparseTensor)):
+                        raise ValueError(
+                            "Training with RaggedTensor or SparseTensor inputs is unsupported. "
```
gabrielspmoreira, line 744:
```suggestion
                            "Training with RaggedTensor or SparseTensor inputs is supported "                            
                            "for training and evaluation, but not for inference "
                            "(with Triton Inference Server or TFServing) because the "
                            "input feature names are lost in SavedModel when ragged "
                            and sparse tensors are fed as input."
```

merlin/models/tf/models/base.py
```diff
@@ -736,6 +736,15 @@ def train_step(self, data):
...
+                            "Please update your loader to pass dense tensors. "
```
gabrielspmoreira, line 745:
```suggestion
                            "Please update your loader to pass dense tensors so that you can save the model, "
                            "and serve it correctly".
```

gabrielspmoreira (commented):
@oliverholworthy I saw you updated the YouTubeDNN tests to avoid sequence-based Loader transforms, but I think we still need to do the same for these tests, so that they don't work as an anti-pattern for our users:
- test_seq_predict_next
- test_seq_predict_last
- test_seq_predict_random

What do you think?

oliverholworthy:
I've removed the loader tests from the test_seq_predict_* tests as suggested to avoid any confusion from those looking at these tests. If we can figure out how to make it work with ragged/sparse inputs in future we might be able to add this back.

merlin/models/tf/models/base.py
oliverholworthy: added something along these lines, but needed to split across more lines due to line length constraints

marcromeyn (approved):

gabrielspmoreira (approved):
Sounds good to me

PR #897, opened 2022-11-20
Skip retrieval_with_hpo notebook test when optuna is not installed

edknv:
The retrieval_with_hpo notebook requires `optuna` but the library is not installed at the container level. As a result, the notebook test will fail when we run `pytest` from the `merlin-tensorflow` container or when we run the tests on Jenkins. This PR proposes to skip all tests in `test_usecase_retrieval_with_hpo.py` if `optuna` is not installed. The test will still run on GIthub actions or when we run tox, because it will be installed in [requirements/dev.txt](https://github.com/NVIDIA-Merlin/models/blob/f0805ad747c1ac783e60068ab22fb4980020c090/requirements/dev.txt#L19).

### DIFF
--- a/tests/unit/tf/examples/test_usecase_retrieval_with_hpo.py
+++ b/tests/unit/tf/examples/test_usecase_retrieval_with_hpo.py
@@ -3,10 +3,7 @@ from testbook import testbook
 
 from tests.conftest import REPO_ROOT
 
-try:
-    import optuna  # noqa: F401
-except ImportError:
-    pytest.skip("No module named 'optuna'", allow_module_level=True)
+optuna = pytest.importorskip("optuna")
 
 
 @testbook(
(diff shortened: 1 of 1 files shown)


tests/unit/tf/examples/test_usecase_retrieval_with_hpo.py
```diff
@@ -1,7 +1,13 @@
+import pytest
 from testbook import testbook
 
 from tests.conftest import REPO_ROOT
 
+try:
+    import optuna  # noqa: F401
+except ImportError:
+    pytest.skip("No module named 'optuna'", allow_module_level=True)
```
oliverholworthy, line 9: could this be achieved with `pytest.importorskip` ?

edknv: Yes, they are equivalent. Changed to `importorskip` in [727f142](https://github.com/NVIDIA-Merlin/models/pull/897/commits/727f142de46a8fcf65827239d94d36c4779505dd).

oliverholworthy (approved):

PR #899, opened 2022-11-22
Add skip if transformers not available for next item prediction notebook test

oliverholworthy:
### Goals :soccer:

Enable correct notebook tests to run when merlin models installed in env without `transformers` dependency

- renames test for next item prediction notebook for clarify
- adds skip for test if transformers is not installed

### Testing Details :mag:

changes to next item prediction notebook test.


rnyak (approved):

rnyak:
rerun tests

Issue #898, opened 2022-11-21
[BUG] SavedModel serving signature contains all input features even when they aren't used


karlhigley:
### Bug description
When I train a model and save it, the resulting saved model file's serving signature contains every feature from the training dataset that was loaded by the dataloader, even when the model doesn't actually use those features.

### Steps/Code to reproduce bug
Run the following code:
```python
import tensorflow
import merlin.models.tf as mm
from merlin.datasets.synthetic import generate_data
from merlin.io import Dataset
from merlin.schema import Tags


train = generate_data("sequence-testing", num_rows=100)

seq_schema = train.schema.select_by_tag(Tags.SEQUENCE).select_by_tag(Tags.CATEGORICAL)
target = train.schema.select_by_tag(Tags.ITEM_ID).column_names[0]

predict_last = mm.SequencePredictLast(schema=seq_schema, target=target)

input_schema = seq_schema
output_schema = seq_schema.select_by_name(target)

loader = mm.Loader(train, batch_size=16, shuffle=False)

d_model = 48
query_encoder = mm.Encoder(
    mm.InputBlockV2(
        input_schema,
        embeddings=mm.Embeddings(
            input_schema.select_by_tag(Tags.CATEGORICAL), sequence_combiner=None
        ),
    ),
    mm.MLPBlock([d_model]),
    mm.GPT2Block(d_model=d_model, n_head=2, n_layer=2),
    tf.keras.layers.Lambda(lambda x: tf.reduce_mean(x, axis=1)),
)

model = mm.RetrievalModelV2(
    query=query_encoder,
    output=mm.ContrastiveOutput(output_schema, negative_samplers="in-batch"),
)

model.compile(metrics={})
model.fit(loader, epochs=1, pre=predict_last)

query_encoder.save(
    f"/tmp/query_encoder"
)
```
Examine the serving signature with:
```
saved_model_cli show --tag_set serve --signature_def serving_default --dir /tmp/query_encoder
```

### Expected behavior
The serving signature should only contain the features that the model actually uses. (Otherwise, when we try to serve the model, we have to provide a bunch of features that don't do anything.)

### Environment details
- Merlin versions:
  - merlin-core                  0.8.0+12.g8612b749e3  
  - merlin-dataloader            0.0.2                 
  - merlin-models                0.9.0+42.g04597b9277  
  - merlin-systems               0.7.0+19.g032be4d9 
- Platform: Docker (merlin_ci_runner image)
- Python version: 3.8.10
[...]

PR #904, opened 2022-11-25
Ensure saved model input signature contains only required features

oliverholworthy:
Relates to #898 

### Goals :soccer:

Ensure that the input signature of a saved merlin model contains only those required for the inputs of the model

- The saved model signautre generated by keras/tensorflow includes all features passed to the model at training time, even if some are ignored by the input layers.
- Adding a property/setter for schema on DataLoader to enable changing of the loader schema after it's initialized
- If a schema is provided to the `InputBlockV2`. Use this to set the schema on the loader.
  - This ensures that the loaded will only return required features and pass them to the fit method at training time.

### Testing Details :mag:

- Adds a test that checks the input signature with an inputblock that has a schema with fewer input features than the dataset


jperez999:
rerun tests

karlhigley:
I'm not the right reviewer for this since I'm not very familiar with the Models code base, but assuming it does what it says on the tin (and I trust that @oliverholworthy's work does so more than just about anyone else's including mine), it sounds like a reasonable solution to me.

merlin/models/tf/outputs/topk.py
```diff
@@ -100,7 +100,7 @@ def extract_ids_embeddings(self, data: merlin.io.Dataset, check_unique_ids: bool
         if check_unique_ids:
             self._check_unique_ids(data=data)
         values = tf_utils.df_to_tensor(data)
-        ids = tf_utils.df_to_tensor(data.index)
+        ids = tf_utils.df_to_tensor(make_df({"index": data.index}))
```
oliverholworthy, line 103: This breaks with the newer versions of cudf. A RangeIndex no longer has an `empty` property and won't work with this df_to_tensor function.

oliverholworthy:
This PR is ready to review/merge now.

edknv (approved):

PR #909, opened 2022-12-01
Fix the inference of transformer-based models trained with masked language modeling

### LINKED ISSUE #907: [BUG]  Transformer-based model's performance discrepancy between evaluation and inference modes
### Bug description
More details about the bug can be found in Transformers4Rec ([here](https://github.com/NVIDIA-Merlin/Transformers4Rec/issues/525) and [here](https://github.com/NVIDIA-Merlin/Transformers4Rec/pull/551))

The experiments and bug reproducibility were defined using Transformers4Rec. 

### Expected behavior
The objective of this task is to port the solution proposed in T4Rec to Merlin Models

sararb:
Fixes #907 

### Goals :soccer:
This PR implements two blocks needed to support the inference of transformer-based next-item prediction models in Merlin Models.
For that, the user should specify two specific blocks: `SequenceMaskLastInference` and `inference_hidden_state` as follows: 
```
BertBlock(
            d_model=48,
            n_head=8,
            n_layer=2,
            pre=mm.SequentialBlock([SequenceMaskLastInference(), mm.ReplaceMaskedEmbeddings()]),
            post="inference_hidden_state",
)
```

- Implement the `SequenceMaskLastInference` block that generates the boolean mask to indicate the position of the next-item during inference.

- Implement the block InferenceHiddenState that extracts the hidden representation related to the `next-item`, from the mask generated by `SequenceMaskLastInference`. 

### Testing Details :mag:

- update the test `test_transformer_with_masked_language_modeling` to showcase the inference logic.


merlin/models/tf/transforms/sequence.py
```diff
@@ -621,6 +621,41 @@ def from_config(cls, config):
...
+        This method is called by Keras after call()
+        and returns the mask that is going to be assigned
+        to the input tensors, being accessible
+        by tensor._keras_mask
+        """
+
+        targets_mask = None
+        if self.inference_mode:
+            if isinstance(inputs, tf.RaggedTensor):
+                row_lengths = inputs.row_lengths(1) + 1
```
sararb, line 648: The inputs passed here are the original ones (not the inputs returned by the call method) so we need to extend the lengths by 1.

merlin/models/tf/transformers/block.py
```diff
@@ -91,7 +91,7 @@ def __init__(
             self.transformer = get_tf_main_layer(transformer)
         else:
             self.transformer = transformer
-
+        self.transformer.supports_masking = True
```
sararb, line 94: set `supports_masking=True` so that the transformer layer can forward the `mask` tensor created by ` SequenceMaskLastInference`

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tests

jperez999:
rerun tasks

jperez999:
rerun tests

merlin/models/tf/transforms/sequence.py
```diff
@@ -664,11 +705,17 @@ def call(
...
+        training : bool, optional
+            Flag that indicates whether in training mode, by default True
+        testing : bool, optional
+            Flag that indicates whether in evaluation mode, by default True
         Returns
         -------
         Union[tf.Tensor, tf.RaggedTensor]
             If training, returns a tensor with the masked inputs replaced by the dummy embedding
         """
+        if not training and not testing:
```
gabrielspmoreira, line 717: Don't we need to replace the masked (selected) item with the trainable embedding even in inference mode, to avoid that issue you found in Transformers4Rec when you don't replace the input by the trainable embedding?

merlin/models/tf/transforms/sequence.py
gabrielspmoreira: Interesting. It makes sense then to extend row lengths of the masks manually.

merlin/models/tf/transformers/transforms.py
```diff
@@ -37,10 +37,62 @@ class LastHiddenState(Layer):
...
+        super().__init__(**kwargs)
+        self.supports_masking = True
+
     def call(self, inputs: TFBaseModelOutputWithPoolingAndCrossAttentions):
         return inputs.last_hidden_state
 
 
+@Block.registry.register("inference_hidden_state")
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class InferenceHiddenState(Layer):
```
gabrielspmoreira, line 50: I don't see in the PR where `InferenceHiddenState` is going to be used yet...

merlin/models/tf/transformers/transforms.py
```diff
@@ -37,10 +37,62 @@ class LastHiddenState(Layer):
...
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self.supports_masking = True
+
+    def call(
+        self,
+        inputs: tf.Tensor,
+        training: bool = False,
+        testing: bool = False,
+        mask: tf.Tensor = None,
```
gabrielspmoreira, line 64: Why do we use `supports_masking` here, and also pass `mask` in call()?
I should be one or the other I guess.
How is going to provide the mask to the `call()`?
Can't you extract the mask from `inputs._keras_mask` ?

merlin/models/tf/transformers/transforms.py
```diff
@@ -37,10 +37,62 @@ class LastHiddenState(Layer):
...
+        Returns
+        -------
+        tf.Tensor
+            If inference, returns a 2-D tensor with the hidden states of
+            the target position
+        """
+        if not training and not testing:
+            if mask is not None:
+                if isinstance(mask, tf.RaggedTensor):
+                    mask = mask.to_tensor()
```
gabrielspmoreira, line 91: I think we should convert the mask from Ragged to Tensor in the sample place where the actual ragged input is converted to dense, by checking if mask is defined and is ragged. For example

```python
if getattr(inputs, "_keras_mask", None) is not None and isinstance(inputs._keras_mask, tf.RaggedTensor):
    inputs._keras_mask = inputs._keras_mask.to_tensor()
```

merlin/models/tf/transforms/sequence.py
```diff
@@ -621,6 +621,41 @@ def from_config(cls, config):
         return cls(schema, target, **config)
 
 
+@tf.keras.utils.register_keras_serializable(package="merlin.models")
+class SequenceMaskLastInference(Block):
```
gabrielspmoreira, line 625: Wondering if we can simplify a bit the API for the user by making `SequenceMaskLastInference` a default `pre` of `ReplaceMaskedEmbeddings`

merlin/models/tf/transformers/transforms.py
sararb: It is used in the unit test using its string representation the parameter `post="inference_hidden_state"`

merlin/models/tf/transformers/transforms.py
sararb: Os from Keras [documentation](https://www.tensorflow.org/guide/keras/masking_and_padding#supporting_masking_in_your_custom_layers), I understand that : 
* `supports_masking` is used to tell the block to forward the input mask to the following layer
* if you want to use the actual `mask` information inside your block, they explained [here](https://www.tensorflow.org/guide/keras/masking_and_padding#supporting_masking_in_your_custom_layers) that you just need to set `mask=None` in the call method. Keras automatically extracts the mask from `inputs._keras_mask` and pass it as additional param to the call function.  so using `inputs._keras_mask` is equivalent to setting `mask` as a param of the module's call. We can

merlin/models/tf/transforms/sequence.py
sararb: You're right! I will update the logic to replace with the `mask` embeddings

merlin/models/tf/transforms/sequence.py
sararb: That's a good point. `ReplaceMaskedEmbeddings` currently inherits from the `Block` class, which doesn't have a `pre` argument. 

We also discussed the solution to instantiate `ReplaceMaskedEmbeddings` as a sub-class of TabularBlock so as to use the `pre` property. However, `TabularBlock` is not updated to use the recent logic of `call_quentially`/`to_call` introduced in MM combinators, outputs, and encoder blocks. Instead, it relies on `_tabular_call` method that does not forward the **kwargs argument to the `pre` layer, such as training/testing flags.  

So, in the meantime, I think we should expose the `SequenceMaskLastInference` to the user. I also added a description in the `ReplaceMaskedEmbeddings` docstrings to explain the importance of `SequenceMaskLastInference`.

merlin/models/tf/transformers/transforms.py
sararb: I updated the method to use the logic of  `inputs._keras_mask` instead, so as to align with other Merlin Models blocks

merlin/models/tf/transformers/transforms.py
sararb: I move the conversion to the `PrepareTransformerInputs` block

merlin/models/tf/transformers/transforms.py
```diff
@@ -37,10 +37,58 @@ class LastHiddenState(Layer):
...
+        super().__init__(**kwargs)
+        self.supports_masking = True
+
+    def call(
+        self,
+        inputs: tf.Tensor,
+        training: bool = False,
+        testing: bool = False,
+    ):
+        """Select the hidden state of the target position, during inference.
```
gabrielspmoreira, line 65:
```suggestion
        """Select the hidden state of the target (last) position, during inference.
```

merlin/models/tf/transformers/transforms.py
```diff
@@ -37,10 +37,58 @@ class LastHiddenState(Layer):
...
+
+        Returns
+        -------
+        tf.Tensor
+            If inference, returns a 2-D tensor with the hidden states of
+            the target position
+        """
+        if not training and not testing:
+            if getattr(inputs, "_keras_mask", None) is not None:
+                inputs = tf.reshape(
```
gabrielspmoreira, line 86: I think we can add an assert here to check if the resulting inputs after `boolean_mask()` has the 1st dim as before. Because we are trusting here that during inference only **one position item (the last one) is masked per example**, otherwise the batch size will differ after this conversion. Adding such assert should help users to understand this is an assumption when using this block.

gabrielspmoreira (commented):
I undestand this PR is focused in Masked Language Modeling. After reviewing, I am wondering if for Causal Language Modeling we shouldn't return during inference the last non-padded prediction per sequence. The thing is that we don't use `_keras_mask` for causal language modeling. 
Thoughts?
-

gabrielspmoreira (approved):

Issue #924, opened 2022-12-15
[BUG] Model schemas and serving signatures don't align with each other


karlhigley:
### Bug description
The saved input and output schema files for Tensorflow models contain information that doesn't match the saved model's serving signature. The serving signature also seems to record dimensions inconsistently.

### Steps/Code to reproduce bug
```
import pytest
import tensorflow as tf

import merlin.models.tf as mm
from merlin.datasets.synthetic import generate_data
from merlin.io import Dataset
from merlin.schema import Tags
from merlin.schema.io.tensorflow_metadata import TensorflowMetadata

def test_model_schemas_match_serving_signature(tmpdir):

    train = generate_data("sequence-testing", num_rows=100)

    seq_schema = train.schema.select_by_tag(Tags.SEQUENCE).select_by_tag(Tags.CATEGORICAL)

    target = train.schema.select_by_tag(Tags.ITEM_ID).column_names[0]
    predict_last = mm.SequencePredictLast(schema=seq_schema, target=target)

    input_schema = seq_schema
    output_schema = seq_schema.select_by_name(target)

    train = Dataset(train.to_ddf(columns=input_schema.column_names).compute())
    loader = mm.Loader(train, batch_size=16, shuffle=False)

    d_model = 48
    query_encoder = mm.Encoder(
        mm.InputBlockV2(
            input_schema,
            embeddings=mm.Embeddings(
                input_schema.select_by_tag(Tags.CATEGORICAL), sequence_combiner=None
            ),
        ),
        mm.MLPBlock([d_model]),
        mm.GPT2Block(d_model=d_model, n_head=2, n_layer=2),
        tf.keras.layers.Lambda(lambda x: tf.reduce_mean(x, axis=1)),
    )

    model = mm.RetrievalModelV2(
        query=query_encoder,
        output=mm.ContrastiveOutput(output_schema, negative_samplers="in-batch"),

    )

    model.compile(metrics={})
    model.fit(loader, epochs=1, pre=predict_last)

    query_encoder.save(
        f"{tmpdir}/query_encoder"
    )

    input_schema = TensorflowMetadata.from_json_file(f"{tmpdir}/query_encoder/.merlin/input_schema.json").to_merlin_schema()
    output_schema = TensorflowMetadata.from_json_file(f"{tmpdir}/query_encoder/.merlin/output_schema.json").to_merlin_schema()

    breakpoint()
```
Then compare the input and output schema (and especially the value counts) to the serving signature:
```
[...]

PR #926, opened 2022-12-19
Add value_count property if output is fixed size tensor

oliverholworthy:
Partial Fix for #924 (issue)

### Goals :soccer:

- Provide information about the output schema that is saved with `model.save`
  -` Adding `value_count` property if output is fixed size tensor

Save value_count property in output schema when the output is of a fixed size.

### Testing Details :mag:

Adds a test for `get_output_schema` function.


oliverholworthy:
rerun tests

karlhigley (approved):

Issue #908, opened 2022-11-30
[BUG] Error when loading the session-based model in example


FredHJC:
# ❓ Questions & Help

## Details

https://github.com/NVIDIA-Merlin/models/blob/main/examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb

We are trying to implement the session-based models shown in the above notebook. However, there is a consistent error when loading the saved model. `TypeError: ('Keyword argument not understood:', 'layer was saved without config')`

We can just run the example notebook and save the trained model. Then the error occurs when trying to load it. It looks like customized layers / prediction tasks should be handled with a manually specified config.

<img width="835" alt="Screen Shot 2022-11-30 at 1 59 41 PM" src="https://user-images.githubusercontent.com/24794992/204885225-48809ea7-b1e8-4b88-84ee-851584ee1800.png">

zhiruiwang:
### Steps/Code to reproduce bug

1. Run the [Dressipi notebook](https://github.com/NVIDIA-Merlin/models/blob/main/examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb) Transformer-based model example after the model is trained:  
```
model_transformer.fit(loader, 
                      validation_data=val_loader,
                      epochs=EPOCHS
                     )
```
2. Save the model to disk, which is successful:
```
model_transformer.save(os.path.join('/ecom_modeling/11-30-session', 'transformer'))
```
It does have some warnings though:
```
WARNING:tensorflow:Skipping full serialization of Keras layer TFSharedEmbeddings(
  (_feature_shapes): Dict(
    (f_47_list_seq): TensorShape([1024, None])
    (f_68_list_seq): TensorShape([1024, None])
    (item_id_list_seq): TensorShape([1024, None])
    (item_id_last): TensorShape([1024, 1])
  )
  (_feature_dtypes): Dict(
    (f_47_list_seq): tf.int32
    (f_68_list_seq): tf.int32
    (item_id_list_seq): tf.int32
    (item_id_last): tf.int32
  )
), because it is not built.
.......
[...]

PR #927, opened 2022-12-19
Fix the serialization of `SequenceSummary` block

sararb:
Saving a transformer-based model with `SequenceSummary` as a post block was throwing an error: 
```
E       NotImplementedError: 
E       Layer SequenceMean has arguments ['self', 'initializer_range']
E       in `__init__` and therefore must override `get_config()`.
E       
E       Example:
```

### Goals :soccer:
- Add a `get_config` method to the `SequenceSummary` block

### Testing Details :mag:
- Update the test `test_transformer_encoder_with_post` to check the serialization of the transformer block defined with a `SequenceSummary` post block.

### DIFF
--- a/merlin/models/tf/transformers/transforms.py
+++ b/merlin/models/tf/transformers/transforms.py
@@ -193,6 +193,10 @@ class SequenceSummary(TFSequenceSummary):
         config = SimpleNamespace(summary_type=summary)
         super().__init__(config, initializer_range=initializer_range, **kwargs)
 
+    def get_config(self):
+        config = super().get_config()
+        return config
+
 
 @Block.registry.register("sequence_last")
 @tf.keras.utils.register_keras_serializable(package="merlin.models")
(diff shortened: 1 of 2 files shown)


rnyak:
fixes the model saving error in https://github.com/NVIDIA-Merlin/models/issues/908 due to SequenceSummary as a post block.

rnyak (approved):

rnyak:
rerun tests

sararb:
rerun tests

sararb:
rerun tests

merlin/models/tf/transformers/transforms.py
```diff
@@ -193,6 +193,10 @@ def __init__(self, summary: str, initializer_range: float = 0.02, **kwargs):
         config = SimpleNamespace(summary_type=summary)
         super().__init__(config, initializer_range=initializer_range, **kwargs)
 
+    def get_config(self):
+        config = super().get_config()
```
oliverholworthy, line 197: we may also need to also save the `summary` argument from the `__init__` in this config, and implement `from_config` to restore the layer to the same state correctly

sararb: You're right, I have updated the `SequenceSummary` to make it more flexible. 
The reason I didn't add `from_config` is that `SequenceSummary` was used as a base class of the transform blocks `SequenceLast`, `SequenceFirst`, `SequenceMean`, and `SequenceClsIndex` where the `summary` parameter was set to a default value for each block.

PR #929, opened 2022-12-19
Remove saving of Merlin Schema temporarily

oliverholworthy:
Remove saving of Merlin Schema temporarily. While we work on getting pickle serialization working with the Merlin Schema saving.


oliverholworthy:
Replaced by #930

PR #930, opened 2022-12-19
Enable model serialization with pickle

### LINKED ISSUE #928: [Task] Enable pickle serialzation of Models with Merlin Schema Saved alongside
### Description

Enable pickle serialzation of Models with Merlin Schema Saved alongside.

Part of https://github.com/NVIDIA-Merlin/models/issues/669

**Additional context**

Noticed on this PR https://github.com/NVIDIA-Merlin/Merlin/pull/770

We currently get a `FileNotFoundError` when tryting to pickle a model.

This is because [keras uses the tf.io in-memory filesystem which I think is a tensorflow io feature (I've not seen ram:// used outside of this context at least) when trying to pickle a model](https://github.com/keras-team/keras/blob/v2.10.0/keras/saving/pickle_utils.py#L62). And to use this RAM filesystem, you can't use regular python file io functionality it seems. We'll need to use tf.io functions to write things into this in-memory filesystem.

This looks like it's [been replaced in TF 2.11 by a regular regular temporary directory on disk](https://github.com/keras-team/keras/commit/2ed044d06d0ae552477672aa8b778f8edafb52f1). However, since we currently aim to support versions back to TF 2.8, will need to handle both cases now.

oliverholworthy:
Fixes #928

### Goals :soccer:

Enable serializing model with pickle.

- Handles case where a path starting with `ram://` is used to save the model for the purposes of serializing the model with pickle.
  - In this case we don't need to save the Merlin schema, because the information about the schema is present in the model object.

### Testing Details :mag:

- Adds test for serializing model with pickle

### DIFF
--- a/merlin/models/io.py
+++ b/merlin/models/io.py
@@ -31,6 +31,16 @@ def save_merlin_metadata(
     output_schema: Optional[Schema],
 ) -> None:
     """Saves data to Merlin Metadata Directory."""
+
+    # When we're asked to save to a path starting with 'ram://'
+    # The use-case is for Keras 2.10 and below to serizlize
+    # the model with pickle.
+    # In this case we don't need to save the Merlin Schema.
+    # If we wanted to, we'd need to use the tf.io.gfile API
+    # https://github.com/keras-team/keras/blob/v2.10.0/keras/saving/pickle_utils.py#L62
+    if isinstance(export_path, str) and export_path.startswith("ram://"):
+        return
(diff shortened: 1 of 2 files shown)


oliverholworthy:
rerun tests

jperez999:
rerun tests

oliverholworthy:
rerun tests

jperez999 (approved):

PR #931, opened 2022-12-20
Set upper-bound on numpy version in tox.ini

oliverholworthy:
Set upper-bound on numpy version in tox.ini. numpy 1.24.0 is incompatible with the latest version of `numba`.

Fixes `SystemError: initialization of _internal failed without raising an exception` from an import `from numba import cuda`

The reason we're getting `numpy==1.24.0` is because it's installed in the Jenkins envionment and we have `sitepackages=true` in our tox config which means we're inheriting the version from the oustide environment.

## Follow-up

We should follow-up with an investigation into what packages are requires from `sitepackages` and whether we're able to set this to false by installing all required packages in the tox env or somehow specify the specific packages that we want to inherit from the system env instead of everything

### DIFF
--- a/merlin/models/tf/loader.py
+++ b/merlin/models/tf/loader.py
@@ -335,10 +335,6 @@ class Loader(merlin.dataloader.tensorflow.Loader):
         self.sparse_as_dense = sparse_as_dense
 
     @property
-    def input_schema(self) -> Schema:
-        return self.dataset.schema
-
-    @property
     def output_schema(self) -> Schema:
         schema = self.input_schema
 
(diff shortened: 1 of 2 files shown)


merlin/models/tf/loader.py
```diff
@@ -334,10 +334,6 @@ def __init__(
         self.sparse_max = sparse_max or {}
         self.sparse_as_dense = sparse_as_dense
 
-    @property
-    def input_schema(self) -> Schema:
```
oliverholworthy, line 338: This is unrelated to this PR, but needs to be removed because the datloader now has the input_schema property with a setter in https://github.com/NVIDIA-Merlin/dataloader/pull/86

edknv (approved):

PR #932, opened 2022-12-20
add clarification and example usage in contrastive class docstrings

rnyak:


merlin/models/tf/outputs/contrastive.py
```diff
@@ -84,6 +84,20 @@ class ContrastiveOutput(ModelOutput):
...
+    which is treated as a `kwargs` arg below.
+
+    Example usage::
+        outputs=mm.ContrastiveOutput(
+            to_call=DotProduct(),
+            negative_samplers="in-batch",
+            schema=schema.select_by_tag(Tags.ITEM_ID),
+            post=post_logits,
+            logits_temperature = 0.2,
+        )
```
sararb, line 100: We can maybe remove the argument `post=post_logits` from the example as it is not defined and might be confusing to the user. 
We could also add an example where the `schema` parameter is not needed (this is the case where the target schema is passed as the `to_call` arg, for example)
```
outputs = mm.ContrastiveOutput(
    to_call = schema.select_by_tag(Tags.ITEM_ID),
    negative_samplers="in-batch",
    logits_temperature = 0.2,
)
```

sararb (approved):
Thank you for the PR, it looks good to me. I just left one comment regarding the example usage of `ContrastiveOutput`

merlin/models/tf/outputs/contrastive.py
rnyak: oh you are right. I forgot to remove that :)

PR #935, opened 2022-12-21
remove transform parameter from loader

edknv:
A follow-up to #845 

### Goals :soccer:
Remove `transform` parameter from `merlin.models.tf.Loader`.

The `transform` parameter was removed when we migrated the loader to use the dataloader package, in favor of dataloader's `transforms` (with an s). As a result, the `transform` parameter will not do anything even if it's used, and not removing it was a miss in the previous PR. This PR properly cleans up the unused parameter.

### Testing Details :mag:
Existing tests.

### DIFF
--- a/merlin/models/tf/loader.py
+++ b/merlin/models/tf/loader.py
@@ -198,8 +198,6 @@ class Loader(merlin.dataloader.tensorflow.Loader):
         `reader_kwargs` will be ignored
     batch_size: int
         Number of samples to yield at each iteration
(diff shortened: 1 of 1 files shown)


gabrielspmoreira:
@edknv I placed a similar PR #937  and only after saw your PR :) . Mine also updates some tests that were using `transform` and removes also unused `multi_target_as_dict`.

edknv:
Closing in favor of #937, which also should fix additional issues.

PR #937, opened 2022-12-21
Removing from Loader unused 'transform' and 'multi_target_as_dict' constructor args

gabrielspmoreira:
### Goals :soccer:
This PR removes from Loader unused 'transform' and 'multi_target_as_dict' constructor args

### Testing Details :mag:
The tests that were using `Loader(..., transform=...)` were updated to use `Loader.map()`, which is the new recommended way.


edknv:
The GPU test depends on #936.

edknv (approved):

PR #939, opened 2022-12-27
Add downstream repo tests

edknv:
### Goals :soccer:
- Add `NVTabular` and `Systems` tests.

Note that these tests run the unit tests in these downstream repositories (not Models unit tests) with changes in Models PR.

The motivation for adding these tests is to detect if changes in Models repository potentially break other downstream packages. It is possible that changes in Models could affect the tests in NVTabular and Systems because Models hosts `model.datasets`. Furthermore, changes in model output for example could break Systems.

From what I can tell, Core and Dataloader currently do not have such dependency on Models, so tests for those repositories are not added in this PR.

The unit tests in the `Merlin` repository depends on `merlin.datasets` or `merlin.models` but the tests in the `Merlin` repository are more like integration tests and are more difficult to set up to run in Github Actions.

These downstream repo tests are similar to [how they are set up in Core](https://github.com/NVIDIA-Merlin/core/blob/cd96ca5f7ec6e9d26eba1257a8b1293b6d1bd7b3/tox.ini).

### Testing Details :mag:
`nvtabular / tests` and `systems /tests` are added to Github Actions CI.


tox.ini
```diff
@@ -46,6 +46,40 @@ commands =
...
 
+[testenv:py38-nvtabular-cpu]
+passenv=GIT_COMMIT
+allowlist_externals = git
+deps =
+    -rrequirements/base.txt
+    -rrequirements/dev.txt
+commands =
+    ; the GIT_COMMIT env is the current commit of the models repo
+    git clone --depth 1 --branch {posargs:main} https://github.com/NVIDIA-Merlin/NVTabular.git nvtabular-{env:GIT_COMMIT}
```
nv-alaiacano, line 57: I think we'd want to clone/install the `GIT_COMMIT` version of the models repo here, no? We are running the NVT tests, but want _this_ commit of models. (same thing for the gpu tests below)

tox.ini
```diff
@@ -46,6 +46,40 @@ commands =
...
+
+[testenv:py38-systems-cpu]
+passenv=GIT_COMMIT
+allowlist_externals = git
+deps =
+    -rrequirements/base.txt
+    -rrequirements/dev.txt
+commands =
+    ; the GIT_COMMIT env is the current commit of the core repo
+    git clone --depth 1 --branch {posargs:main} https://github.com/NVIDIA-Merlin/systems.git systems-{env:GIT_COMMIT}
```
nv-alaiacano, line 74: Similar to the other comment - we want to install this commit of `models`, and add a line to install the `main` branch of `systems` below.

tox.ini
edknv: Isn't the `GIT_COMMIT` version of Models already cloned by Github Actions automatically to the current directory? And we install this version of Models in line 63 with `pip install .`.

I mostly copy-pasted how it's set up in [Core](https://github.com/NVIDIA-Merlin/core/blob/cd96ca5f7ec6e9d26eba1257a8b1293b6d1bd7b3/tox.ini#L34-L51) but the workflow looks like:
- Line 57: clone NVT tests.
- Line 58-62: Install dependencies. (Just noticed that we don't need line 62. I will remove that).
- Line 63: GHA already cloned the `GIT_COMMIT` version to the current directory, so we just need to run `pip install .` to install Models.
- Line 64: Run NVT tests (which will use the version of Models installed in line 63 with `pip install .`).

What do you think?

nv-alaiacano: Ah yes I think I was thrown by the existence of line 62. Yeah this makes sense to me.

nv-alaiacano (approved):
lgtm! thanks for doing this.

PR #946, opened 2023-01-09
[WIP] Add support of sample weights to retrieval models trained with contrastive task

### LINKED ISSUE #789: [FEA] Easy functionality for using class_weights and sample_weights for Retrieval models
# 🚀 Feature request

- Since Inbatch sampler converts the retrieval task to multi-class classification,  we need an easy functionality to be able to use `class_weight` and `sample_weight` args in the model.fit() for retrieval models, without users to deal with customization.
- add unit test for using `sample_weight` and `class_weight` args for retrieval tasks in [here](https://github.com/NVIDIA-Merlin/models/blob/main/tests/unit/tf/models/test_retrieval.py)
- we need examples or code snippets in the docs to showcase the functionality

This request comes from a customer.

sararb:
Fixes #789

### Goals :soccer:
- This PR adds a `post` block to the `ContrastiveOutput` instance to support different sample weights for positive and negative sampled candidates.

- [x] Add a new block `ContrastiveSampleWeight` to a `tf/experimental`  package as this functionality is specific to a customer's use case.

 
- [x] This new block supports different ways of setting the weights of positive and negative candidates.
        ==> Positive candidates' weights can be:
                  1- An additional input feature fed to the model. 
                  2- A pre-computed tensor of shape (#unique-items) containing the weight of each candidate in the catalog. 
                  3- A scalar value, fixed for all positive candidates.
        ==> Negative candidates' weights can be:
                  1- A pre-computed tensor of shape (#unique-items) containing the weight of each candidate in the catalog. 
                  2- A scalar value, fixed for all positive candidates.

- [ ] Test support of pairwise/listwise losses.

### Testing Details :mag:
- Add a unit-test `test_two_tower_v2_with_contrastive_sample_weight` that checks different ways to set the sample weights for a RetrievalModel trained with a binary cross-entropy loss.


tests/unit/tf/models/test_retrieval.py
```diff
@@ -997,3 +997,51 @@ def test_youtube_dnn_topk_evaluation(sequence_testing_data: Dataset, run_eagerly
...
+        schema=data.schema.select_by_tag(Tags.ITEM_ID),
+        negative_samplers="in-batch",
+        store_negative_ids=True,
+    )
+    model = mm.TwoTowerModelV2(query, candidate, outputs=output, schema=data.schema)
+    model.compile(run_eagerly=False, loss="binary_crossentropy")
+
+    batch = mm.sample_batch(data, batch_size=16)
+    # loss without sample weight
+    metrics1 = model.train_step(batch)
```
gabrielspmoreira, line 1038: I think calling `train_step()` twice with the same batch results in different `loss` because it does parameters optimization under-the-hood. You could change it to `test_step()` if you add support to evaluation mode in `ContrastiveSampleWeight`.

merlin/models/tf/experimental/sample_weight.py
```diff
@@ -0,0 +1,165 @@
...
+        outputs: Prediction,
+        features,
+        training=False,
+        testing=False,
+    ) -> Prediction:
+        """Update the predictions returned by the
+        ConstrastiveOutput with a 2-D sample weights
+        for negative and positive candidates.
+        """
+        if training:
```
gabrielspmoreira, line 99: Why do you apply `sample_weight` only during `training`?
During evaluation (`testing == True`) I think we should also apply the sample weights, so that `weighted_metrics` (if provided to `model.compile()`) take the sample weights into account.

tests/unit/tf/models/test_retrieval.py
```diff
@@ -997,3 +997,51 @@ def test_youtube_dnn_topk_evaluation(sequence_testing_data: Dataset, run_eagerly
...
+    # loss with sample weight
+    model.blocks[-1].post = ContrastiveSampleWeight(
+        pos_class_weight=pos_class_weight, neg_class_weight=neg_class_weight, schema=data.schema
+    )
+    metrics2 = model.train_step(batch)
```
gabrielspmoreira, line 1043: You need to call `model.compiled_metrics.reset_state()`  between consecutive `train_step()`/`test_step()`, otherwise the **loss** will be the average of the two steps.

tests/unit/tf/models/test_retrieval.py
```diff
@@ -997,3 +997,51 @@ def test_youtube_dnn_topk_evaluation(sequence_testing_data: Dataset, run_eagerly
...
+
+    output = mm.ContrastiveOutput(
+        DotProduct(),
+        post=None,
+        schema=data.schema.select_by_tag(Tags.ITEM_ID),
+        negative_samplers="in-batch",
+        store_negative_ids=True,
+    )
+    model = mm.TwoTowerModelV2(query, candidate, outputs=output, schema=data.schema)
+    model.compile(run_eagerly=False, loss="binary_crossentropy")
```
gabrielspmoreira, line 1034: I suggest adding `weighted_metrics` here, and testing later if the metrics are the same, but `weighed_metrics` are different when setting `sample_weight`.
As a reference, I do some tests like that [here](https://github.com/NVIDIA-Merlin/models/blob/a0d1d455b34b354f19eadec9536c0870abdc077a/tests/unit/tf/outputs/test_block.py#L302-L514)

tests/unit/tf/models/test_retrieval.py
sararb: You’re right, I fixed it by using `test_step` instead.

merlin/models/tf/experimental/sample_weight.py
sararb: I didn’t know about the weighted_metrics option, as usually the weights are not available at inference and the evaluation metrics should also account for this scenario. But I can see that, in some situations, applying the weights during evaluation can be useful (to take into account the importance of getting a specific set of item-ids right). Thank you for pointing out this option, I included the evaluation mode too (testing==True)

tests/unit/tf/models/test_retrieval.py
sararb: I added the call `model.compiled_metrics.reset_state()`, thanks for the tip!

tests/unit/tf/models/test_retrieval.py
sararb: I added tests of `weighted_metrics`, and observed that top-k metrics were inconsistent when passing 2-d sample weights to the `compute_metrics`. In fact, the 2-d sample weights are used to differentiate between positive and sampled negative candidates of each row. However, the MM ranking metrics are computed with respect to the position of the positive sample in the sorted top-k list. So the information needed by weighted metrics is the weight of the positive candidate only.  
The proposed solution is to extract the positive weights from the 2-d matrix (by convention the positive weights are stored in the first column of this matrix). 
So I added the function `_extract_positive_sample_weights` to `test_step` and `train_step` right before calling `compute_metrics`. Curious to have your thoughts about the proposed function?

gabrielspmoreira (approved):
Sounds good to me

PR #947, opened 2023-01-10
fix oom issue in example unit tests

rnyak:
Currently we get OOM errors from CI when running unit test of certain example notebooks like `test_usecase_incremental_training_layer_freezing.py`. to avoid I switched the order of `TF_GPU_ALLOCATOR=cuda_malloc_async` and `import tensorflow as tf` based on recommendation by Julio.


edknv (approved):
LGTM. We've been having problems with `test_usecase_incremental_training_layer_freezing.py` being flaky due to OOM errors and it looks like the changes in this PR will resolve those issues.

PR #956, opened 2023-01-19
Add multi-gpu github actions tests using horovod

edknv:
### Goals :soccer:
This PR adds a GitHub Actions workflow for multi-GPU tests.

- We currently have the workflow [horovod (cpu)](https://github.com/NVIDIA-Merlin/models/blob/51e8678b9784a1714e3a63a3596561d22db7dfb6/.github/workflows/cpu-horovod.yml) for running unit tests that depend on Horovod, but it only runs on CPU.
- We use Conda environment mainly for `cudatoolkit` because installing the necessary Cuda and MPI libraries is easier and cleaner if we use Conda. It also has the advantage of testing the actual workflow for users that do not use the Docker environment and instead opt for installing the individual libraries. Note we also use Conda for the CPU test in [horovod-cpu-environment.yml](https://github.com/NVIDIA-Merlin/models/blob/51e8678b9784a1714e3a63a3596561d22db7dfb6/requirements/horovod-cpu-environment.yml) to make installing MPI cleaner.

### Testing Details :mag:
- GitHub Actions


karlhigley (approved):

.github/workflows/multi-gpu-ci.yml
```diff
@@ -0,0 +1,37 @@
...
+  gpu-ci:
+    runs-on: 2GPU
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Install conda
+      run: |
+        mkdir -p /tmp/miniconda
```
jperez999, line 20: If we believe this is the only way, so be it. However, I would caution against this. Conda is not something we are trying to push people towards using. We are actually trying to move away from this. Are we sure this doesnt work in our environment. We had this testing in NVTabular before and those test ran https://github.com/NVIDIA-Merlin/NVTabular/blob/main/tests/unit/loader/test_tf_dataloader.py#L53. The cuda toolkit should exist in the users environment, that should be separate from installing merlin. I believe MPI should follow the same strategy. The user should have that installed. 
We really want to stay away from conda, if the user does not start by installing conda and then run all pip installs inside of that environment, the user wont have those packages. Also running in conda will be a terrible experience the moment you run into dependency problems. These are some of the reasons we are telling people to move away from that methodology. And On GPU CI running this will blow up the environment right (forcing the download of cuda toolkit which is around 1gig https://anaconda.org/anaconda/cudatoolkit/files)? We would essentially be creating this environment for every test run, right? I also see that the support for cuda toolkit stops at 11.3. Does this mean our environment will during testing will use 11.8 toolkit on the environment for cudf and 11.3 for horovod stuff? Given that we have the container with these requirements(cuda toolkit and MPI) already, do we need this? If our container does not have MPI and cuda toolkit, it might be better to add that to the container directly.

requirements/horovod-gpu-environment.yml
```diff
@@ -0,0 +1,23 @@
+name: horovod-gpu
+
+channels:
+  - nvidia
+  - conda-forge
+  - defaults
+
+dependencies:
+  - ccache
+  - cmake=3.22
```
jperez999, line 10: this environment will add 1+ gb every time it runs(on every test execution for models repo, every commit). This is probably not what we want.

tox.ini
```diff
@@ -24,7 +24,36 @@ commands =
...
     python -m pytest --cov-report term --cov merlin -rxs tests/unit/
 
+[testenv:py38-multi-gpu]
+; Runs in: Github Actions
+; Runs multi-GPU tests.
+requires = tox-conda
+passenv =
+    CUDA_HOME
+setenv =
+    TF_GPU_ALLOCATOR=cuda_malloc_async
```
jperez999, line 34: Given we only use horovod for dataloader we should be able to set this information in our environment. We could put this information in documentation instead of running this in our testing.

jperez999:
Can I get more information as to why we need to set this up in an environment that should already have both the cuda toolkit and mpi? Like I can go on the merlin containers right now and I have mpirun and the cuda toolkit is that not enought?

jperez999 (changes requested):
Can we use the versions already on the container so we dont need to pull all these artifacts for each run, which would happen on each commit. the toolkit and cudnn alone comprise of more than 1.5+ gigs.

edknv:
@jperez999 I thought we were moving in the direction of using `tox` for tests on PRs, which is the main reason I had to install all the dependencies in the tox environment. If we don't want to be using conda and if we are okay with running some tests outside tox, I'm okay with that too.

jperez999:
@edknv ok now I get it. Yes we are running in tox but anything that is available on the actual environment we can use. For instance most of our tests run cudf, but we dont install cudf to each environment. I think it works where anything not in the environment is picked up from the main environment. So I think it should be possible to run the tox and use the packages available on the container. If this is not the case based on your testing let me know... because that will mean we have a ton of other problems... 😃 Al our other tox environments are running under the same premise.

edknv:
I will close this and move it to #1009.

PR #968, opened 2023-01-31
add get_config() to  AverageEmbeddingsByWeightFeature class

rnyak:
Currently  `AverageEmbeddingsByWeightFeature` class is missing a get_config() function, so we get error when we try to use it in the embeddings layer.

issues needs to be solved as a follow up:

- this class is not reducing the 3D to 2D which is expected. 
- this class is not called when we use shared embeddings.


sararb (approved):

rnyak:
rerun tests

rnyak:
rerun tests

PR #974, opened 2023-02-07
[WIP] Introduce distributed embeddings

edknv:
Fixes # (issue)

### DIFF
--- a/merlin/models/tf/__init__.py
+++ b/merlin/models/tf/__init__.py
@@ -128,6 +128,8 @@ from merlin.models.tf.prediction_tasks.multi import PredictionTasks
 from merlin.models.tf.prediction_tasks.regression import RegressionTask
 from merlin.models.tf.prediction_tasks.retrieval import ItemRetrievalTask
 from merlin.models.utils.dependencies import is_transformers_available
+from merlin.models.tf.distributed.embedding import DistributedEmbeddingTable
+
 
 if is_transformers_available():
     from merlin.models.tf.transformers.block import (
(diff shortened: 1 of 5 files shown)


bschifferer:
The distributed embedding examples uses a custom train step functions:
https://github.com/NVIDIA-Merlin/distributed-embeddings/blob/main/examples/dlrm/main.py#L201-L215

In my understanding, distributed embedding does NOT work with keras model.fit function:
https://github.com/NVIDIA-Merlin/models/pull/974/files#diff-1e42e5c4771f01c26b3c78c545eb341590a4406b2c5af8da0491ab4b7ea51464R80

I think we need the distributed embedding team to review the PR

FDecaYed (commented):
Looks good to me.
On model fit support: 
Current code using model.fit should work since nothing in distributed-embedding conflicts with keras model fit/compile.

The reason we have custom train_step() example is for when user wants hybrid data/model parallel. The way horovod data parallel support model.fit() api is wrapping optimizer, which will break now if distributed model parallel is also used. By my understanding, merlin-model also have integration with horovod data parallel through distributedoptimizer? if so user could run into problem when they use both integration

We will support model.fit+hvd.distributedoptimizer in next release, so code here should just work. One caveat is the fix(and later version of DE) will be depend on a later horovod version.

Alternatively, I think merlin-model would need to implement custom train_step in block/model using DE? That'll be a much bigger change though.

rnyak:
The PR needs to be update based on the dataloader changes. There is a new version of DE.

FDecaYed:
@rnyak Sorry, this fell off my list.
As of now, DE already added support for model fit. So some of the problems should be gone. I'm willing to jump in and help if needed.

On the other hand, I'm not familiar with merlin models and the dataloader change you mentioned. @edknv do you know what it is and could you help bring the code up to date?

PR #979, opened 2023-02-09
Makes ColumnBasedSampleWeight serializable

gabrielspmoreira:
### Goals :soccer:
This PR fix a serialization error when using `ColumnBasedSampleWeight` in the model.

- It makes `ColumnBasedSampleWeight` a serializable layer by defining get_config() and annotating the class to be serializable by Keras.
- When `model.save()` was called the `ColumnBasedSampleWeight.call()` was being called with args `training=True` and `targets=None`, which was causing an error as `targets` should always be provided if either `training=True` or `testing=True`. I had to change the logic to check if `targets is not None` to avoid an error when model is saved.

### Testing Details :mag:
- The unit tests on multi-task learning models (using  `ColumnBasedSampleWeight`  or not) were updated to ensure they are serializable.

### DIFF
--- a/merlin/models/tf/outputs/block.py
+++ b/merlin/models/tf/outputs/block.py
@@ -14,6 +14,7 @@
 # limitations under the License.
 #
 
+import warnings
 from typing import Dict, Optional, Sequence, Set, Tuple, Union
 
 import tensorflow as tf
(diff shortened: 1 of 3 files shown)


sararb (approved):
The changes look good to me. Thanks for the PR!

PR #975, opened 2023-02-07
enhancing the next-item prediction notebook documentation

MarkMoTrin:
Fixes # (issue)

### Goals :soccer:
Made changes based on Ronay's suggestions. Removed all extra cells, nvidia-smi, and pwd references.

PR #982, opened 2023-02-13
add a new word to ignore_codespell_words

rnyak:
This [PR](https://github.com/NVIDIA-Merlin/models/pull/975) is blocked by codespell  error bcs of automatic output coming from a cell it is not something we typed, it is output of `model.fit()` automatic keras msg.. it complains about  the  word `initalizer` printed out in this automatic msg. To unblock the PR I am adding the word `initalizer` in the `ci/ignore_codespell_words.txt`.

### DIFF
--- a/ci/ignore_codespell_words.txt
+++ b/ci/ignore_codespell_words.txt
@@ -2,3 +2,4 @@ als
 coo
 childrens
 te
+initalizer


edknv:
Just throwing some ideas out there:
- Would it be better to remove the outputs from the notebook in the blocked PR, until this is fixed upstream? I understand there are benefits to having the output visible to someone who is browsing, but this might an exception?
- I don't know who has admin privileges, but it looks like this a clear case of when the admin can step in and override the blockers?
My two cents, this change probably takes a single click to be merged in, but it might take years for anyone to notice.

oliverholworthy:
> * I don't know who has admin privileges, but it looks like this a clear case of when the admin can step in and override the blockers?

I agree it would be a good case to merge despite the error if the blocked PR had some time-sensitive need to get merged. 

However, we'd still need to fix the source of the error somehow to avoid getting this on every subsequent PR that builds after it. Since the lint check runs on all files every time it runs, not only the files that have changed.

Configuring codespeell to skip notebook files would be an option. If we want to keep checking for spelling errors in notebooks then this PR #984 stripping out notebook output before running the lint checks might be a reasonable option.

rnyak:
@oliverholworthy I am closing this since your PR https://github.com/NVIDIA-Merlin/models/pull/984 is a better solution.

PR #983, opened 2023-02-14
Fixes lower train metrics when using Keras Masking (SequenceMaskRandom, SequenceMaskLast)

### LINKED ISSUE #961: [BUG] Inconsistent results between training and eval metrics when training a session-based model
### Bug description

Currently, we can obverse that there is a big difference between metrics coming from `fit()` and `model.evaluate()` in this [example notebook](https://github.com/NVIDIA-Merlin/models/blob/main/examples/usecases/transformers-next-item-prediction.ipynb).  We can see that the model.evaluate() metric results are `4X` higher than those coming from training step. In the example we are using `SequenceMaskRandom` in the training phase, i.e. in the model.fit().

To investigate if `SequenceMaskLast` is working correctly, we also trained a model using `SequenceMaskLast` as pre arg in the `model.fit()` and then we checked out the `model.evaluate()` results using train set as well but the metric results did not match.

```
model.compile(run_eagerly=False, optimizer='adam', loss="categorical_crossentropy")
model.fit(train, batch_size=512, epochs=2, pre=mm.SequenceMaskLast(schema=seq_schema, target=target))

model.evaluate(
    train,
    batch_size=512,
    pre=mm.SequenceMaskLast(schema=seq_schema, target=target),
    return_dict=True
)
```
We assume the issue might be due to two issues:

- `SequenceMaskLast` does not work as expected
-  calculation of metrics are not correct either in `model.fit()` or in `model.evaluate()` step

Please run this code [example notebook](https://github.com/NVIDIA-Merlin/models/blob/main/examples/usecases/transformers-next-item-prediction.ipynb) to repro the issue. you can replace  `SequenceMaskRandom` in the model.fit() with `SequenceMaskLast` as well.

### Expected behavior

- we want to know that all masking methods are working fine under the hood and we can get consistent results from model.fit() and model.evaluate(). 

### Environment details
merlin-tensorflow:22.12 image with the latest main branches are pulled.

gabrielspmoreira:
Fixes #961 

### Goals :soccer:
This PR fix an issue that caused metrics obtained with `model.fit()` being much lower than the ones obtained with `model.evaluate()` when Keras Masking is used.
This bug was observed when comparing training and evaluation metrics of a Transformer example (as described in #961 ), which makes usage of Keras Masking (SequenceMaskRandom, SequenceMaskLast) to select items of the sequence for training / eval.

- After investigation, I found out that the issue was been caused by the `@tf.function` decorator we had in `model.train_compute_metrics()`. After replacing a condition inside that function by `tf.cond()`, it was possible to remove the `@tf.function` decorator and fix the error when using Keras Masking (i.e., setting `predictions._keras_mask`).

### Testing Details :mag:
- Included tests in graph mode for `test_train_metrics_steps`, to double check that the logic inside `model.train_compute_metrics()` that skips steps for computing metrics continue to working in eager and graph mode.

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -978,7 +978,6 @@ class BaseModel(tf.keras.Model):
 
         return self(x, training=False)
 
-    @tf.function
     def train_compute_metrics(self, outputs: PredictionOutput, compiled_metrics: MetricsContainer):
         """Returns metrics for the outputs of this step.
 
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -986,9 +986,11 @@ class BaseModel(tf.keras.Model):
         # Compiled_metrics as an argument here because it is re-defined by `model.compile()`
         # And checking `self.compiled_metrics` inside this function results in a reference to
         # a deleted version of `compiled_metrics` if the model is re-compiled.
-        if self._should_compute_train_metrics_for_batch:
-            return self.compute_metrics(outputs, compiled_metrics)
-        return self.metrics_results()
+        return tf.cond(
+            self._should_compute_train_metrics_for_batch,
+            lambda: self.compute_metrics(outputs, compiled_metrics),
+            lambda: self.metrics_results(),
+        )
 
     def compute_metrics(
         self,
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -396,7 +396,7 @@ class BaseModel(tf.keras.Model):
             name="should_compute_train_metrics_for_batch",
             trainable=False,
             synchronization=tf.VariableSynchronization.NONE,
-            initial_value=lambda: False,
+            initial_value=lambda: True,
         )
 
         num_v1_blocks = len(self.prediction_tasks)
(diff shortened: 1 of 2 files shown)


gabrielspmoreira:
rerun tests

edknv:
The decorator was added to fix that dataloader issue. There is unsuitability with list columns in the dataloader and adding the decorator fixed it. Do we still have issues with metrics if we use both `tf.cond` and `tf.function`?

rnyak:
@gabrielspmoreira I tested the PR and now I am getting more consistent results  between model.fit() and model.evaluate().

sararb (approved):

rnyak:
rerun tests

gabrielspmoreira:
> The decorator was added to fix that dataloader issue. There is unsuitability with list columns in the dataloader and adding the decorator fixed it. Do we still have issues with metrics if we use both `tf.cond` and `tf.function`?

Hi Edward. I remember you have added some @tf.function decorator to deal with list features.
For this fix, I removed the @tf.function only from the train_compute_metrics(), which according to [git blame](https://github.com/NVIDIA-Merlin/models/blame/63eb16152859babff1b73a5820f6aabd76a79996/merlin/models/tf/models/base.py#L981) was added by myself a while ago to be able to compute train metrics each N steps, in order to speed up training. So I think that @tf.function decorator is not related to you additions to fix the dataloader issue.

edknv (approved):
Based on the discussion offline, it sounds like the CI failure is unrelated to `tf.function`. Please ignore my previous comment.

PR #991, opened 2023-02-17
Fix BroadcastToSequence to enable context features in sequential models

### LINKED ISSUE #989: [BUG] BroadcastToSequence class does not work properly to broadcast the scalar inputs to list inputs
### Bug description

When I use `post = mm.BroadcastToSequence(context_schema, seq_schema),` to broadcast a continuous scalar item as a list column, I noticed that the input block does not include my scalar item.. Looks like ` mm.BroadcastToSequence()` is not working as expected. 

Please run the code below to reproduce the issue/error:

```
from merlin.schema.tags import Tags
from merlin.io.dataset import Dataset
from merlin.datasets.synthetic import generate_data
train = generate_data("sequence-testing", num_rows=1000)
seq_schema =train.schema.select_by_name(['item_id_seq', 'categories', 'item_age_days_norm'])
context_schema = train.schema.select_by_name(['user_age'])
train.schema = train.schema.select_by_name(['item_id_seq', 'categories', 'item_age_days_norm', 'user_age'])

target_schema = train.schema.select_by_tag(Tags.ITEM_ID)
target = target_schema.column_names[0]
print(target)

item_id_name = train.schema.select_by_tag(Tags.ITEM_ID).first.properties['domain']['name']
dmodel =32
manual_dims = {
    'item_id_seq': dmodel, 
    'categories': 16
}

mlp_block = mm.MLPBlock(
                [128,dmodel],
                activation='relu',
                no_activation_last_layer=True,
)


input_block = mm.InputBlockV2(
        context_schema + seq_schema,    # if we only feed seq_schema it works but it wont consider the broadcasted feature.
        embeddings=mm.Embeddings(
            seq_schema.select_by_tag(Tags.CATEGORICAL), 
            sequence_combiner=None,
            dim=manual_dims
        ),
        post = mm.BroadcastToSequence(context_schema, seq_schema),
        aggregation=None
)

dense_block =mm.SequentialBlock(
    input_block,
    mlp_block,
    mm.XLNetBlock(d_model=dmodel, n_head=4, n_layer=2, 
                   pre=mm.ReplaceMaskedEmbeddings(),
                   post="inference_hidden_state",
                  )
)

mlp_block2 = mm.MLPBlock(
                [128,dmodel],
                activation='relu',
                no_activation_last_layer=True,
)

prediction_task= mm.CategoricalOutput(
    to_call=input_block["categorical"][item_id_name],
)
model_transformer = mm.Model(dense_block, mlp_block2, prediction_task)

model_transformer.compile(run_eagerly=False, optimizer='adam', loss="categorical_crossentropy",
              metrics=mm.TopKMetricsAggregator.default_metrics(top_ks=[4])
             )
model_transformer.fit(train, batch_size=512, epochs=1, pre=mm.SequenceMaskRandom(schema=seq_schema, target=target, masking_prob=0.3))
```
Then check the input block output with a sample batch

```
batch= mm.sample_batch(train, batch_size=100, include_targets=False, to_ragged=True)
input_block(batch).keys()
```

you can see that it does not print out the `user_age` which is the context feature that is supposed to be broadcasted. Similarly if you print `model_transformer` you will see that the model architecture does not have `user_age` feature in any layer.

PLEASE NOTE THAT IF I feed `train.schema` to input block instead of `seq_schema` I get an error. How should I make this work?

### Expected behavior
The `BroadcastToSequence` should work as expected.

### Environment details
- Merlin version: 
- Platform:
- Python version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?):

I am using `merlin-tensorflow:22.12` with the latest main branches pulled.

gabrielspmoreira:
Fixes #989

### Goals :soccer:
This PR fixes `BroadcastToSequence`, which was not working properly in some cases

- `BroadcastToSequence` was not working in graph mode because the contextual features were being expanded to match the shape of sequential features using a logic like the following snippet, which was making the first dim (batch size) of the resulting tensor fixed, while the other sequential tensors had the first dim as None (bcs the batch size might vary).
```python
sequence_length = inputs["a_sequential_feature"].row_lengths()
broadcasted_contextual_feature = tf.RaggedTensor.from_row_lengths(
    tf.repeat(inputs["a_contextual_feature"], sequence_length, axis=0), 
    sequence_length 
)
```

The solution I found to keep the first dim of the expanded context feature tensor matching the other sequential features was using `tf.ones_like()` to create a (ragged) tensor matching the sequential feature shape and then multiply by the context feature (equivalent to repeat it), as the following example

```python
first_seq_feature_name = list(seq_features_shapes.keys())[0]
non_seq_target[fname] = tf.ones_like(
                            inputs[first_seq_feature_name][..., :1]
                         ) * tf.expand_dims(inputs["a_contextual_feature"], -1)
```

This PR also includes other fixes:
- `Continuous` block now forwards masking (`supports_masking=True`)
- `BroadcastToSequence` was refactored to separate the logic that checks if sequential and context features in the input match the schema, otherwise exceptions are raised now.
- In `BroadcastToSequence.compute_mask()`  I don't call `self._broadcast()` and the logic there was made simpler: the expanded contextual feature mask should match the sequential feature mask.
- The `SequenceTargetAsInput` now returns a `tuple` instead of a `Prediction`, so that Keras can better align the `inputs` and `targets` output from `call()` and `compute_mask()` of child classes.
- In `ProcessList`, reshaping scalar features (is_list=False,is_ragged=False) to be 2D (batch size, 1), as the last dim was None in graph mode and was causing issues when concatenating

### Testing Details :mag:
Have added new tests for testing `BroadcastToSequence` usage as a `post` of `InputBlockV2`, with both categorical and continuous context (non-sequential) features and also to test it in a Transformer model trained with masked language modeling.

### DIFF
--- a/merlin/models/tf/transforms/tensor.py
+++ b/merlin/models/tf/transforms/tensor.py
@@ -84,7 +84,7 @@ class ProcessList(TabularBlock):
 
         for name, val in inputs.items():
             is_ragged = True
-            if name in self.schema:
+            if name in self.schema.column_names:
                 val_count = self.schema[name].properties.get("value_count")
                 if (
                     val_count
@@ -102,7 +102,11 @@ class ProcessList(TabularBlock):
                 ragged = val
             else:
                 # Expanding / setting last dim of non-list features to be 1D
-                if not self.schema[name].is_list and not self.schema[name].is_ragged:
+                if (
+                    name in self.schema.column_names
+                    and not self.schema[name].is_list
+                    and not self.schema[name].is_ragged
+                ):
                     val = tf.reshape(val, (-1, 1))
                 outputs[name] = val
                 continue
--- a/merlin/models/tf/transforms/tensor.py
+++ b/merlin/models/tf/transforms/tensor.py
@@ -101,6 +101,9 @@ class ProcessList(TabularBlock):
             elif isinstance(val, tf.RaggedTensor):
                 ragged = val
             else:
+                # Expanding / setting last dim of non-list features to be 1D
+                if not self.schema[name].is_list and not self.schema[name].is_ragged:
+                    val = tf.reshape(val, (-1, 1))
                 outputs[name] = val
                 continue
 
--- a/merlin/models/tf/inputs/continuous.py
+++ b/merlin/models/tf/inputs/continuous.py
@@ -47,6 +47,7 @@ class Continuous(Filter):
     ):
         if inputs is None:
             inputs = Tags.CONTINUOUS
+        self.supports_masking = True
         super().__init__(inputs, **kwargs)
 
 
--- a/merlin/models/tf/transforms/features.py
+++ b/merlin/models/tf/transforms/features.py
@@ -812,7 +812,7 @@ def reshape_categorical_input_tensor_for_encoding(
 
 
 @tf.keras.utils.register_keras_serializable(package="merlin.models")
-class BroadcastToSequence(tf.keras.layers.Layer):
+class BroadcastToSequence(Block):
     """Broadcast context features to match the timesteps of sequence features.
 
     This layer supports mask propagation. If the sequence features have a mask. The
@@ -884,6 +884,8 @@ class BroadcastToSequence(tf.keras.layers.Layer):
     @tf.function
     def _broadcast(self, inputs, target):
         seq_features_shapes, sequence_length = self._get_seq_features_shapes(inputs)
+        first_seq_feature_name = list(seq_features_shapes.keys())[0]
+        first_seq_feature_value = inputs[first_seq_feature_name]
         if len(seq_features_shapes) > 0:
             non_seq_features = set(inputs.keys()).difference(set(seq_features_shapes.keys()))
             non_seq_target = {}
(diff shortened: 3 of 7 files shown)


merlin/models/tf/transforms/tensor.py
```diff
@@ -101,6 +101,13 @@ def call(self, inputs: TabularData, **kwargs) -> TabularData:
             elif isinstance(val, tf.RaggedTensor):
                 ragged = val
             else:
+                # Expanding / setting last dim of non-list features to be 1D
```
oliverholworthy, line 104: is this relevant to `ProcessList`. intuitively `ProcessList` sounds like it might only be transforming list features

gabrielspmoreira: That is a good point @oliverholworthy . The `ProcessList` is currently used at [core places](https://github.com/NVIDIA-Merlin/models/search?q=process_list) to ensure the features are in good shape for models. When the change making dataloader outputs scalars as 1D happens, we will also need this fix that makes scalars 2D (batch size, 1) for models.
What if we rename `ProcessList` to `PrepareFeatures` and have it as a generic block that works as a translation layer between dataloader and models (used in the same [places](https://github.com/NVIDIA-Merlin/models/search?q=process_list) `ProcessList` is currently used)?

marcromeyn (approved):

merlin/models/tf/transforms/tensor.py
oliverholworthy: The rename sounds like a reasonable thing to do and better matches it's purpose. Can be in another PR if preferred

oliverholworthy (approved):

merlin/models/tf/transforms/tensor.py
gabrielspmoreira: @oliverholworthy I have created another separate PR just for renaming `ProcessList` to `PrepareFeatures`: PR #992

Issue #819, opened 2022-10-24
[Task] Add RNN-based model apis for session-based recommendation


rnyak:
### Description

Currently there is no out-of-the-box api for creating an LSTM, BiLSTM and GRU architectures for a session-based (or sequential) task.

- For LSTM , I assume we'd use tf.keras.layers.LSTM api.
- For BiLSTM shall we use [tf.keras.layers.Bidirectional](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Bidirectional)?
- For GRU, can we not port GRU4Rec model code from TF4Rec to Merlin Models?

Looks like there were two tickets before but they were closed without completion:

https://github.com/NVIDIA-Merlin/models/issues/732
https://github.com/NVIDIA-Merlin/models/issues/733

zhiruiwang:
Just a wild suggestion, maybe you can also add SAS4Rec by using tf.keras.layers.MultiHeadAttention. I assume it will be a very similar block structure once LSTM and GRU blocks are done

rnyak:
closing this since we are creating an example for  BiLSTM where we can easily  use [tf.keras.layers.Bidirectional](https://www.tensorflow.org/api_docs/python/tf/keras/layers/Bidirectional)?

PR #999, opened 2023-02-24
Updates Models to support new dataloader format for lists (__values and __offsets in dict) and scalar (1D)

gabrielspmoreira:
Fixes [#819](https://github.com/NVIDIA-Merlin/Merlin/issues/819)
It also resolves the [Capture shapes everywhere #823](https://github.com/NVIDIA-Merlin/Merlin/issues/823) for the Merlin Models library

### Goals :soccer:
This PR changes Merlin Models to support dataloader changes from [PR #101](https://github.com/NVIDIA-Merlin/dataloader/pull/101):
- List features are now represented by two dict keys with the following format: `{feature_name}__values`, `{feature_name}__offsets`, instead of a tuple
- Scalar features are now 1D tensors, instead of 2D

It also updates code from using `ColumnSchema` `"value_count"` to use the new ColumnSchema `shape` API.

Merlin Models had code in many places that converts the tuple representation into RaggedTensor, SparseTensor, or Tensor.
This PR centralizes the conversion from the dataloader representation of list features and eliminate duplicate code on lists conversion.

- Created centralized blocks to prepare scalar and list features (`PrepareFeatures`, `PrepareListFeatures`)
- Removed duplicated blocks to manage lists (`ListToDense`, `ListToRagged`, `ListToSparse`)
- Replaced usage of "properties/value_count" to use the new `shape`

*Other*
- Changed NVT preproc workflow of `examples/usecases/transformers-next-item-prediction.ipynb` to fix tagging and make it simpler

### Testing Details :mag:
- Fixed dozens of unit and integration tests to match the new data loader and shape API

### DIFF
--- a/merlin/datasets/testing/sequence_testing/schema.json
+++ b/merlin/datasets/testing/sequence_testing/schema.json
@@ -26,8 +26,7 @@
       "name": "item_age_days_norm",
       "type": "FLOAT",
       "valueCount": {
-        "min": "1",
-        "max": "4"
+        "min": "1"
       },
       "floatDomain": {
         "name": "item_age_days_norm",
@@ -37,7 +36,7 @@
       "annotation": {
         "tag": [
           "continuous",
-          "item", 
+          "item",
           "sequence"
         ]
       }
@@ -46,8 +45,7 @@
       "name": "event_hour_sin",
       "type": "FLOAT",
       "valueCount": {
-        "min": "1",
-        "max": "4"
+        "min": "1"
       },
       "floatDomain": {
         "name": "event_hour_sin",
@@ -65,8 +63,7 @@
       "name": "event_hour_cos",
       "type": "FLOAT",
       "valueCount": {
-        "min": "1",
-        "max": "4"
+        "min": "1"
       },
       "floatDomain": {
         "name": "event_hour_cos",
@@ -84,8 +81,7 @@
       "name": "event_weekday_sin",
       "type": "FLOAT",
       "valueCount": {
-        "min": "1",
-        "max": "4"
+        "min": "1"
       },
       "floatDomain": {
         "name": "event_weekday_sin",
@@ -103,8 +99,7 @@
       "name": "event_weekday_cos",
       "type": "FLOAT",
       "valueCount": {
-        "min": "1",
-        "max": "4"
+        "min": "1"
       },
       "floatDomain": {
         "name": "event_weekday_cos",
@@ -115,7 +110,6 @@
           "continuous",
           "item",
           "sequence"
-
         ]
       }
     },
@@ -123,8 +117,7 @@
       "name": "item_id_seq",
       "type": "INT",
       "valueCount": {
-        "min": "1",
-        "max": "4"
+        "min": "1"
       },
       "intDomain": {
         "name": "item_id_seq",
@@ -144,8 +137,7 @@
     {
       "name": "categories",
       "valueCount": {
-        "min": "1",
-        "max": "4"
+        "min": "1"
       },
       "type": "INT",
       "intDomain": {
--- a/merlin/models/tf/core/combinators.py
+++ b/merlin/models/tf/core/combinators.py
@@ -615,7 +615,11 @@ class ParallelBlock(TabularBlock):
     def _maybe_filter_layer_inputs_using_schema(self, name, layer, inputs):
         maybe_schema = getattr(layer, "_schema", None)
         if maybe_schema and isinstance(inputs, dict):
-            layer_inputs = {k: v for k, v in inputs.items() if k in maybe_schema.column_names}
+            layer_inputs = {
+                k: v
+                for k, v in inputs.items()
+                if k.replace("__values", "").replace("__offsets", "") in maybe_schema.column_names
+            }
         else:
             layer_inputs = inputs
 
--- a/merlin/models/tf/core/encoder.py
+++ b/merlin/models/tf/core/encoder.py
@@ -160,10 +160,11 @@ class Encoder(tf.keras.Model):
         return merlin.io.Dataset(predictions)
 
     def call(self, inputs, training=False, testing=False, targets=None, **kwargs):
+        inputs = self.prepare_features(inputs)
         return combinators.call_sequentially(
             list(self.to_call),
             inputs=inputs,
-            features=self.prepare_features(inputs),
+            features=inputs,
             targets=targets,
             training=training,
             testing=testing,
--- a/merlin/models/tf/core/encoder.py
+++ b/merlin/models/tf/core/encoder.py
@@ -167,7 +167,9 @@ class Encoder(tf.keras.Model):
         return merlin.io.Dataset(predictions)
 
     def call(self, inputs, targets=None, training=False, testing=False, **kwargs):
-        inputs, targets = self._prepare_features(inputs, targets=targets)
+        inputs = self._prepare_features(inputs, targets=targets)
+        if isinstance(inputs, tuple):
+            inputs, targets = inputs
         return combinators.call_sequentially(
             list(self.to_call),
             inputs=inputs,
--- a/merlin/models/tf/core/encoder.py
+++ b/merlin/models/tf/core/encoder.py
@@ -166,7 +166,7 @@ class Encoder(tf.keras.Model):
 
         return merlin.io.Dataset(predictions)
 
-    def call(self, inputs, targets=None, training=False, testing=False, **kwargs):
+    def call(self, inputs, *, targets=None, training=False, testing=False, **kwargs):
         inputs = self._prepare_features(inputs, targets=targets)
         if isinstance(inputs, tuple):
             inputs, targets = inputs
--- a/merlin/models/tf/core/encoder.py
+++ b/merlin/models/tf/core/encoder.py
@@ -166,8 +166,8 @@ class Encoder(tf.keras.Model):
 
         return merlin.io.Dataset(predictions)
 
-    def call(self, inputs, training=False, testing=False, targets=None, **kwargs):
-        inputs = self._prepare_features(inputs)
+    def call(self, inputs, targets=None, training=False, testing=False, **kwargs):
+        inputs, targets = self._prepare_features(inputs, targets=targets)
         return combinators.call_sequentially(
             list(self.to_call),
             inputs=inputs,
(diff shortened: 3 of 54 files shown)


merlin/models/tf/inputs/base.py
```diff
@@ -208,7 +200,13 @@ def InputBlock(
             name="continuous_projection",
         )
 
-    return ParallelBlock(branches, aggregation=aggregation, post=post, is_input=True, **kwargs)
+    _pre = PrepareFeatures(schema)
```
oliverholworthy, line 203: does every model have PrepareFeatures as the first layer? if so, maybe we don't need this in the InputBlock?

oliverholworthy:
Re: issue with loader returning SparseTensor's. I think this is because in some cases we're now setting the value_count.max property (specifying the shape of ragged dimensions) which wasn't previously set. Until we remove the sparse output option in the dataloader https://github.com/NVIDIA-Merlin/dataloader/pull/103 . One intermediate step could be to update `_augment_schema`  helper function to always remove `value_count` from the properties if it's there so that ragged columns will always return the ragged representation.

merlin/models/tf/inputs/base.py
gabrielspmoreira: I changed that. As there are many blocks that don't use `InputBlockV2` and `InputBlock`, I moved the calling of `PrepareFeatures` to `Model`, `Encoder` and to the blocks that can be used as `pre` for `fit() / evaluate()` and the ones that can be used as a transform for `Loader.map()`

merlin/models/tf/transforms/features.py
```diff
@@ -104,23 +105,282 @@ class ToDense(FeaturesTensorTypeConversion):
...
+                                    f"List feature {name} does not have maximum"
+                                    "length set in the schema ({col_schema_shape})"
+                                )
+                            seq_length = int(col_schema_shape.dims[1].max)
+
+                        output_shapes[name] = tf.TensorShape([batch_size, seq_length, 1])
+
+                    else:
+                        if len(input_shapes[name]) == 2:
+                            output_shapes[name] = input_shapes[name] + [1]
```
karlhigley, line 245: Based on the error message from the test failures about how only tuples and not lists can be concatenated onto tuples, I think this might need to be:
```suggestion
                            output_shapes[name] = input_shapes[name] + (1,)
```

gabrielspmoreira: Hey @karlhigley . Thanks for the review and suggestion. Typically we can concatenate `tf.TensorShape()` (`input_shapes[name]`) with `list`. I think CI is using the dataloader from main branch, which still returns tuple for list features.
I suspect for some reason the changes I did in [tox.ini](https://github.com/NVIDIA-Merlin/models/pull/999/commits/83f195334578e8fb5e0c7d0c04fe50882d7ca118) are not being able to install the dataloader from [PR #101](https://github.com/NVIDIA-Merlin/dataloader/pull/101).

karlhigley: Curious: Does `TensorShape` concatenate with tuples too, or only lists?

karlhigley: It looks like either one of these should work with `TensorShape` but tuples will only concat other tuples, so I think it's safe to switch over to a tuple here:
```
>>> shape = tf.TensorShape((1,2,3))
>>> shape + [4]
TensorShape([1, 2, 3, 4])
>>> shape + (4,)
TensorShape([1, 2, 3, 4])

>>> shape = (1,2,3)
>>> shape + [4]
  File "<stdin>", line 1, in <module>
TypeError: can only concatenate tuple (not "list") to tuple
>>> shape + (4,)
(1, 2, 3, 4)
```

gabrielspmoreira: Sure @karlhigley . Changed it to tuple as you suggested

tox.ini
```diff
@@ -38,8 +41,10 @@ setenv =
 sitepackages=true
 commands =
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git@{posargs:main}
-    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@{posargs:main}
+    #python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@{posargs:main}
+    python -m pip install --upgrade git+https://github.com/bschifferer/dataloader.git@change_output
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/nvtabular.git@{posargs:main}
+    python -m pip install --upgrade git+https://github.com/bschifferer/dataloader.git@change_output
     horovodrun -np 2 sh examples/usecases/multi-gpu/hvd_wrapper.sh pytest -m horovod -rxs tests/unit
```
edknv, line 48: I noticed that the multi-gpu test is failing. Can we try `python -m pytest` here? i.e., `horovodrun -np 2 sh examples/usecases/multi-gpu/hvd_wrapper.sh python -m pytest -m horovod -rxs tests/unit`

gabrielspmoreira: Thanks @edknv . Let me try that

karlhigley: Looks like this hits version conflicts on Tensorflow. Maybe we have different versions installed on the CI container vs in the Tox environment?

edknv: PR for bumping the TF version: https://github.com/NVIDIA-Merlin/Merlin/pull/849.

@gabrielspmoreira In the meantime, we could also try removing:
```
deps =
    -rrequirements/test.txt
```
on lines 32-33 to unblock this. Those line will install TF 2.10 in the tox environment, while the ci-runner still has TF 2.9, which seems to be leading to conflict. Maybe worth a try?

gabrielspmoreira: Just pushed that change to try

gabrielspmoreira: Yeah @karlhigley . We still have the horovod multi-GPUs tests failing for not using the modified dataloader

edknv: Sorry I think we should still have `python -m pytest` in the GPU test (AND remove `-rrequirements/test.txt` at the same time). :crossed_fingers:

merlin/datasets/testing/sequence_testing/schema.json
```diff
@@ -26,8 +26,7 @@
       "name": "item_age_days_norm",
       "type": "FLOAT",
       "valueCount": {
-        "min": "1",
-        "max": "4"
```
oliverholworthy, line 30: Why do we need to remove the max value count from this schema and others?

merlin/models/tf/core/combinators.py
```diff
@@ -615,7 +615,11 @@ def build(self, input_shape):
     def _maybe_filter_layer_inputs_using_schema(self, name, layer, inputs):
         maybe_schema = getattr(layer, "_schema", None)
         if maybe_schema and isinstance(inputs, dict):
-            layer_inputs = {k: v for k, v in inputs.items() if k in maybe_schema.column_names}
+            layer_inputs = {
+                k: v
+                for k, v in inputs.items()
+                if k.replace("__values", "").replace("__offsets", "") in maybe_schema.column_names
```
oliverholworthy, line 621: what situation is this change required for. If we're handling the ragged inputs in the first layer how are the values/offsets making their way through to a ParallelBlock?

merlin/models/tf/core/encoder.py
```diff
@@ -159,11 +166,14 @@ def batch_predict(
 
         return merlin.io.Dataset(predictions)
 
-    def call(self, inputs, training=False, testing=False, targets=None, **kwargs):
+    def call(self, inputs, targets=None, training=False, testing=False, **kwargs):
```
oliverholworthy, line 169: is the re-ordering of the params a functional change. e.g. are there any places where these are being passed in as positional params instead of keyword. One option could be to enforce this by adding a `*` after inputs. That way we can be sure that the order of the keyword arguments don't matter. 

```python
def call(self, inputs, *, targets=None, training=False, testing=False, **kwargs):
```

merlin/models/tf/models/base.py
```diff
@@ -142,18 +142,13 @@ def get_output_schema(export_path: str) -> Schema:
...
-            if not is_ragged:
-                properties["value_count"] = {"min": shape[1], "max": shape[1]}
+        if shape.rank > 1 and (shape[1] is not None and shape[1] > 1):
             col_schema = ColumnSchema(
                 output_name,
                 dtype=output_spec.dtype.as_numpy_dtype,
-                is_list=True,
-                is_ragged=is_ragged,
-                properties=properties,
+                dims=((0, None), (shape[1], shape[1])),
```
oliverholworthy, line 149: This can be simplified to `dims=(None, shape[1])`. Equivalent to what is already here.

merlin/datasets/testing/sequence_testing/schema.json
gabrielspmoreira: The issue is that when min and max are provided and are different, the dataloader returns a tf.SparseTensor, but now MM can only take either tf.Tensor or our convension for ragged tensors (__values and __offsets). So I remove the max to enforce it to return the ragged tensors.
I know you have already a [draft PR](https://github.com/NVIDIA-Merlin/dataloader/pull/103) that will make it return ragged tensors in such cases. As soon as it is merged we can add back the `value_count.max` from the test dataset schemas if needed.
Does that make sense?

merlin/models/tf/core/combinators.py
gabrielspmoreira: You are right. In general we use `PrepareFeatures` in the `Model` right at the beginning to ensure that lists are converted from our ragged tensor convension (__values and __offsets) to `tf.RaggedTensor`.
But for `RetrievalModelV2` we cannot `PrepareFeatures` in `Model`, because in that case the `Encoders` (which are `tf.keras.Model`) would receive `tf.RaggedTensor` as inputs and that would cause problems when the query encoder is saved to be served on Triton.
That is why on `RetrievalModelV2` constructor we set `prep_features=False` and let the `Encoder` class to `prep_features=True`. As the `RetrievalModelV2` is a `ParallelBlock` of `Encoder` that is when it would receive the ragged convension (`__values` and `__offsets`) and this change is to ensure that the ragged representation is cascaded to the `Encoder` block

merlin/models/tf/models/base.py
gabrielspmoreira: Done

merlin/models/tf/core/encoder.py
gabrielspmoreira: Done

bschifferer:
rerun tests

PR #1002, opened 2023-02-27
Save and load LightFM model and schemas

edknv:
Partially addresses https://github.com/NVIDIA-Merlin/Merlin/issues/825.

### Goals :soccer:
Enable LightFM to save and load the model and schemas.

- Adds `save()` that saves the LightFM object as a pickle. It also outputs input and output schemas along with the model.
- Adds `load()` that can load the saved model and schema back into memory.

### Testing Details :mag:
Added unit tests to `tests/unit/lightfm/test_lightfm.py`.

### DIFF
--- a/merlin/models/io.py
+++ b/merlin/models/io.py
@@ -17,7 +17,6 @@ import os
 import pathlib
 from typing import Optional, Union
 
-from merlin.models.api import MerlinModel
 from merlin.models.utils.schema_utils import schema_to_tensorflow_metadata_json
 from merlin.schema import Schema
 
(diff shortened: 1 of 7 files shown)


merlin/models/io.py
```diff
@@ -26,7 +25,6 @@
 
 def save_merlin_metadata(
     export_path: Union[str, os.PathLike],
-    model: MerlinModel,
```
edknv, line 29: Removed because it looks like `model` is not used.

marcromeyn (approved):

PR #1005, opened 2023-02-28
Install gpu requirements in tox config

oliverholworthy:
### Goals :soccer:

Run GPU tests with cudf and transformers

Adding gpu requirements to gpu.txt and installing all requirements inside `py38-gpu` tox env

### Testing Details :mag:

- [ ] Check that GPU test runner doesn't skip any tests


oliverholworthy:
This was an alternative to setting sitepackages=True in our tox config. We may wish to revisit this option in future if sitepackages turns out to be an issue. 

Closing this in favour of setting sitepackages in #1011

PR #1008, opened 2023-02-28
New design of the transformer API to support causal and masked pre-training approach

### LINKED ISSUE #918: Support masking for causal language modeling
The current API does not generate the `_keras_mask` for causal language modeling.  This implementation has two constraints: 
- During training and evaluation: The loss is computed for all positions (even the padded index)
- During inference, the model returns the scores for all positions (actual input candidates + padded position). While we usually expect the model to return the score only for one position (the last non-padded one).
- Simplify the masking API of transformer-based models by adding `masking` string argument to the TransformerBlock and hide the custom `pre` blocks such as: `SequenceMaskLastInference` and `ReplaceMaskedEmbeddings`

_Originally posted by @gabrielspmoreira in https://github.com/NVIDIA-Merlin/models/pull/909#pullrequestreview-1214326395_
      
In addition,  for CLM, we should be able to train with `SequencePredictNext` and evaluate with `SequencePredictLast`.

sararb:
Fixes #918 

- This work is required for the GTC tutorial 2023 so It is built on top of the 23.02 release branch to be able to merge it with the main branch of the tutorial image (the image uses the latest released container)

- The new API for defining a transformer-based model is defined as follows:

```
    transformer_input_dim = 48
    transformer_block = BertBlock(d_model=transformer_input_dim, n_head=8, n_layer=2)
    model = mm.Model(
        mm.InputBlockV2(...),
        transformer_block,
        mm.CategoricalOutput(...),
    )
    seq_mask_random = mm.SequenceMaskRandom(
        schema=seq_schema, target=target, masking_prob=0.3, transformer=transformer_block
    )
    model.compile(...)
    model.fit(..., pre=seq_mask_random)
```

* I created this doc for more details about the Merlin Transformer API and the proposed changes: https://docs.google.com/document/d/1auNJCOPFjyVxlhx4eUUthpMLm-rva0jC-lo0ChSZWzI/edit#

### Goals :soccer:

This PR aims to address the following limitations in the current Transformer API:

- [ ] 1. **Inference for CausalLM is not supported**: the model returns scores for all positions in the input sequence but we are only interested on the last position at inference time.
- [ ] 2. **CLM only supports the transform SequencePredictNext**: We cannot train or evaluate a CLM model on the last item of the sequence only (i.e using SequencePredictLast)
- [ ] 3. **Specialized and complex API**: the support of masking and inference is done via specialized Merlin Blocks (`ReplaceMaskedEmbeddings`, `SequenceMaskLastInference`). So the user should be familiar with all these custom blocks to correctly define and train a transformer-based model with CLM or MLM approach.
- [ ] 4. **Output of the model is a padded dense tensor of scores**: As the HuggingFace transformer layer is requiring a dense input. We are converting the ragged inputs to dense (by 0-padding to the maximum length seen in the given batch) right before calling the HF layer. The ops that follow this block are then applied to the padded dense tensor. This means that we compute the logit scores for all positions (even the padded ones) and which can be costly (such in the weight tying multiplication between the hidden representation all items embeddings).

- [x] Add Inference support to a transformer-based model trained with CLM: i.e. select the prediction score of the last non-padded position.
- [x] Support evaluating a transformer-based model trained with CLM (trained with SequencePredictNext) on the last item in the sequence (i.e using SequencePredictLast).
- [x] Simplify the transformer API by abstracting the definition of the `masking_pre` and `masking_post` blocks using the `pre` transform set in the `fit()` method.
- [x] Optimize the predictions generation of transformer-based model: Convert the dense tensor returned by the transformer layer to a `tf.RaggedTensor` so that all the logic happening after the transformer block (MLP projections, softmax layer...) is applied only on actual positions.
- [x] Add masking support for `SequencePredictNext` and `SequencePredictLast`
- [x] Add checks to ensure the SequenceTransform used in evaluate() is aligned with the `masking_pre` used to train the transformer model. 
- [ ] Implement the following specialized blocks (needed for CLM and MLM support): 
    *  `SequenceCausalLastInference` to generate a mask indicating the last non-padded position of each input sequence at inference time. 
    * `ExtractMaskFromTargets` to move the logic defined in `ReplaceMaskedEmbeddings` and which consists of inferring the mask information based on targets. 
    * `TransformerOutputToRagged` to convert the output of the transformer layer to Ragged based on the mask information. 
- [ ] Update the example notebooks 

### Testing Details :mag:
- Update `test_transformer_with_masked_language_modeling` and `test_transformer_with_causal_language_modeling` to account for the changes. 
- Update the


merlin/models/tf/utils/tf_utils.py
```diff
@@ -469,3 +469,37 @@ def list_col_to_ragged(col: Tuple[tf.Tensor, tf.Tensor]):
...
+        Returns True if the shape of the input and mask tensors are compatible, False otherwise.
+
+    Notes
+    -----
+       The function assumes that the `inputs` tensor has one more dimension than the `mask` tensor,
+       with the extra dimension typically related to the embeddings dimension.
+    """
+    result = False
+    if type(inputs) == type(mask) and (
+        inputs.shape.as_list()[: len(inputs.shape) - 1] == mask.shape.as_list()
```
gabrielspmoreira, line 497: You might want to replace `[: len(inputs.shape) - 1]` by `[:-1]`

merlin/models/tf/models/base.py
```diff
@@ -812,60 +813,215 @@ def call_train_test(
...
+            )
+
+        # Ensuring targets and preds have the same dtype
+        target = tf.cast(target, prediction.dtype)
+
+        # Take the flat values of predictions and targets as Keras
+        # losses does not support RaggedVariantTensor on GPU:
+        prediction = prediction.flat_values
+        if isinstance(target, tf.RaggedTensor):
+            target = target.flat_values
```
gabrielspmoreira, line 971: As you are flattening the values here to 1D, is there a way to reshape the losses output back to be a RaggedTensor? Otherwise the 1D loss will not match the sample weights, that can be either 1D or 2D (ragged).

merlin/models/tf/models/base.py
```diff
@@ -1154,6 +1310,8 @@ def fit(
         if pre:
             self._reset_compile_cache()
             self.train_pre = pre
+            if isinstance(self.train_pre, SequenceTransform):
+                self.train_pre.on_train_begin()
```
gabrielspmoreira, line 1314: I think that using methods named like callback (`on_train_begin()` and `on_test_begin()`) might confuse the user, as it might think that a callback event is being called here. And they might think that such methods for `train_end` or `test_end` might also be available...
I would suggest renaming these methods to something like `configure_for_train()`,  `configure_for_test()`

gabrielspmoreira (commented):
Sara. Does our new Transformer API already support dense tensors for inputs and targets instead of RaggedTensor?
The dataloader provides dense tensors for sequential features in some cases (as summarized in this [ADR](https://docs.google.com/document/d/1PTyWofAtFQhyTp-l0v6xUYzMTXWMaO_b3G6sUbCrKJE/edit#heading=h.keelfimvyige)):
- In the current dataloader API, if `value_count.max is not None and is_ragged == False` 
- In the future dataloader API, if `is_ragged == False`

sararb:
> Sara. Does our new Transformer API already support dense tensors for inputs and targets instead of RaggedTensor? The dataloader provides dense tensors for sequential features in some cases (as summarized in this [ADR](https://docs.google.com/document/d/1PTyWofAtFQhyTp-l0v6xUYzMTXWMaO_b3G6sUbCrKJE/edit#heading=h.keelfimvyige)):
> 
> * In the current dataloader API, if `value_count.max is not None and is_ragged == False`
> * In the future dataloader API, if `is_ragged == False`

Thank you for checking out the PR!  This PR only addresses how to support the different masking approaches in the TransformerBlock, but we still need to work on extending the SequenceTransforms to support dense tensors as inputs (as mentioned in this [ADR](https://docs.google.com/document/d/1KcIDzEFjz-Bp4Y-s80__teervpt56yMvihphSXyXWaw/edit)).

sararb:
closing as this was a placeholder for the tutorial image.

PR #1009, opened 2023-03-01
Add multi-gpu github action tests using horovod

edknv:
### Goals :soccer:
This PR adds a GitHub Actions workflow for multi-GPU tests.

- This is an alternative implementation of #956. Based on the feedback in #956, we should favor using packages that are already installed in the CI container, if they are not available in the tox environment.

### Testing Details :mag:
GitHub Actions workflow:  `horovod (2GPU) / gpu-ci`


.github/workflows/multi-gpu-ci.yml
```diff
@@ -0,0 +1,30 @@
...
+    - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+    - name: Run tests
+      run: |
+        ref_type=${{ github.ref_type }}
+        branch=main
+        if [[ $ref_type == "tag"* ]]
+        then
+          git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/release*:refs/remotes/origin/release*
```
karlhigley, line 27: Just curious, asking for my own edification: what's different in `protocol.version=2`?

karlhigley (approved):

.github/workflows/multi-gpu-ci.yml
oliverholworthy: this was copied from other places that have this pattern. now available in [the branch-name action here](https://github.com/NVIDIA-Merlin/.github/tree/main/actions/branch-name). The reason for the `protocol.version=2` was to match what [the GitHub actions/checkout](https://github.com/actions/checkout/blob/v3.3.0/src/git-command-manager.ts#L205) does in the fetch. It may no longer be strictly necessary since the default protocol is v2 with recent versions of git.

karlhigley: It looks like v2 enables more commands to be sent over a single connection, among other things: https://git-scm.com/docs/protocol-v2

PR #1011, opened 2023-03-01
Use globally insatlled installed packages for GPU tests

edknv:
For GPU tests, we should use globally installed packages as we don't install cudf, cupy, etc. in the test environment. By setting `sitepackages=true` in tox, tox will use the packages in the CI container if not available in the environment. We had `sitepackages=true` before but it got dropped at some point. This PR restores `sitepackages=true` in the tox test `py38-gpu`.

### DIFF
--- a/tests/unit/lightfm/test_lightfm.py
+++ b/tests/unit/lightfm/test_lightfm.py
@@ -62,7 +62,7 @@ def test_reload_no_target_column(tmpdir):
     model = LightFM(learning_rate=0.05, loss="warp", epochs=10)
     model.fit(train)
 
-    _ = model.evaluate(valid)
+    _ = model.evaluate(valid, k=10)
 
     model_dir = Path(tmpdir) / "lightfm_model"
 
@@ -88,7 +88,7 @@ def test_reload_with_target_column(tmpdir):
     model = LightFM(learning_rate=0.05, loss="warp", epochs=10)
     model.fit(train)
 
-    _ = model.evaluate(valid)
+    _ = model.evaluate(valid, k=10)
 
     model_dir = Path(tmpdir) / "lightfm_model"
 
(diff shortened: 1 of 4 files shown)


tox.ini
```diff
@@ -15,9 +15,9 @@ commands =
 deps =
     -rrequirements/test.txt
 setenv =
+    CUDA_VISIBLE_DEVICES=0
```
edknv, line 18: Without setting `CUDA_VISIBLE_DEVICES=0`, we get an error:
```
tensorflow.python.framework.errors_impl.InvalidArgumentError: Device ordinals must be set for all virtual devices or none. But the device_ordinal is specified for 1 while previous devices didn't have any set.
```
We didn't have to set this with TF 2.9. Is it specific to TF 2.10? Is there a better way to handle this?

tests/unit/lightfm/test_lightfm.py
```diff
@@ -62,8 +62,6 @@ def test_reload_no_target_column(tmpdir):
     model = LightFM(learning_rate=0.05, loss="warp", epochs=10)
     model.fit(train)
 
-    _ = model.evaluate(valid)
-
```
edknv, line 66: Removed because this is unnecessary for these tests, and `evaluate()` is already tested in one of the other unit tests.

marcromeyn (approved):

tox.ini
oliverholworthy: If would like to run these tests with only 1 GPU, we can also change the runner we use in the workflow config. Currently set to `2GPU`

https://github.com/NVIDIA-Merlin/models/blob/e2276f44e77390b55462e2bd83b6f10f212d9aec/.github/workflows/gpu-ci.yml#L15

PR #1016, opened 2023-03-10
Tensorflow 2.11 support

edknv:
### Goals :soccer:
Support Tensorflow 2.11

- Tensorflow 2.11 has introduced a new set of Keras optimizers and moved the old Keras optimizers into the namespace `tf.keras.optimizers.legacy`. Most of the changes in this PR involve checking the Tensorflow version and using the correct namespace for the version.
- `MultiOptimizer` only works with legacy optimizers, and it does not work with newly introduced Keras optimizers. This PR opted to throw an error if the user tries to use it with the new optimizers. Supporting new optimizers in MultiOptimizer is left as future work.
- There was another change in TF 2.11 with saving formet. See #1040.

### Testing Details :mag:
Needs horovod 0.27. Related: https://github.com/NVIDIA-Merlin/Merlin/pull/883
For GPU test, manually built docker image with https://github.com/NVIDIA-Merlin/Merlin/pull/883 and ran tests.

### DIFF
--- a/merlin/models/tf/blocks/optimizer.py
+++ b/merlin/models/tf/blocks/optimizer.py
@@ -15,6 +15,7 @@
 #
 
 import collections
+import importlib
 import warnings
 from dataclasses import dataclass
 from typing import Callable, List, Optional, Sequence, Tuple, Union
@@ -31,6 +32,12 @@ Tensor = Union[tf.Tensor, tf.SparseTensor, tf.RaggedTensor]
 FloatTensorLike = Union[tf.Tensor, float, np.float16, np.float32, np.float64]
 
 
+if importlib.util.find_spec("tensorflow.keras.optimizers.legacy") is not None:
+    keras_optimizers = tf.keras.optimizers.legacy
+else:
+    keras_optimizers = tf.keras.optimizers
+
+
 @dataclass
 class OptimizerBlocks:
     """dataclass for a pair of optimizer and blocks that the optimizer should apply to.
@@ -39,7 +46,7 @@ class OptimizerBlocks:
         ml.OptimizerBlocks("adam", item_tower)
     """
 
-    optimizer: Union[str, tf.keras.optimizers.Optimizer]
+    optimizer: Union[str, keras_optimizers.Optimizer]
     blocks: Sequence[Block]
 
     def get_config(self):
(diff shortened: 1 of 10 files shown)


tests/unit/tf/examples/test_usecase_incremental_training_layer_freezing.py
```diff
@@ -14,6 +15,10 @@
     execute=False,
 )
 @pytest.mark.notebook
+@pytest.mark.skipif(
+    version.parse(tf.__version__) < version.parse("2.9.0"),
+    reason="tf.keras.optimizers.legacy is not available in TF <= 2.8",
+)
```
edknv, line 21: Since we need to use `legacy` optimizers in MultiOptimizer, I updated the incremental training notebook to use `tf.keras.optimizers.legacy` optimizers. And since this namespace was not available in TF 2.8, I opted to skip testing for TF 2.8, effectively dropping support for TF 2.8 for this notebook. Open questions: Is there a better way to handle this? It's likely not feasible to support all previous versions forever. Do we need an SLA or something for when to drop support for old dependency versions?

merlin/models/tf/blocks/optimizer.py
```diff
@@ -289,18 +316,26 @@ def weights(self) -> List[tf.Variable]:
...
         weights = []
         for optimizer_blocks in self.optimizers_and_blocks:
-            weights += optimizer_blocks.optimizer.weights
+            optimizer = optimizer_blocks.optimizer
+            if hasattr(optimizer, "weights"):  # Tensorflow < 2.11
+                weights += optimizer_blocks.optimizer.weights
+            elif hasattr(optimizer, "variables") and callable(
+                optimizer.variables
+            ):  # Tensorflow >= 2.11
+                weights += optimizer_blocks.optimizer.variables()
```
oliverholworthy, line 325: if `variables()` was present in earlier versions of tensorflow (2.8, 2.9, 2.10) and equivalent to `.weights` ? then we might be able to simplify to be only this line?

tests/unit/tf/examples/test_usecase_incremental_training_layer_freezing.py
oliverholworthy: Maybe the last 4 releases of TensorFlow or something would be reasonable. Might depend a bit on the level of complexity in supporting  each one

oliverholworthy (approved):

PR #1017, opened 2023-03-14
Add a `pytest.ini` file to register custom markers (e.g. `notebook`)

karlhigley:
### Goals :soccer:
This should silence a bunch of repetitive warnings when the tests run.

Adds a `pytest.ini` file that registers the custom markers already in use.

### Testing Details :mag:
No functional change to the library, but should (not) show up on subsequent test runs.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+markers =
+    notebook: mark a test as testing notebooks
+    version: mark a test as testing the version number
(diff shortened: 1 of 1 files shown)


pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+markers =
```
oliverholworthy, line 2: do we need to register any of these markers as well?
https://github.com/NVIDIA-Merlin/models/blob/e7fe759c7628e3902e029b9930e6b155f478ef25/tests/conftest.py#L86-L110

oliverholworthy (approved):

PR #1019, opened 2023-03-14
Use pip version of horovod in multi-gpu test env

oliverholworthy:
Try using pip version of horovod in multi-gpu test env instead of external horovodrun


oliverholworthy:
experimented locally instead, to avoid neeting to install `horovodrun` with MPI again without `sitepackages=true` is probably possible but not strictly needed if we can use the correct version of the packages

PR #1022, opened 2023-03-16
New design of the transformer API 

### LINKED ISSUE #918: Support masking for causal language modeling

### LINKED ISSUE #1024: [FEA] Simplify the Transformer API
# 🚀 Feature request
The current API provides support for masking and inference through specialized Merlin Blocks, such as `ReplaceMaskedEmbeddings` and `SequenceMaskLastInference`. These blocks need to be explicitly set by the user in the pre/post arguments of the `TransformerBlock`. However, this adds complexity to the high-level MM API, as the user needs to be familiar with these custom blocks in order to correctly define and train a transformer-based model using CLM or MLM approaches.  

It would be helpful to have a more streamlined approach for incorporating masking and inference support into transformer models, without requiring users to have extensive knowledge of specialized Merlin Blocks.

### LINKED ISSUE #1025: [FEA] Convert the output of TransformerBlock to ragged tensor
To align with the HuggingFace transformer layer that requires dense inputs, we convert ragged inputs to dense before calling the TransformerBlock. As a result of this conversion, the outputs are also dense.

This approach can be costly because it means computing logit scores for all positions, even the padded ones. For example, this can impact performance when applying weight-tying multiplication between the hidden representation and all items' embeddings. 

It would be helpful to convert the output of the transformer block to a ragged format, which would eliminate the need for padding and avoid unnecessary computation.

### LINKED ISSUE #1026: [FEA] Add masking support to SequencePredictNext, SequencePredictLast, and SequencePredictRandom
# 🚀 Feature request
The current sequence transforms - SequencePredictNext, SequencePredictLast, and SequencePredictRandom - don't generate _keras_mask information for inputs and targets. This is because these transforms extract the targets from the input sequence and return new truncated inputs that don't contain the target information. 

However, the Model output API uses mask information to adjust the shape of predictions and targets. To ensure that the Model output API can work with all supported sequence transforms, we should add masking support to all sequence transforms and rely on the Model output API to align the model's predictions and targets accordingly. 

This would ensure that the model output API can handle a range of sequence transforms and produce correct outputs, regardless of the specific transform used.

sararb:
Fixes #918 
Fixes #1024 
Fixes #1025
Fixes #1026

- This work was necessary for the GTC tutorial 2023  and is fixing issues related to Causal LM and at the same time simplifying the high-level transformer API. 

- The new API for defining a transformer-based model is defined as follows:

```
    transformer_input_dim = 48
    transformer_block = BertBlock(d_model=transformer_input_dim, n_head=8, n_layer=2)
    model = mm.Model(
        mm.InputBlockV2(...),
        transformer_block,
        mm.CategoricalOutput(...),
    )
    seq_mask_random = mm.SequenceMaskRandom(
        schema=seq_schema, target=target, masking_prob=0.3, transformer=transformer_block
    )
    model.compile(...)
    model.fit(..., pre=seq_mask_random)
```

* I created this doc for more details about the Merlin Transformer API and the proposed changes: https://docs.google.com/document/d/1auNJCOPFjyVxlhx4eUUthpMLm-rva0jC-lo0ChSZWzI/edit#

### Goals :soccer:

This PR aims to address the following limitations in the current Transformer API:

- [x] 1. **Inference for CausalLM is not supported**: the model returns scores for all positions in the input sequence but we are only interested on the last position at inference time.
- [x] 2. **CLM only supports the transform SequencePredictNext**: We cannot train or evaluate a CLM model on the last item of the sequence only (i.e using SequencePredictLast)
- [x] 3. **Specialized and complex API**: the support of masking and inference is done via specialized Merlin Blocks (`ReplaceMaskedEmbeddings`, `SequenceMaskLastInference`). So the user should be familiar with all these custom blocks to correctly define and train a transformer-based model with CLM or MLM approach.
- [x] 4. **Output of the model is a padded dense tensor of scores**: As the HuggingFace transformer layer is requiring a dense input. We are converting the ragged inputs to dense (by 0-padding to the maximum length seen in the given batch) right before calling the HF layer. The ops that follow this block are then applied to the padded dense tensor. This means that we compute the logit scores for all positions (even the padded ones) and which can be costly (such in the weight tying multiplication between the hidden representation all items embeddings).

- [x] Add Inference support to a transformer-based model trained with CLM: i.e. select the prediction score of the last non-padded position.
- [x] Support evaluating a transformer-based model trained with CLM (trained with SequencePredictNext) on the last item in the sequence (i.e using SequencePredictLast).
- [x] Simplify the transformer API by abstracting the definition of the `masking_pre` and `masking_post` blocks using the `pre` transform set in the `fit()` method.
- [x] Optimize the predictions generation of transformer-based model: Convert the dense tensor returned by the transformer layer to a `tf.RaggedTensor` so that all the logic happening after the transformer block (MLP projections, softmax layer...) is applied only on actual positions.
- [x] Add masking support for `SequencePredictNext` and `SequencePredictLast`
- [x] Add checks to ensure the SequenceTransform used in evaluate() is aligned with the `masking_pre` used to train the transformer model. 
- [x] Implement the following specialized blocks (needed for CLM and MLM support): 
    *  `SequenceCausalLastInference` to generate a mask indicating the last non-padded position of each input sequence at inference time. 
    * `ExtractMaskFromTargets` to move the logic defined in `ReplaceMaskedEmbeddings` and which consists of inferring the mask information based on targets. 
    * `TransformerOutputToRagged` to convert the output of the transformer layer to Ragged based on the mask information. 
- [x] Update the example notebooks 

### Testing Details :mag:
- Update `test_transformer_with_masked_language_modeling` and `test_transformer_with_causal_language_modeling` to account for the changes. 
- Update the

### DIFF
--- a/examples/usecases/transformers-next-item-prediction.ipynb
+++ b/examples/usecases/transformers-next-item-prediction.ipynb
@@ -673,6 +673,25 @@
   },
   {
    "cell_type": "markdown",
+   "id": "0a87439c",
+   "metadata": {},
+   "source": [
+    "Align the schema of train and validation datasets with the model's schema"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "2b90424a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "train_set_processed.schema = seq_schema\n",
+    "validation_set_processed.schema = seq_schema"
+   ]
+  },
+  {
+   "cell_type": "markdown",
    "id": "8d422833",
    "metadata": {},
    "source": [
--- a/examples/usecases/transformers-next-item-prediction.ipynb
+++ b/examples/usecases/transformers-next-item-prediction.ipynb
@@ -724,7 +724,7 @@
    "id": "0a460e4c",
    "metadata": {},
    "source": [
-    "For the transformer portion of our model, we will use the `XLNet` architecture. Additionally, we are passing `mm.ReplaceMaskedEmbeddings()` as our `pre` block. We will be training a masked language model and this parameter is responsible for the masking of our sequences."
+    "For the transformer portion of our model, we will use the `XLNet` architecture."
    ]
   },
   {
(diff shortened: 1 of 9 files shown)


gabrielspmoreira (approved):
Looks great to me @sararb .

PR #1023, opened 2023-03-17
Restoring tox.ini and .github/workflows/tensorflow.yml 

gabrielspmoreira:
### Goals :soccer:
Restoring tox.ini and .github/workflows/tensorflow.yml to reverse changes done to run models CI on a specific [dataloader branch](https://github.com/NVIDIA-Merlin/models/pull/999)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -20,11 +20,8 @@ setenv =
 sitepackages=true
 commands =
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git@{posargs:main}
-    # Enforcing the PR that implements the dataloader features changes
-    #python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@{posargs:main}
-    python -m pip install --upgrade git+https://github.com/bschifferer/dataloader.git@change_output
+    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@{posargs:main}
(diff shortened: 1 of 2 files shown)


jperez999 (approved):

tox.ini
```diff
@@ -20,11 +20,8 @@ setenv =
 sitepackages=true
 commands =
     python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/core.git@{posargs:main}
-    # Enforcing the PR that implements the dataloader features changes
-    #python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@{posargs:main}
-    python -m pip install --upgrade git+https://github.com/bschifferer/dataloader.git@change_output
+    python -m pip install --upgrade git+https://github.com/NVIDIA-Merlin/dataloader.git@{posargs:main}
```
oliverholworthy, line 23: We might need to add `--no-deps` to each of these pip installs of merlin packages to avoid the subsequent pip installs changing the version of a previous merlin package

oliverholworthy: I think it's safe to do `--no-deps` because the merlin packages installed here (e.g. the dataloader) are already part of the Merlin Models package requirements which get installed as part of every tox environment by default. 

My understanding is that main reason we have these lines here is to override the version that's installed for the release branch builds.

edknv (approved):

PR #1030, opened 2023-03-23
Skip SOK tests if SOK is not installed

edknv:
Follow-up to #863 

This PR proposes to skip unit tests for `SOKEmbedding` if SOK is not installed.

### DIFF
--- a/tests/unit/tf/horovod/test_embedding.py
+++ b/tests/unit/tf/horovod/test_embedding.py
@@ -1,3 +1,5 @@
+import importlib
+
 import numpy as np
 import pytest
 import tensorflow as tf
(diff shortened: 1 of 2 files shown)


oliverholworthy (approved):

tests/unit/tf/horovod/test_embedding.py
```diff
@@ -8,6 +10,7 @@
 
 
 @pytest.mark.skipif(not HAS_GPU, reason="No GPU available")
+@pytest.mark.skipif(importlib.util.find_spec("sparse_operation_kit") is None, reason="needs sok")
```
oliverholworthy, line 13: There's only one test class here so this works. However, if this whole test file should be skipped (even if more tests were added in future), then perhaps a top-level `pytest.importorskip("sparse_operation_kit", reason="needs SOK")` would match the intent here too?

oliverholworthy:
Unrelated to the specific change in this PR, it's relevant though so may be a good place to mention. I noticed the tests here are under a directory called `horovod`, while the corresponding modlues being tested are under a directory called `distributed`. would it be cleared if those two matched?

PR #1040, opened 2023-03-28
Enable pickle of model with TensorFlow 2.11  

oliverholworthy:
Supports #1016 

### Goals :soccer:

Enable pickle of model with TensorFlow 2.11  

- TensorFlow/Keras 2.11 enabled v3 saving format and it appears to now be expected that the `from_config` method builds the model, so that every layer is instantiated with variables needed.
- Moved the creation of variable `should_compute_train_metrics_for_batch` to `__init__` method of Model so that it creates the variable correctly when reloading model.

### Testing Details :mag:

- Existing test `test_pickle` was failing with TensorFlow 2.11

### DIFF
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -29,7 +29,13 @@ import tensorflow as tf
 from keras.engine.compile_utils import MetricsContainer
 from keras.utils.losses_utils import cast_losses_to_common_dtype
 
-from keras.saving.experimental import saving_lib
+# Required for tf/keras 2.11 Support which enabled saving_lib usage for model save/load
+# we're using this in the `Model.from_config` method
+try:
+    from keras.saving.experimental import saving_lib
+except ImportError:
+    saving_lib = None
+
 from packaging import version
 from tensorflow.keras.losses import Loss
 from tensorflow.keras.metrics import Metric
@@ -1862,14 +1868,20 @@ class Model(BaseModel):
 
         model = cls(*layers, pre=pre, post=post, schema=schema)
 
-        if getattr(saving_lib._SAVING_V3_ENABLED, "value", False):
+        # This is required for tf/keras 2.11 which enabled v3 saving
+        # and requires calling model build. The block below is a copy
+        # of part of the keras.Model.from_config method in 2.11
+        if (
+            saving_lib
+            and hasattr(saving_lib, "_SAVING_V3_ENABLED")
+            and getattr(saving_lib._SAVING_V3_ENABLED, "value", False)
+        ):
             if build_input_shape:
                 model.build(build_input_shape)
             if compile_config is not None:
                 model._compile_from_config(compile_config, base_class=Model)
 
         return model
-
 
     def get_config(self):
         super_config = super().get_config()
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -1706,7 +1706,6 @@ class Model(BaseModel):
         input_shape : tf.TensorShape, optional
             The input shape, by default None
         """
-        super().build(input_shape)
         last_layer = None
 
         if self.prep_features:
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -292,6 +292,18 @@ class ModelBlock(Block, tf.keras.Model):
 
 
 class BaseModel(tf.keras.Model):
+    def __init__(self, **kwargs):
+        super(BaseModel, self).__init__(**kwargs)
+
+        # Initializing model control flags controlled by MetricsComputeCallback()
+        self._should_compute_train_metrics_for_batch = tf.Variable(
+            dtype=tf.bool,
+            name="should_compute_train_metrics_for_batch",
+            trainable=False,
+            synchronization=tf.VariableSynchronization.NONE,
+            initial_value=lambda: True,
+        )
+
     def compile(
         self,
         optimizer="rmsprop",
@@ -1576,14 +1588,7 @@ class Model(BaseModel):
     ):
         super(Model, self).__init__(**kwargs)
 
-        # Initializing model control flags controlled by MetricsComputeCallback()
-        self._should_compute_train_metrics_for_batch = tf.Variable(
-            dtype=tf.bool,
-            name="should_compute_train_metrics_for_batch",
-            trainable=False,
-            synchronization=tf.VariableSynchronization.NONE,
-            initial_value=lambda: True,
-        )
+        
 
         context = context or ModelContext()
         if len(blocks) == 1 and isinstance(blocks[0], SequentialBlock):
--- a/merlin/models/tf/models/base.py
+++ b/merlin/models/tf/models/base.py
@@ -33,11 +33,14 @@ from tensorflow.keras.losses import Loss
 from tensorflow.keras.metrics import Metric
 from tensorflow.keras.utils import unpack_x_y_sample_weight
 
-# This is to handle TensorFlow 2.11 Saving V3 triggering with model pickle
+# This is to handle TensorFlow 2.11/2.12 Saving V3 triggering with model pickle
 try:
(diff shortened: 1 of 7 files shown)


merlin/models/tf/models/base.py
```diff
@@ -1692,6 +1701,7 @@ def build(self, input_shape=None):
         input_shape : tf.TensorShape, optional
             The input shape, by default None
         """
+        super().build(input_shape)
```
oliverholworthy, line 1704: calling super().build doesn't work in the general case with out input types. I'm going to update to perform a model call with data in `from_config` directly which is what this line is intended to achieve

merlin/models/tf/models/base.py
```diff
@@ -284,6 +285,18 @@ def _set_save_spec(self, inputs, args=None, kwargs=None):
 
 
 class BaseModel(tf.keras.Model):
+    def __init__(self, **kwargs):
+        super(BaseModel, self).__init__(**kwargs)
+
+        # Initializing model control flags controlled by MetricsComputeCallback()
+        self._should_compute_train_metrics_for_batch = tf.Variable(
```
oliverholworthy, line 292: This needs to be moved to the `__init__` method (from the compile) method so that this variable is created when reloading the model.

merlin/models/tf/models/base.py
```diff
@@ -1343,6 +1347,9 @@ def fit(
         x = _maybe_convert_merlin_dataset(x, batch_size, **kwargs)
         self._maybe_set_schema(x)
 
+        if hasattr(x, "batch_size"):
+            self._batch_size = x.batch_size
```
oliverholworthy, line 1351: We need to save the batch size used during fit so that when we reload the mode we pass it the same shape inputs. There were some tests of the two tower model that required this for example. This might indicate something to further investigate. I'm not sure it should matter what batch size we use when re-loading the model.

merlin/models/tf/transforms/features.py
```diff
@@ -63,6 +63,8 @@ def compute_output_shape(self, input_shape):
                 col_schema_shape = self.schema[name].shape
                 if col_schema_shape.is_list:
                     max_seq_length = col_schema_shape.dims[1].max
+                    if max_seq_length is not None:
+                        max_seq_length = int(max_seq_length)
```
oliverholworthy, line 67: When serializing/deserializing the schema, we get back a float value of the shape. If/when we fix that in core, this line can be removed

tests/unit/tf/experimental/test_sample_weight.py
```diff
@@ -93,11 +93,13 @@ def test_contrastive_sample_weight_serialization(ecommerce_data: Dataset, run_ea
...
     candidate = mm.Encoder(item_inputs, mm.MLPBlock([128, tower_dim]))
 
+    item_id_cardinality = item_schema["item_id"].int_domain.max + 1
+
     output = mm.ContrastiveOutput(
         DotProduct(),
         post=ContrastiveSampleWeight(
-            pos_class_weight=tf.random.uniform(shape=(1000,)),
-            neg_class_weight=tf.random.uniform(shape=(1000,)),
+            pos_class_weight=tf.random.uniform(shape=(item_id_cardinality,)),
```
oliverholworthy, line 101: This test was randomly failing depending if you get unlucky and get an item id of the max value 1000.

tests/unit/tf/core/test_combinators.py
```diff
@@ -251,7 +251,7 @@ def test_with_model(self, run_eagerly, music_streaming_data):
             layer,
             tf.keras.layers.Dense(1),
             mm.BinaryClassificationTask("click"),
-            schema=music_streaming_data.schema,
+            schema=music_streaming_data.schema.select_by_name("item_recency"),
```
oliverholworthy, line 254: This test started failing after the change to `Model.from_config` because we were passing in an inconsistent schema to the model compared with what it expected.

oliverholworthy:
I've got the tests passing locally now. The GitHub Actions jobs are now stuck in a queued state though

karlhigley:
I think it's just taking time to catch up with all the actions that got queued by previous pushes. You can go into the Actions tab and cancel those to speed it up

edknv (approved):

oliverholworthy:
> I've got the tests passing locally now.

I said I'd got this working. However, actually this test in `tests/unit/tf/examples/test_06_advanced_own_architecture.py` is now failing again. I realised that the notebook test was fixed accidentally, as a result of model.save somehow swalliowing errors raised in `Model.from_config`. After the last 2 commits to correclty create the correct inputs the call to `model(inputs)` no longer raises, which causse the reload to fail later when it finds variables with different shapes

merlin/models/tf/models/base.py
```diff
@@ -1855,13 +1866,65 @@ def from_config(cls, config, custom_objects=None):
...
+        if inputs:
+            model(inputs)
+
+        return model
+
+    def get_sample_inputs(self, batch_size=None):
+        batch_size = batch_size or 2
+        if self.input_schema is not None:
+            inputs = {}
+            for column in self.input_schema:
```
edknv, line 1881: Maybe the issue is schema filtering/propagation? If for example I change these lines to go through `self.schema` instead of `input_schema`, i.e.,
```
        if self.schema is not None:
            inputs = {}
            for column in self.schema:
````
and change the notebook to use `sub_schema` in the model by passing `sub_schema`: 
```
model = mm.Model(deep_dlrm_interaction, binary_task, schema=sub_schema)
```
I can get all tests to succeed.

oliverholworthy: The test passing is not equivalent to it running correctly unfortunately. Using self.schema instead of self.inout_schema results in an exception being thrown inside the model from_config method where we call `model(inputs)`. This is because we end up passing the target `rating_binary` as an inout which the model does not accept. That exception is caught somewhere in TensorFlow/Keras, which seems to result in essentially skipping this model call, and reverting to the previous mechanism for building the model during load.

oliverholworthy: It turns out that this new saving v3 lib in keras seems like it's only used during pickle and not a regular saved model that we do with model.save(). I've added a condition to apply this model build in from config based on this threadlocal that is set in there. 

https://github.com/NVIDIA-Merlin/models/pull/1040/commits/8e55be10195f4828c374b36dc40a0dc16d92c73d

oliverholworthy: This means that most models will be pickleable. Although we only explictly test this in one test `test_pickle`. However, there is at least one kind of model that isn't pickleable (the one in define your own architecture notebook).

oliverholworthy: This same saving_lib that is being used for pickle. is now the main saving lib in 2.12. So I think we'll have some more work to get that working reliably for all models.

edknv: This seems like a good solution. I believe it's the same with the upstream keras models, i.e., most but not all keras models are pickable, but the recommended way is to use the `save()` method not pickle.

edknv: It looks like `test_pickle` is failing after 8e55be1 and a916914 :(

oliverholworthy: Looks like GPU CI was testing against tensorflow/keras 2.12 where the `saving_lib` moved from `keras.saving.experimental.saving_lib` to `keras.saving.saving_lib`. I've added an extra import for this and seems to work. This feels kind of brittle in the sense that this could break easily in future versions when the temporary `_SAVING_V3_ENABLED` variable is removed. Unless we have some way to detect if the model is being saved with this new keras format (used for pickle) vs the saved model format.

oliverholworthy: Perhaps we can figure out if we'd like to try supporting the keras native format properly which does seem to have uncovered some bugs in various model layers, so from that perspective seems worrthwhile.

PR #1049, opened 2023-04-04
Add Transformers4Rec repo test

edknv:
This PR adds Github actions that run the unit tests in `transformers4rec`. The motivation for adding these tests is to detect if changes in Dataloader could potentially break T4Rec since T4Rec now depends on Models. These downstream repo tests are similar to [how they are set up in Core](https://github.com/NVIDIA-Merlin/core/blob/45617769166f17aeec3d455fd1a1a7a83653d259/tox.ini).

### DIFF
--- a/.github/workflows/cpu-t4r.yml
+++ b/.github/workflows/cpu-t4r.yml
@@ -32,13 +32,10 @@ jobs:
     - name: Install and upgrade python packages
       run: |
         python -m pip install --upgrade pip setuptools==59.4.0 wheel tox
+    - name: Get Branch name
+      id: get-branch-name
+      uses: NVIDIA-Merlin/.github/actions/branch-name@main
     - name: Run tests
       run: |
-        ref_type=${{ github.ref_type }}
-        branch=main
-        if [[ $ref_type == "tag"* ]]
-        then
-          git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/release*:refs/remotes/origin/release*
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -99,6 +99,19 @@ commands =
     python -m pip install .
     python -m pytest -m "not notebook" systems-{env:GIT_COMMIT}/tests/unit
 
+[testenv:py38-transformers4rec-cpu]
+passenv=GIT_COMMIT
+allowlist_externals = git
+commands =
+    ; the GIT_COMMIT env is the current commit of the core repo
```
nv-alaiacano, line 106: It should be the current commit of `models` repo.

.github/workflows/cpu-t4r.yml
```diff
@@ -0,0 +1,44 @@
...
+    - name: Install and upgrade python packages
+      run: |
+        python -m pip install --upgrade pip setuptools==59.4.0 wheel tox
+    - name: Run tests
+      run: |
+        ref_type=${{ github.ref_type }}
+        branch=main
+        if [[ $ref_type == "tag"* ]]
+        then
+          git -c protocol.version=2 fetch --no-tags --prune --progress --no-recurse-submodules --depth=1 origin +refs/heads/release*:refs/remotes/origin/release*
```
nv-alaiacano, line 41: Could you leave a comment about what is happening here?

nv-alaiacano (approved):

.github/workflows/cpu-t4r.yml
karlhigley: I think this is just standard boilerplate for fetching a shallow copy of minimal size

PR #1045, opened 2023-03-31
MM transformer training and serving example

radekosmulski:
This is just meant to be a WIP issue to share code/information. This is the file of interest `examples/usecases/transformers-next-item-prediction-single-column.ipynb`

PR #1062, opened 2023-04-19
implement review suggestions

radekosmulski:
We merged #1045 to meet the cutoff, but there were some outstanding review comments from @rnyak and @edknv. This PR implements the suggestions mentioned on #1045

### DIFF
--- a/tests/unit/tf/examples/test_usecase_transformers_next_item_prediction.py
+++ b/tests/unit/tf/examples/test_usecase_transformers_next_item_prediction.py
@@ -33,16 +33,27 @@ def test_next_item_prediction(tb):
             return date
         df['checkin'] = [generate_date() for _ in range(df.shape[0])]
         df['checkout'] = [generate_date() for _ in range(df.shape[0])]
-        df.to_csv('/tmp/train_set.csv')
+        df.to_csv('/tmpdir/train_set.csv')
         """
     )
     tb.cells[4].source = tb.cells[4].source.replace("get_booking('/workspace/data')", "")
     tb.cells[4].source = tb.cells[4].source.replace(
-        "read_csv('/workspace/data/train_set.csv'", "read_csv('/tmp/train_set.csv'"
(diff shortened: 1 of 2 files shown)


radekosmulski:
rerun tests

tests/unit/tf/examples/test_usecase_transformers_next_item_prediction.py
```diff
@@ -33,16 +34,27 @@ def generate_date():
...
-        "read_csv('/workspace/data/train_set.csv'", "read_csv('/tmp/train_set.csv'"
+        "read_csv('/workspace/data/train_set.csv'", "read_csv('/tmpdir/train_set.csv'"
     )
     tb.cells[31].source = tb.cells[31].source.replace("epochs=5", "epochs=1")
-    tb.cells[37].source = tb.cells[37].source.replace("/workspace/ensemble", "/tmp/ensemble")
+    tb.cells[37].source = tb.cells[37].source.replace("/workspace/ensemble", "/tmpdir/ensemble")
     tb.execute_cell(list(range(0, 38)))
 
-    with run_triton_server("/tmp/ensemble", grpc_port=8001):
+    with run_triton_server("/tmpdir/ensemble", grpc_port=8001):
```
oliverholworthy, line 48: This will need to be something like `f"{tmpdir}/ensemble` to convert the temp directory name into a the path.

And the `tmpdir` fixture needs to be added as a named in the arguments list to be available inside the body of the test. There's a notebook test here that uses tmpdir as an example
https://github.com/NVIDIA-Merlin/models/blob/v23.02.00/tests/unit/tf/examples/test_06_advanced_own_architecture.py#L13

oliverholworthy:
> rerun tests

adding comments to the PR to trigger re-runs of tests I don't think is required/works anymore. This was previously to re-trigger a test run of the GPU-based tests that were running in Jenkins. Now that we have a self-hosted GitHub Actions runner for that, the way to trigger a re-run is to open the _Details_ page of the relevant failed check and use the Re-run jobs button at the top right of the page.

image

oliverholworthy (approved):

PR #1075, opened 2023-04-27
Add publish step for conda package

oliverholworthy:
### Goals :soccer:

- Publish conda package for merlin-models

- Adds `packages.yaml` GitHub Actions Workflow with PyPI and Conda build/release jobs
- Updates `conda/recipe/meta.yaml` to match pattern from other merlin repos.


conda/recipes/meta.yaml
```diff
@@ -18,15 +16,19 @@ source:
...
+    - setuptools
   run:
-  {% for req in setup_py_data.get('install_requires', []) %}
-    - {{ req }}
-  {% endfor %}
     - python
+    {% for req in setup_py_data.get('install_requires', []) %}
+    - {{ req }}
+    {% endfor %}
     - tensorflow
```
oliverholworthy, line 32: wondering if we should remove the dependency on tensorflow and nvtabular in the conda package since they're optional dependencies in the pip package

conda/recipes/meta.yaml
```diff
@@ -18,17 +16,18 @@ source:
 
 build:
   number: {{ git_revision_count }}
+  noarch: python
```
oliverholworthy, line 19: This indicates that we have a pure Python package that can run on any Python version.

conda/recipes/meta.yaml
```diff
@@ -18,17 +16,18 @@ source:
...
 requirements:
+  build:
+    - python
+    - setuptools
   run:
-  {% for req in setup_py_data.get('install_requires', []) %}
-    - {{ req }}
-  {% endfor %}
     - python
-    - tensorflow
```
oliverholworthy, line 30: Removed tensorflow and nvtabular deps from this conda package config so that the install matches the pip version with these as optioinal dependencies

marcromeyn (approved):

conda/recipes/meta.yaml
oliverholworthy: removed to match the way the pip package dependencies work

PR #1079, opened 2023-05-02
Split up the test suite into three parts

karlhigley:
This moves the example tests to `tests/integration/examples` and creates a new pytest marker to separate out tests that validate full models from tests that validate smaller sub-components.


karlhigley:
- [ ]

PR #1084, opened 2023-05-04
Update minimum version of TensorFlow from 2.8  to 2.9

oliverholworthy:
- Updates minimum version of TensorFlow from 2.8  to 2.9

Removes some patching of `_set_save_spec` that was added for support for earlier TensorFlow versions.

_Motivation_. Newer versions of numpy have dropped support for the [deprecated builtin type aliaes](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations) (e.g. `np.object`). 
This causes an expression like `tf.ragged.constant([[1, 2], [3]])` to raise `AttributeError: module 'numpy' has no attribute 'object'.`

### DIFF
--- a/.github/workflows/tensorflow.yml
+++ b/.github/workflows/tensorflow.yml
@@ -20,7 +20,7 @@ jobs:
       matrix:
         python-version: [3.8]
         os: [ubuntu-latest]
-        tensorflow-version: ["~=2.8.0", "~=2.9.0", "~=2.10.0", "~=2.11.0"]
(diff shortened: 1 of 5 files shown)


edknv (approved):
Should we also bump the minimum version in `requirements/tensorflow.txt`? https://github.com/NVIDIA-Merlin/models/blob/866492b0c874d3a1436307932da0dadf5f61b01f/requirements/tensorflow.txt#LL1C11-L1C11

PR #1087, opened 2023-05-11
Introducing Block

marcromeyn:
### Goals :soccer:
First PR for the new PyTorch backend, this PR introduces the following user-facing classes:
- `Block`
- `Batch`
- `Sequence`

And internal classes:
- `TorchScriptWrapper`
- `BlockContainer`

`TorchScriptWrapper` makes it easier to work with torch-script. It will check the forward method and then stores various properties of it as constants. This way, torchscript doesn't need to worry about it.

### Testing Details :mag:
Since these are some of the lowest-level classes, I tried to maximize test-coverage as much as possible

CleanShot 2023-05-12 at 13 47 22@2x

### DIFF
--- a/merlin/models/torch/batch.py
+++ b/merlin/models/torch/batch.py
@@ -61,9 +61,8 @@ class Sequence:
 
 @torch.jit.script
 class Batch:
-
     """
-    A PyTorch scriptable class representing a batch of tabular data.
+    A PyTorch scriptable class representing a batch of data.
 
     Attributes:
         features (Dict[str, torch.Tensor]): A dictionary mapping feature names to their
@@ -116,6 +115,25 @@ class Batch:
         targets: Optional[Dict[str, torch.Tensor]] = None,
         sequences: Optional[Sequence] = None,
     ) -> "Batch":
+        """
+        Create a new `Batch` instance, replacing specified attributes with new values.
+
+        Parameters
+        ----------
+        features : Optional[Dict[str, torch.Tensor]]
+            A dictionary of tensors representing the features of the batch. Default is None.
+        targets : Optional[Dict[str, torch.Tensor]]
+            A dictionary of tensors representing the targets of the batch. Default is None.
+        sequences : Optional[Sequence]
+            An instance of the Sequence class representing sequence lengths and masks for the
+            batch. Default is None.
+
+        Returns
+        -------
+        Batch
+            A new Batch object with replaced attributes.
+        """
+
         return Batch(
             features=features if features is not None else self.features,
             targets=targets if targets is not None else self.targets,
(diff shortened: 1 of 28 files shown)


merlin/models/torch/container.py
```diff
@@ -0,0 +1,151 @@
...
+        Parameters
+        ----------
+        module : nn.Module
+            The PyTorch module to be prepended.
+
+        Returns
+        -------
+        self
+        """
+        return self.insert(0, module)
```
edknv, line 75: Should we wrap the `module` here with `self.wrap_module`?

merlin/models/torch/container.py
```diff
@@ -0,0 +1,151 @@
...
+        self.values[idx] = module
+
+        return self
+
+    def __delitem__(self, idx: Union[slice, int]) -> None:
+        self.values.__delitem__(idx)
+
+    def __add__(self, other) -> "BlockContainer":
+        for module in other:
+            self.append(module)
```
edknv, line 141: Do we need to wrap the `module` here with `self.wrap_module`?

merlin/models/torch/utils/torchscript_utils.py
```diff
@@ -0,0 +1,146 @@
...
+            An optional batch object.
+
+        Returns
+        -------
+        torch.Tensor or Dict[str, torch.Tensor]
+            The output of the model.
+        """
+
+        if self.accepts_batch:
+            if self.requires_batch:
```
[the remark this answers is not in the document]
marcromeyn: It's so that if `requires_batch` is false, we cast `Optional[Batch]` into `Batch` and throw an RuntimeException if it's None.

merlin/models/torch/batch.py
```diff
@@ -0,0 +1,206 @@
...
+        if targets is None:
+            _targets = {}
+        elif torch.jit.isinstance(targets, Dict[str, torch.Tensor]):
+            _targets = targets
+        else:
+            raise ValueError("Targets must be a tensor or a dictionary of tensors")
+        self.targets: Dict[str, torch.Tensor] = _targets
+        self.sequences: Optional[Sequence] = sequences
+
+    def replace(
```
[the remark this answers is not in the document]
marcromeyn: Similar to the replace method of a `dataclass`. We still need to iron out the details but one potential use-case is to allow the `pre` in `Model` to transform the `Batch`. For instance to set the `sequences` (by doing padding + making). Another option is to do that transformation as part of the `Trainer` class.

merlin/models/torch/container.py
marcromeyn: Right now the wrapping of the module happens inside the `insert` method. This might be confusing. Do you think it's cleaner if we do it in 2 places? (The 2nd time that would amount to a no-op since it's already wrapped).

merlin/models/torch/container.py
marcromeyn: Right now it happens inside `self.append`.

merlin/models/torch/container.py
edknv: On second look, this looks fine. I think I confused `self.insert` with `self.values.insert`.

merlin/models/torch/container.py
edknv: Same with `self.insert`. This looks fine on second look.

edknv (approved):

PR #1088, opened 2023-05-12
Adding ParallelBlock

marcromeyn:
### Goals :soccer:
Adding `ParallelBlock`. 

The definition of this block is slightly different than we are used to in TF. In this implementation each branch is of type `BlockContainer`, this makes it easier to append/prepend to it. 

### Testing Details :mag:
Since it's a low-level class, I maximized the test-coverage
CleanShot 2023-05-12 at 15 59 57@2x


.github/workflows/pytorch.yml
```diff
@@ -20,41 +20,37 @@ jobs:
...
-          python -m pip install "torch${{ matrix.torch-version }}"
-          python -m pip install .[pytorch-dev]
-      - name: Build
-        run: |
-          python setup.py develop
-      - name: Run unittests
-        run: |
-          make tests-torch
\ No newline at end of file
+#     steps:
```
edknv, line 23: Why comment out the CPU test?

PR #1092, opened 2023-05-15
Adding simple aggregations: concat & stack

marcromeyn:
### Goals :soccer:
This PR introduces two classes: `Concat` & `Stack`, which allow for more flexible stacking and concatenation of dict of tensors.

### DIFF
--- a/merlin/models/torch/transforms/agg.py
+++ b/merlin/models/torch/transforms/agg.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import Dict, Union
 
 import torch
 from torch import nn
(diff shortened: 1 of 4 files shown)


merlin/models/torch/transforms/agg.py
```diff
@@ -0,0 +1,193 @@
...
+            A tensor that is the result of concatenating
+            the input tensors along the specified dimension.
+
+        Raises
+        ------
+        RuntimeError
+            If the input tensor shapes don't match for concatenation
+            along the specified dimension.
+        """
+        sorted_tensors = [inputs[name] for name in sorted(inputs.keys())]
```
edknv, line 58: Sorting the dictionary by keys makes sense from an implementation perspective, but I wonder if it would be confusing from the user perspective, given that Python dictionaries are insertion ordered. Not sure what's best, maybe just make this behavior clear in the docstring?

marcromeyn: Yeah, I agree we should add it to the doc-string. I think in most cases it will be insertion order of the data-loader which is not user-code so ordering alphabetically at least leads to consistent behaviour.

PR #1096, opened 2023-05-16
Adding RouterBlock

marcromeyn:
### Goals :soccer:
Introducing a `RouterBlock` to make it easier to route features. This is intended to be the base-class of `TabularInputBlock` & `OutputBlock`.

### DIFF
--- a/tests/unit/torch/utils/test_selection_utils.py
+++ b/tests/unit/torch/utils/test_selection_utils.py
@@ -4,7 +4,7 @@ from merlin.models.torch.utils.selection_utils import Selectable, select_schema,
 from merlin.schema import ColumnSchema, Schema, Tags
 
 
-class Test_select_schema:
+class TestSelectSchema:
     @pytest.fixture(autouse=True)
     def setup_method(self, music_streaming_data):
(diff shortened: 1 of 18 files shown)


tests/unit/torch/utils/test_selection_utils.py
```diff
@@ -0,0 +1,85 @@
+import pytest
+
+from merlin.models.torch.utils.selection_utils import Selectable, select_schema, selection_name
+from merlin.schema import ColumnSchema, Schema, Tags
+
+
+class Test_select_schema:
```
oliverholworthy, line 7: For consistency with the other test classes, perhaps we should align on a similar style across our tests for the naming here. e.g. CamelCase like in the `test_router.py`?

marcromeyn: So far I did camel-case for classes and snake-case for functions. Do you think it's better to use camel-case for all?

edknv: `Test_select_schema` -> `TestSelectSchema`?

edknv (approved):

PR #1099, opened 2023-05-17
Add BinaryOutput

edknv:
### Goals :soccer:
This PR introduces the following classes for the pytorch backend:
- ModelOutput
- BinaryOutput

### Testing Details :mag:

Coverage:
```
Name                                            Stmts   Miss  Cover
-------------------------------------------------------------------
merlin/models/torch/outputs/__init__.py             0      0   100%
merlin/models/torch/outputs/base.py                23      0   100%
merlin/models/torch/outputs/classification.py      16      0   100%
-------------------------------------------------------------------
TOTAL                                              39      0   100%
```

### DIFF
--- a/merlin/models/torch/outputs/base.py
+++ b/merlin/models/torch/outputs/base.py
@@ -1,0 +1,92 @@
+#
+# Copyright (c) 2023, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from typing import Optional, Sequence
+
+import torch
+from torch import nn
+from torchmetrics import Metric
+
+from merlin.models.torch.block import Block
+from merlin.schema import ColumnSchema, Schema
+
+
+class ModelOutput(Block):
+    """A base class for prediction tasks.
+
+    Parameters
+    ----------
+    loss: nn.Module
+        The loss function used for training.
+    metrics: Sequence[Metric]
+        The metrics used for evaluation.
+    name: Optional[str]
+        The name of the model output.
+    schema: Optional[ColumnSchema]
+        The schema defining the column properties.
+    """
+
+    def __init__(
+        self,
+        *module: nn.Module,
+        loss: nn.Module,
+        metrics: Sequence[Metric] = (),
+        name: Optional[str] = None,
+        schema: Optional[ColumnSchema] = None,
+    ):
+        """Initializes a ModelOutput object."""
+        super().__init__(*module, name=name)
+
+        self.loss = loss
+        self.metrics = metrics
+        self._output_schema = None
+
+        if schema:
+            self.setup_schema(schema)
+        self.register_buffer("target", torch.zeros(1, dtype=torch.float32))
+
+    def setup_schema(self, schema: Optional[ColumnSchema]):
+        """Set up the schema for the output.
+
+        Parameters
+        ----------
+        schema: ColumnSchema or None
+            The schema defining the column properties.
+        """
+        self._output_schema = Schema([schema])
+
+    @property
+    def output_schema(self):
+        """The schema of the output.
+
+        Returns
+        -------
+        Schema or None
+            The schema defining the column properties.
+        """
+        return self._output_schema
+
+    def eval(self):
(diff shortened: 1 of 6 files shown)


merlin/models/torch/outputs/base.py
```diff
@@ -0,0 +1,92 @@
...
+        metrics: Sequence[Metric] = (),
+        name: Optional[str] = None,
+        schema: Optional[ColumnSchema] = None,
+    ):
+        """Initializes a ModelOutput object."""
+        super().__init__(*module, name=name)
+
+        self.loss = loss
+        self.metrics = metrics
+        self._output_schema = None
```
marcromeyn, line 54:
```suggestion
        self._output_schema: Schema = Schema()
```

merlin/models/torch/outputs/base.py
```diff
@@ -0,0 +1,92 @@
...
+
+        Parameters
+        ----------
+        schema: ColumnSchema or None
+            The schema defining the column properties.
+        """
+        self._output_schema = Schema([schema])
+
+    @property
+    def output_schema(self):
```
marcromeyn, line 71: It's probably better to remove this and just do `self.output_schema` in the constructor. This because torchscript doesn't allow `@property` and only allows for methods.

merlin/models/torch/outputs/base.py
```diff
@@ -0,0 +1,92 @@
...
+    def eval(self):
+        """Sets the module in evaluation mode.
+
+        Returns
+        -------
+        nn.Module
+            The module in evaluation mode.
+        """
+        # Reset target
+        self.target = torch.zeros(1, dtype=torch.float32)
```
marcromeyn, line 90: What about having a `create_target_buffer` method that can be used in the constructor as well?

```python
def create_target_buffer(self):
   self.register_buffer("target", torch.zeros(1, dtype=torch.float32))
```

That way, we could allow for `ModelOutput` to have targets with different dtypes if needed.

merlin/models/torch/outputs/classification.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(
+        self,
+        loss=nn.BCEWithLogitsLoss(),
+        metrics: Sequence[Metric] = (
+            Accuracy(task="binary"),
+            AUROC(task="binary"),
+            Precision(task="binary"),
+            Recall(task="binary"),
+        ),
+        schema: Optional[ColumnSchema] = None,
```
marcromeyn, line 48: Maybe we should put schema on top, since that might be the argument that will be set the most.

merlin/models/torch/outputs/base.py
```diff
@@ -0,0 +1,92 @@
...
+
+        self.loss = loss
+        self.metrics = metrics
+        self._output_schema = None
+
+        if schema:
+            self.setup_schema(schema)
+        self.register_buffer("target", torch.zeros(1, dtype=torch.float32))
+
+    def setup_schema(self, schema: Optional[ColumnSchema]):
```
marcromeyn, line 60: Do you think we need to make things more explicit and turn this into `setup_column_schema`?

tests/unit/torch/outputs/test_classification.py
```diff
@@ -0,0 +1,56 @@
...
+from torch import nn
+from torchmetrics import AUROC, Accuracy, Precision, Recall
+
+import merlin.dtypes as md
+import merlin.models.torch as mm
+from merlin.models.torch.utils import module_utils
+from merlin.schema import ColumnSchema, Schema
+
+
+class TestBinaryOutput:
```
marcromeyn, line 26: Would be great to add at least 2 extra tests:
1. Validate loss-calculation (after a forward-pass) 
2. Validate metric-calculation (after a forward-pass)

merlin/models/torch/outputs/base.py
```diff
@@ -0,0 +1,92 @@
...
+        The name of the model output.
+    schema: Optional[ColumnSchema]
+        The schema defining the column properties.
+    """
+
+    def __init__(
+        self,
+        *module: nn.Module,
+        loss: nn.Module,
+        metrics: Sequence[Metric] = (),
```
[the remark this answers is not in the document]
edknv: `torcheval` looks interesting, but I think we should focus on just `torchmetrics` for now. `torcheval` says it's in alpha and is not stable, so I feel it could become an operational burden with maintaining dependencies, figuring out backward-incompatible changes, etc.

merlin/models/torch/outputs/base.py
edknv: I guess we could go either way, but I kept `setup_schema` since `output_schema` is a `Schema` not a `ColumnSchema`.

tests/unit/torch/outputs/test_classification.py
edknv: I added a couple of tests that check the loss and the metrics, but it looks to me like the actual calculation would be better addressed after we implement `calculate_loss`.

marcromeyn (approved):

PR #1100, opened 2023-05-18
update pretrained embs example

radekosmulski:
This updates the pretrained embeddings use case to reflect the new functionality.


bschifferer:
It looks great to me. Can we add a description and print

train.schema
loader.output_schema

and explain that the loader adds the column `pretrained_movie_embeddings` to the schema + the Tags.Embedding ?

bschifferer:
otherwise, the example looks good and we can merge it

bschifferer (approved):
It looks great to me. Can we add a description and print

train.schema
loader.output_schema

and explain that the loader adds the column pretrained_movie_embeddings to the schema + the Tags.Embedding ?

PR #1101, opened 2023-05-18
[WIP] add docstrings in mms classes and functions

rnyak:
This is still WIP.

### DIFF
--- a/merlin/models/tf/blocks/dlrm.py
+++ b/merlin/models/tf/blocks/dlrm.py
@@ -34,52 +34,55 @@ def DLRMBlock(
     *,
     embedding_dim: int = None,
     embedding_options: EmbeddingOptions = None,
+    embeddings: Optional[Block] = None,
     bottom_block: Optional[Block] = None,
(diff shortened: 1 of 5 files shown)


merlin/models/tf/blocks/dlrm.py
```diff
@@ -34,52 +34,55 @@ def DLRMBlock(
     *,
     embedding_dim: int = None,
     embedding_options: EmbeddingOptions = None,
+    embeddings: Optional[Block] = None,
```
nv-alaiacano, line 37: Let's not move the order of arguments if possible - it's a breaking change in the unlikely case where people provide all of the arguments in order as opposed to using named arguments.

rnyak: @nv-alaiacano thanks for your comment. Make sense, I talked Oliver, I think we are safe changing the position of the arg thanks to `*` arg there. It makes everything forced to be a keyword arg, rather than a positional arg.

PR #1102, opened 2023-05-19
add session-based example with pretrained embeddings 

radekosmulski:
A branch/PR for sharing code/facilitating dicussion on adding the SIGIR dataset


examples/usecases/transformers-next-item-prediction-with-pretrained-embeddings.ipynb
```diff
@@ -0,0 +1,731 @@
+{
```
rnyak, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    
</pre><p>I think we should not have two workflows.. we need to merge these two csv files first (you can do it with cudf, or pandas upfront, before NVT), and then apply categorify and other ops.. otherwise it will be wrongly mapped.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

rnyak, line 1: <pre class="ql-syntax" spellcheck="false">Line #3.    cd /workspace &amp;&amp; pip install . 
</pre><p>why do we need to do pip install here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

rnyak, line 1: <pre class="ql-syntax" spellcheck="false">Line #1.    generate_data('sigir-sku', 1000).head()
</pre><p>In the original The file&nbsp;<code>sku_to_content.csv</code>&nbsp;contains also <code>category_hash</code> feature. If you want to remove it may be you can add a note here?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

radekosmulski: The way you describe it would be a good way to do it, but because of the high cardinality of the pretrained embeddings (50 in this case, but customers might likely encounter even larger sizes) our data would expand very significantly (we would need to add this information onto the 36+ million rows in the dataset).

This is only a WIP, I just wanted to see if I would be able to generate this dataset from a schema (was mostly concerned about the list columns that contain pretrained embedding information).

From what I understand that functionality is still in the works, but what we might be able to do is provide the train set along with information for pretrained embeddings separately to nvtabular/dataloader and the dataloder will be able to do the linking. 

And sorry, I just reread your comment - yes, you are spot on, that is a great observation! Yes, if this information were to be provided separately to two workflows, indeed the mapping after Categorify would be broken. But I don't believe we have the machinery available as of now to support providing both at the same time to Categorify (unless we do the merge like you describe, but this is not what we probably will want to do in the final example, once the functionality is fully implemented).

radekosmulski: I am modifying `merlin-models` to include the schema for the new datasets (`sigir-browsing` and `sigir-sku`) . In order to make them available to me in the notebook, I need to install the modified library (which inside docker lives in `/workspace`). I am just doing this to help with working on the PR but you are right, this should not exist in the final version.

radekosmulski: Yes, you are right! Well spotted, I was not aware I was missing it here, thank you for pointing it out. I brought it back (added it to `schema.json` for the synthetic data and also added it to the workflow for processing the actual data) but I have removed the `image_vector` and added a note as you advise since I don't believe we will be using it and rather will go with the `description_vector` which lends itself much better to this example (the `image_vector` is of length greater than 68k in the original data)

examples/usecases/transformers-next-item-prediction-with-pretrained-embeddings.ipynb
```diff
@@ -0,0 +1,2198 @@
...
```
rnyak, line 1: <p>read --&gt; ready</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='29'/>

radekosmulski: thank you, made the change!

examples/usecases/transformers-next-item-prediction-with-pretrained-embeddings.ipynb
```diff
@@ -0,0 +1,5658 @@
...
```
bschifferer, line 1: <p>Can we make an if-else statement using a boolean variable to indicate if we use synthetic or real data?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

bschifferer, line 1: <p>Maybe we need a new headline here to differeniate between downloading and preparing the data and the NVT workflow?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

bschifferer, line 1: <p>I do not think we should use workflow outputs from NVTabular. This would not scale to a large dataset. I think we should store them to disk via to_parquet.</p><p> </p><p>Users will adopt our examples to their dataset and it is likely they would run out of memory?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

bschifferer, line 1: <pre class="ql-syntax" spellcheck="false">Line #2.    wf = nvt.Workflow(out + 'description_vector')
</pre><p>Can you combine a list with a string?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='11'/>

bschifferer, line 1: <p>Maybe we can reference our other examples?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='18'/>

bschifferer, line 1: <pre class="ql-syntax" spellcheck="false">Line #14.        batch_size=10,
</pre><p>If you run it on the full dataset, batch_size=10 will take a long time?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='23'/>

bschifferer, line 1: <p>Can you display the input_batch?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='32'/>

bschifferer, line 1: <p>Which workflow do you deploy here?</p><p> </p><p>It seems that you expected the categoried input to Triton?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1102/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='40'/>

radekosmulski: <p>change made! indeed, this is much more elegant with using a variable than commenting</p>

radekosmulski: <p>change made</p>

radekosmulski: <p>good question! 🙂 don't believe so, just tried that a second ago and Python does the sane thing (instead of converting the string to an array of chars or something like that) and throws an error</p><p> </p><p>here however out is not a <code>list</code>  but a <code>&lt;Node Categorify&gt;</code> so it all works </p>

radekosmulski: <p>I tried this in a live notebook and it actually doesn't look that bad, rows are getting hidden behind <code>...</code>,will make a commit with the batch showing and will make a change if need be  </p>

bschifferer (approved):

bschifferer:
rerun tests

bschifferer:
rerun tests

PR #1109, opened 2023-05-23
Adding SchemaTrackingMixin

marcromeyn:
### Goals :soccer:
This PR introduces `SchemaTrackingMixin `, a mixin class for PyTorch modules to track the output shapes and dtypes of the forward pass. This is used in order to automatically generate the output-schema. It registers a hook to capture this information and provides methods to access the output schema.


merlin/models/torch/utils/schema_utils.py
```diff
@@ -0,0 +1,98 @@
...
+
+        columns = []
+
+        for name, shape in self._output_shapes.items():
+            dtype = self._output_dtypes[name]
+            dims = (None,) + tuple(shape)
+            tags = None
+
+            if len(shape) > 1 and dtype != torch.int32:
+                tags = [Tags.EMBEDDING]
```
oliverholworthy, line 86: this part seems like it might add embedding tags to things that are not embeddings if we used this mixin on modules that didn't produe embeddings?

marcromeyn: I would say that any intermediate representation throughout the model that follows the shape + dtype if-clause, could be used as an embedding?

oliverholworthy: Yeah, I suppose that makes sense, what about the multi-dimensional case, where len(shape) here is 3 or more?

merlin/models/torch/utils/schema_utils.py
```diff
@@ -0,0 +1,98 @@
...
+        output : torch.Tensor or dict
+            The output of the forward method.
+        """
+        if not module._forward_called:
+            if isinstance(output, dict):
+                for key, value in output.items():
+                    module._output_shapes[key] = value.shape
+                    module._output_dtypes[key] = value.dtype
+            else:
+                module._output_shapes["output"] = output.shape
```
oliverholworthy, line 39: this `"output"` key is for the case where a module outputs a single tensor? Would something that uses this output schema depend on the name of this column name elsewhere?

oliverholworthy (approved):

merlin/models/torch/utils/schema_utils.py
```diff
@@ -0,0 +1,98 @@
...
+        """
+        self._forward_called = False
+        self._handle = None
+        self._output_shapes = {}
+        self._output_dtypes = {}
+
+        if self._handle is None:
+            self._handle = self.register_forward_hook(self._post_forward_hook)
+
+    def output_schema(self) -> Schema:
```
oliverholworthy, line 56: Perhaps this could be a property instead? To make this consistent with the TF models API and other places we've got `output_schema` defined on our classes?

merlin/models/torch/utils/schema_utils.py
marcromeyn: That would amount to a sequence, which I would argue still could be used as an embedding. As long as we track the dims, downstream users know what to do with it.

merlin/models/torch/utils/schema_utils.py
marcromeyn: Yeah, I would love to. The issue is that torchscript doesn't allow properties. So unfortunately for now I think we need to have it as a method.

PR #1110, opened 2023-05-23
Add docstrings

edknv:
Adding docstrings from the doc bash.


merlin/models/tf/core/encoder.py
```diff
@@ -526,23 +627,90 @@ def fit(self, *args, **kwargs):
 
 @tf.keras.utils.register_keras_serializable(package="merlin.models")
 class EmbeddingEncoder(Encoder):
+    """Creates an Encoder from an EmbeddingTable.
+    Typically used with RetrievalModelV2.
+    Parameters
```
nv-alaiacano, line 632: We need a whitespace line before `Parameters` and before `Returns` or else it won't format right. 

Here's an example of how this looks when the spacing is missing: https://nvidia-merlin.github.io/models/review/pr-1110/generated/merlin.models.tf.EmbeddingEncoder.html#merlin.models.tf.EmbeddingEncoder

Other places where this occurs:

* ModelBlock
* MetricsComputeCallback
* _maybe_convert_merlin_dataset
* get_task_name
* _get_col_set_by_tags
* _set_task_block
* default_binary_metrics
* default_categorical_prediction_metrics

nv-alaiacano: I resolved these and some other instances of this pattern in the repo.

nv-alaiacano (approved):

PR #1115, opened 2023-05-26
Add RegressionOutput

edknv:
This PR introduces the `RegressionOutput` class, a prediction block for regression tasks, for the pytorch backend.

Depends on #1099.


merlin/models/torch/outputs/regression.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def setup_schema(self, target: Optional[ColumnSchema]):
+        """Set up the schema for the output.
+
+        Parameters
+        ----------
+        target: Optional[ColumnSchema]
+            The schema defining the column properties.
+        """
+        _target = target.with_dtype(md.float32)
```
marcromeyn, line 61: Would be good to add the `CONTINUOUS` tag.

marcromeyn (approved):

PR #1120, opened 2023-05-26
[TMP] Try reverting change to item-id cardinality in sequence testing schema

oliverholworthy:
Fixes # (issue)


oliverholworthy:
Validated that this change in cardinality triggered tolerance test errors in `test_retrieval_transformer`. Possible solution in #1122

PR #1122, opened 2023-05-26
Use `tf.matmul` instead of `np.dot` in `test_retrieval_transformer` 

oliverholworthy:
Fixes issue with `test_retrieval_transformer` failing in Blossom tests

- The test `test_retrieval_transformer` is currently failing in Blossom with a tolerance error. This seems to be something to do with small differences in numerical results when running the computation on CPU vs GPU.

Using `tf.matmul` to run this operation on the same GPU device (when available) as the operation in the model runs.

### Testing Details :mag:

Tested this by submitting a custom blossom job and running the test multiple times with pytest-repeat


marcromeyn (approved):

gabrielspmoreira (approved):
Thanks for catching this @oliverholworthy !

PR #1118, opened 2023-05-26
Fixing ci skipping

marcromeyn:
### Goals :soccer:
https://github.com/NVIDIA-Merlin/models/pull/1113 contained some bugs. This PR fixes those.

PR #1124, opened 2023-05-26
Fix CI path detection for changed files

edknv:
Follow-up to #1118.

`ci_utils.get_changed_backends()` returns the full names such as `tensorflow` and `xgb` but we are checking if these exist in the paths such as `merlin/models/tf` or `merlin/models/xgb`, so currently tests are not performed even if there is a change in tensorflow or xgboost. This PR fixes this.


marcromeyn:
We check if the alias is present in the path and if so, we output the full name [source](https://github.com/NVIDIA-Merlin/models/blob/main/merlin/models/utils/ci_utils.py#L105-L106). That should be correct, right? We tested it in this [PR](https://github.com/NVIDIA-Merlin/models/pull/1118/checks), where all backend-tests were run successfully when one of the shared modules was updated.

marcromeyn (approved):

PR #1126, opened 2023-05-30
Add Model class

edknv:
This PR introduces the `Model` class for constructing models from individual blocks in the pytorch backend.

Some follow-up will be necessary, such as testing with a toy dataset using dataloader, `batch_predict`, etc.


merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Example usage
+    -------------
+    >>> model = Model(
+    ...    TabularInputBlock(schema),
+    ...    MLPBlock([32, 16]),
+    ...    BinaryOutput(schema.select_by_tag(Tags.TARGET).first),
+    ... )
+    ... trainer = Trainer(max_epochs=1)
+    ... with Loader(dataset, batch_size=16) as loader:
```
marcromeyn, line 58: Is it required to use it as a context-manager?

merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def configure_optimizers(self):
+        """Configures the optimizer for the model."""
+        return self.optimizer(self.parameters())
+
+    def model_outputs(self) -> List[ModelOutput]:
+        """Finds all instances of `ModelOutput` in the model."""
+        return module_utils.find_all_instances(self, ModelOutput)
+
+    def first(self) -> nn.Module:
```
marcromeyn, line 118: Isn't this part of the `Block` class as well?

merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+        """Returns the first block in the model."""
+        return self.blocks.values[0]
+
+    def last(self) -> nn.Module:
+        """Returns the last block in the model."""
+        return self.blocks.values[-1]
+
+    def input_schema(self) -> Schema:
+        """Returns the input schema of the model."""
+        if self.schema:
```
marcromeyn, line 128: Maybe add a `TODO` here that we need to infer this from the `TabularInputBlock` when we get there.

merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+    """
+    if len(model_outputs) < 1:
+        raise RuntimeError("No model outputs found.")
+    if (
+        isinstance(predictions, torch.Tensor)
+        and isinstance(targets, torch.Tensor)
+        and len(model_outputs) > 1
+    ):
+        raise RuntimeError("Multiple outputs but only one target was provided.")
+
```
marcromeyn, line 200: Do we need to check if all `model_outputs` have a `target` property set?

merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+    def initialize(self, data: Union[Dataset, Loader, Batch]):
+        """Initializes the model based on a given data set."""
+        return initialize(self, data)
+
+    def forward(
+        self, inputs: Union[torch.Tensor, Dict[str, torch.Tensor]], batch: Optional[Batch] = None
+    ):
+        """Performs a forward pass through the model."""
+        outputs = inputs
```
marcromeyn, line 89: Same here as for the __init__, should we break this out in a function so it can be shared between here and the `Block` class.

merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+        for block in self.blocks.values:
+            outputs = block(outputs, batch=batch)
+        for post in self.post.values:
+            outputs = post(outputs, batch=batch)
+        return outputs
+
+    def training_step(self, batch, batch_idx):
+        """Performs a training step with a single batch."""
+        del batch_idx
+        inputs, targets = batch
```
marcromeyn, line 101: We need some logic here to construct the `Batch` class and pass it to the forward-pass.

merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+    def last(self) -> nn.Module:
+        """Returns the last block in the model."""
+        return self.blocks.values[-1]
+
+    def input_schema(self) -> Schema:
+        """Returns the input schema of the model."""
+        if self.schema:
+            return self.schema
+        return Schema([])
+
```
marcromeyn, line 131: We would need to add a method here for `output_schema` that combines all the output-schema's of the various model-outputs.

merlin/models/torch/models/base.py
edknv: No, it's probably not necessary in the torch case, but I want to promote this idiom using the context manager everywhere, because the tensorflow equivalent has a memory leak in some cases without a context manager.

merlin/models/torch/models/base.py
```diff
@@ -0,0 +1,226 @@
...
+        self,
+        *blocks: nn.Module,
+        schema: Optional[Schema] = None,
+        optimizer=torch.optim.Adam,
+    ):
+        """Initializes `Model` class"""
+        super().__init__()
+        self.schema = schema
+
+        self.pre = BlockContainer(name="pre")
```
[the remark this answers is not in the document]
edknv: Yeah, that would be cleaner, but I'm having trouble with torchscript shenanigans.

merlin/models/torch/models/base.py
edknv: I updated the logic to check if `model_outputs` has a `target` when no targets are provided in [lines 205-206](https://github.com/NVIDIA-Merlin/models#diff-ae80586315e21aadf21557b8b7f837710d8a5b5138f3787ca51d68dacd158359R205-R206).

merlin/models/torch/models/base.py
edknv: Added an `output_schema()` method.

marcromeyn (approved):

PR #1134, opened 2023-06-05
Remove use of deprecated start_index from Categorify

oliverholworthy:
### Goals :soccer:

Fix use of Categorify to support new version of NVTabular

- `start_index` was removed from `Categorify` in https://github.com/NVIDIA-Merlin/NVTabular/pull/1692

### DIFF
--- a/examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb
+++ b/examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb
@@ -468,7 +468,7 @@
    "source": [
     "%%time\n",
     "item_features_names = ['f_' + str(col) for col in [47, 68]]\n",
-    "cat_features = [['item_id', 'purchase_id']] + item_features_names >> nvt.ops.Categorify(start_index=1)\n",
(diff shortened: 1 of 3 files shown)


examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb
```diff
@@ -468,7 +468,7 @@
    "source": [
     "%%time\n",
     "item_features_names = ['f_' + str(col) for col in [47, 68]]\n",
-    "cat_features = [['item_id', 'purchase_id']] + item_features_names >> nvt.ops.Categorify(start_index=1)\n",
```
rnyak, line 471: May be we can add a note for the user something like that:

```
Categorify op maps nulls to `1`, OOVs to `2` and the frequent categories are encoded starting from `3`. Please note that we reserve `0` for padding value.
```

rnyak (approved):

examples/usecases/ecommerce-session-based-next-item-prediction-for-fashion.ipynb
nv-alaiacano: can I propose merging this fix to get things working then you can update the copy?

nv-alaiacano (approved):

PR #1135, opened 2023-06-06
Move horovod and sok init functions to user code

edknv:
### Goals :soccer:

- Remove `hvd.init()` and `sok.init()` and have the users run them.
- Introduce `singlegpu` and `multigpu` markers.

In some cases when horovod is not built correctly or not configured correctly with MPI, `hvd.init()` or `sok.init()` will throw an MPI_LIB errors on the C-level, which we cannot catch from Python. This happened in https://github.com/NVIDIA-Merlin/models/pull/1134 for example.

Following @oliverholworthy's proposal, this PR leaves `hvd.init()` and/or `sok.init()` up to the user.

We still need to initialize them to run unit tests but incorrect installation in the ci-runner will still block the CI, so this PR also introduces `singlegpu` and `multigpu` pytest markers to skip the horovod tests in the single-gpu case, similarly to https://github.com/NVIDIA-Merlin/Merlin/pull/999.

### DIFF
--- a/.github/workflows/gpu.yml
+++ b/.github/workflows/gpu.yml
@@ -34,7 +34,7 @@ jobs:
           if [[ "${{ github.ref }}" != 'refs/heads/main' ]]; then
               extra_pytest_markers="and changed"
           fi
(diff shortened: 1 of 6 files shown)


edknv:
Ci is unblocked for now in https://github.com/NVIDIA-Merlin/models/pull/1136. Will close this PR and revisit if necessary.

PR #1140, opened 2023-06-09
Set upper bound on `transformers` dependency to `<4.30`

oliverholworthy:
Fixes Error we're seeing in our tests with TensorFlow versions prior to 2.11

```
/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/transformers/modeling_tf_utils.py:84: in <module>
    from tensorflow.python.keras.engine import call_context
E   ImportError: cannot import name 'call_context' from 'tensorflow.python.keras.engine' (/opt/hostedtoolcache/Python/3.8.16/x64/lib/python3.8/site-packages/tensorflow/python/keras/engine/__init__.py)
```

The newest version of the transformers package 4.30.0 started importing a module that only exists in TensorFlow 2.11 and above.

Setting this upper bound for now until we drop support for TensorFlow 2.10 and earlier, or this is updated in `transformers` to restore support for earlier versions of TensorFlow ([related issue here](https://github.com/huggingface/transformers/issues/24133))

### DIFF
--- a/requirements/transformers.txt
+++ b/requirements/transformers.txt
@@ -1,1 +1,5 @@
-transformers>=4.23,<5
+# Version 4.30 of `transformers` started using functionality in TensorFlow 2.11
+# Until either that is changed in transformers or, we drop support
+# for TensorFlow versions prior to 2.11, then we need to keep this
+# upper bound on the version specifier
+transformers>=4.23,<4.30


oliverholworthy:
4.30.1 of `transformers` was released with a fix that restores support for earlier versions of TensorFlow

PR #1146, opened 2023-06-13
Run tests marked with `notebook` marker in examples workflow

edknv:
### Goals :soccer:

- Replace `example` pytest markers with `examples` (with an s).
- Run tests marked with `pytest.mark.notebook` in `examples` workflow.

- `ci_utils` was searching for tests with `/example/` (without an s) in their paths but none was being marked with the marker because tests were in `/examples/` (with an s). https://github.com/NVIDIA-Merlin/models/blob/92833fa0cfca002a7cf4e009b7c09bae53a336f2/tests/conftest.py#L101-L103 https://github.com/NVIDIA-Merlin/models/blob/92833fa0cfca002a7cf4e009b7c09bae53a336f2/merlin/models/utils/ci_utils.py#L28 This PR replaces all `example` markers with `examples` so the right paths can be extracted.
- Since all notebook tests belong to examples, this PR includes `notebook` tests in the `examples` workflows so that even if the pytest marks from path extraction in the above fails notebook tests run in examples workflows.

### DIFF
--- a/merlin/models/utils/ci_utils.py
+++ b/merlin/models/utils/ci_utils.py
@@ -25,7 +25,7 @@ BACKEND_ALIASES = {
     "lightfm": "lightfm",
     "xgb": "xgboost",
 }
-OTHER_MARKERS = {"unit", "integration", "example", "datasets", "horovod", "transformers"}
+OTHER_MARKERS = {"unit", "integration", "examples", "datasets", "horovod", "transformers"}
 SHARED_MODULES = {
     "/datasets/",
     "/models/config/",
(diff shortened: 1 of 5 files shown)


oliverholworthy (approved):

merlin/models/utils/ci_utils.py
```diff
@@ -25,7 +25,7 @@
     "lightfm": "lightfm",
     "xgb": "xgboost",
 }
-OTHER_MARKERS = {"unit", "integration", "example", "datasets", "horovod", "transformers"}
+OTHER_MARKERS = {"unit", "integration", "examples", "datasets", "horovod", "transformers"}
```
oliverholworthy, line 28: I was wondering why the examples tests were running in the main tensorflow test job. Looks like this is why, because the tests directory name is `../examples/...`, and we were looking for `example` so didn't tag the tests with the corresponding marker.

PR #1156, opened 2023-06-26
Match shape and dtype in loss calculation in pytorch backend

edknv:
This PR uncomments a basic use case of training a model with `TabularInputBlock` using the Lightning trainer.


edknv:
Redundant after #1171.

PR #1158, opened 2023-06-26
Add CategoricalOutput

marcromeyn:
This PR adds `CategoricalOutput`, a prediction block for categorical features.


merlin/models/torch/inputs/embedding.py
```diff
@@ -374,6 +374,14 @@ def update_feature(self, col_schema: ColumnSchema) -> "EmbeddingTable":
 
         return self
 
+    def feature_weights(self, name: str):
+        if name not in self.domains:
+            raise ValueError()
```
oliverholworthy, line 379: Are we missing an exception message here?

edknv (approved):

PR #1161, opened 2023-06-27
first version of sequence transforms applied to Batch input

sararb:
Fixes # (issue)

### Goals :soccer:
- Add support for padding, transforming, and masking sequential inputs data in MM Pytorch backend
- The implemented transform classes should: 
        -  Support multiple targets 
        -  Be used for training, evaluation, and inference
 

- [ ] Implement `TabularBatchPadding` to pad a group of sequential inputs
- [ ] Implement `TabularPredictNext` for generating targets of causal next item prediction 
- [ ] Implement `TabularPredictLast` for generating targets of last item prediction 
- [ ] Implement `TabularPredictRandom` for generating targets of predicting one random item and truncate the sequence so that the random item is at the last position. 
- [ ] Implement `TabularMaskRandom` for masked language modeling training (MLM) strategy 
- [ ] Implement `TabularMaskLast` for masking last item in the sequence, generally used to evaluate models trained with MLM. 

### Testing Details :mag:
- Defined tests for padding and the different sequence transformation

### DIFF
--- a/merlin/models/torch/sequences.py
+++ b/merlin/models/torch/sequences.py
@@ -1,0 +1,422 @@
+#
+# Copyright (c) 2023, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+from typing import Dict, List, Optional, Tuple, Union
+
+import torch
+import torch.nn.functional as F
+from torch import nn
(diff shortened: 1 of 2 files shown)


merlin/models/torch/sequences.py
```diff
@@ -0,0 +1,422 @@
...
+import torch.nn.functional as F
+from torch import nn
+
+from merlin.models.torch.batch import Batch, Sequence
+from merlin.schema import ColumnSchema, Schema, Tags
+
+MASK_PREFIX = "__mask"
+
+
+class TabularBatchPadding(nn.Module):
```
marcromeyn, line 28: Should we rename to `TabularPadding`?

merlin/models/torch/sequences.py
```diff
@@ -0,0 +1,453 @@
...
+from torch import nn
+
+from merlin.models.torch.batch import Batch, Sequence
+from merlin.models.torch.schema import Selection, select
+from merlin.schema import ColumnSchema, Schema, Tags
+
+MASK_PREFIX = "__mask"
+
+
+class TabularPadding(nn.Module):
```
marcromeyn, line 29: I think it might be better to break padding out into a smaller PR first, and then do masking afterwards.

PR #1162, opened 2023-06-29
Add DLRM block

edknv:
This PR adds the DLRM block in the PyTorch backend.

### DIFF
--- a/merlin/models/torch/blocks/dlrm.py
+++ b/merlin/models/torch/blocks/dlrm.py
@@ -1,0 +1,135 @@
+from typing import Dict, Optional
+
+import torch
+from torch import nn
+
+from merlin.models.torch.block import Block
+from merlin.models.torch.inputs.embedding import EmbeddingTables
+from merlin.models.torch.inputs.tabular import TabularInputBlock
+from merlin.models.torch.link import Link
+from merlin.models.torch.transforms.agg import MaybeAgg, Stack
+from merlin.models.utils.doc_utils import docstring_parameter
+from merlin.schema import Schema, Tags
+
+_DLRM_REF = """
+    References
+    ----------
+    ..  [1] Naumov, Maxim, et al. "Deep learning recommendation model for
+        personalization and recommendation systems." arXiv preprint arXiv:1906.00091 (2019).
+"""
+
+
+@docstring_parameter(dlrm_reference=_DLRM_REF)
+class DLRMInputBlock(TabularInputBlock):
+    """ "Input-block for DLRM model.
+
+    Parameters
+    ----------
+    schema : Schema, optional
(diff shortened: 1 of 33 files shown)


merlin/models/torch/blocks/dlrm.py
```diff
@@ -0,0 +1,135 @@
...
+
+@docstring_parameter(dlrm_reference=_DLRM_REF)
+class DLRMInteraction(nn.Module):
+    """
+    This class defines the forward interaction operation as proposed
+    in the DLRM
+     `paper https://arxiv.org/pdf/1906.00091.pdf`_ [1]_.
+
+    This forward operation performs elementwise multiplication
+    of the embeddings followed by a reduction sum.
```
radekosmulski, line 62:
```suggestion
    followed by a reduction sum (equivalent to a dot product) of all embedding pairs.
```
I read this and I was a bit confused what we are reducing, are we summing up the dot products of all embeddings?

Took me a moment to realize what is really happening here is that we are performing a dot product on each of the embedding pairs and outputting a concatenation of these operations as `interactions_flat`.

At least I believe that my understanding here is correct but I haven't run the code... anyhow, please take this comment with a grain of salt, feel free to disregard if you don't find it helpful!

tests/unit/torch/outputs/test_tabular.py
```diff
@@ -29,6 +29,8 @@ def test_exceptions(self):
         with pytest.raises(ValueError, match="not found"):
             mm.TabularOutputBlock(self.schema, init="not_found")
 
+    def test_no_route_for_non_existent_tag(self):
         outputs = mm.TabularOutputBlock(self.schema)
-        with pytest.raises(ValueError):
-            outputs.add_route(Tags.CATEGORICAL)
+        outputs.add_route(Tags.CATEGORICAL)
```
edknv, line 34: This was changed to not raise an error because it is possible that there is no continuous features in the DLRM block. Alternatively, we could have a try-except in the DLRM block, but this felt more natural to me.

tests/unit/torch/models/test_base.py
```diff
@@ -133,11 +133,11 @@ def test_training_step_with_dataloader(self):
             mm.BinaryOutput(ColumnSchema("target")),
         )
 
-        feature = [[1.0, 2.0], [3.0, 4.0]]
-        target = [[0.0], [1.0]]
+        feature = [2.0, 3.0]
+        target = [0.0, 1.0]
         dataset = Dataset(pd.DataFrame({"feature": feature, "target": target}))
```
edknv, line 138: This test case was incorrect because `feature = [[1.0, 2.0], [3.0, 4.0]]` will produce a list column.

marcromeyn (approved):

PR #1171, opened 2023-07-03
Add DLRM Model

edknv:
This PR adds a `DLRMModel` function that builds a `Model` using the `DLRMBlock` added in #1162.


merlin/models/torch/models/ranking.py
```diff
@@ -0,0 +1,75 @@
...
+from torch import nn
+
+from merlin.models.torch.block import Block
+from merlin.models.torch.blocks.dlrm import DLRMBlock
+from merlin.models.torch.models.base import Model
+from merlin.models.torch.outputs.tabular import TabularOutputBlock
+from merlin.schema import Schema
+
+
+def DLRMModel(
```
marcromeyn, line 12: I think it would be better to turn this into a class that extends `Model`. That would provide more flexibility to our users since they could extend it. In the past, I think we might have over-used using functions (would make sense if the output-type could differ).

marcromeyn (approved):

PR #1173, opened 2023-07-03
Adding MMOE & PLE

marcromeyn:
Fixes # (issue)

### DIFF
--- a/merlin/models/torch/blocks/experts.py
+++ b/merlin/models/torch/blocks/experts.py
@@ -1,0 +1,113 @@
+from typing import Dict, Optional
+
+import torch
+from torch import nn
+
+from merlin.models.torch.block import (
+    Block,
+    ParallelBlock,
+    ShortcutBlock,
+    repeat_parallel,
+    repeat_parallel_like,
+)
+from merlin.models.torch.transforms.agg import Stack
+
+
+class MMOEBlock(Block):
+    def __init__(
+        self, expert: nn.Module, num_experts: int, outputs: Optional[ParallelBlock] = None
+    ):
+        super().__init__(
+            ShortcutBlock(repeat_parallel(expert, num_experts, agg="stack"), output_name="experts")
+        )
+        if isinstance(outputs, ParallelBlock):
+            self.append(repeat_parallel_like(ExpertGateBlock(len(outputs)), outputs))
+        else:
+            self.append(ExpertGateBlock(1))
+
+
+class PLEBlock(Block):
+    def __init__(
+        self,
+        expert: nn.Module,
+        num_shared_experts: int,
+        num_task_experts: int,
+        depth: int,
+        outputs: ParallelBlock,
+    ):
+        cgc = CGCBlock(
+            expert, num_shared_experts, num_task_experts, outputs=outputs, shared_gate=True
+        )
+        super().__init__(*cgc.repeat(depth - 1))
+        self.append(CGCBlock(expert, num_shared_experts, num_task_experts, outputs=outputs))
+
+
+class CGCBlock(Block):
+    def __init__(
+        self,
+        expert: nn.Module,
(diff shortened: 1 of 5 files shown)


tests/unit/torch/blocks/test_experts.py
```diff
@@ -0,0 +1,123 @@
...
+            assert mmoe[0][0][str(i)][1].out_features == 2
+            assert mmoe[0][0][str(i)][3].out_features == 2
+        assert isinstance(mmoe[0][0].post[0], mm.Stack)
+        assert isinstance(mmoe[1], ExpertGateBlock)
+        assert mmoe[1][0][0].out_features == 2
+
+    def test_init_with_outputs(self):
+        outputs = mm.ParallelBlock({"a": mm.BinaryOutput(), "b": mm.BinaryOutput()})
+        outputs.prepend_for_each(mm.MLPBlock([2]))
+        outputs.prepend(MMOEBlock(mm.MLPBlock([2, 2]), 2, outputs))
```
gabrielspmoreira, line 48: The API is very clean!. But building MTL models from the output to the input kind of works, but might be counter-intuitive. Will that be the pattern for MTL models?
But I think I understand the challenges of building such MMOE models from the inputs, as gates depend on # experts and number of gates and towers depend on numbers of outputs.

merlin/models/torch/blocks/experts.py
```diff
@@ -0,0 +1,244 @@
...
+    expert : nn.Module
+        The base expert model to be used.
+    num_shared_experts : int
+        The number of shared experts.
+    num_task_experts : int
+        The number of task-specific experts.
+    outputs : ParallelBlock
+        The output block.
+    shared_gate : bool, optional
+        If true, use a shared gate for all tasks. Defaults to False.
```
gabrielspmoreira, line 117: I think this argument name and docstrings is misleading. When `num_shared_experts>0` all task gates will use the shared experts, right?
This `shared_gate` argument seems to be responsible to make it available the output of the shared expert (`shortcut` and `experts` keys) from one CGCBlock to the next one in PLE architecture? If that is the case, we could try and clarify that.

merlin/models/torch/blocks/experts.py
```diff
@@ -0,0 +1,244 @@
...
+        outputs: ParallelBlock,
+    ):
+        cgc_kwargs = {
+            "expert": expert,
+            "num_shared_experts": num_shared_experts,
+            "num_task_experts": num_task_experts,
+            "outputs": outputs,
+        }
+        super().__init__(*CGCBlock(shared_gate=True, **cgc_kwargs).repeat(depth - 1))
+        self.append(CGCBlock(**cgc_kwargs))
```
gabrielspmoreira, line 97: Good trick to avoid outputting the shared experts in the last layer.

merlin/models/torch/blocks/experts.py
```diff
@@ -0,0 +1,244 @@
...
+
+class PLEExpertGateBlock(Block):
+    """
+    Progressive Layered Extraction (PLE) Expert Gate Block.
+
+    Parameters
+    ----------
+    num_experts : int
+        The number of experts used.
+    experts : nn.Module
```
gabrielspmoreira, line 188: Could this `experts` argument maybe be renamed to `task_experts`, to make it clear they are different from the shared experts whose output is fed within the gate?

gabrielspmoreira (approved):
The API is very clean. I debugged the tests to better understand and implementation seems ok. Just added some optional suggestions

tests/unit/torch/blocks/test_experts.py
marcromeyn: Yeah, it might make sense to add a `MMOEOutputs` or something that does this under-the-hood.

PR #1180, opened 2023-07-05
add 1st pytorch example

radekosmulski:
This adds the first pytorch example 🥳

Additionally, I propose in `DLRMModel` we rename `dim` to `embedding_dim`. This aligns us with the tf api and (which is probably more important) is more informative to the reader (this is the language used in the paper, otherwise it is a bit confusing what does `output` refer to -- is it the output of the model? of mlp hidden layers? `embedding_dim` is more informative)


examples/pytorch/01-Getting-started.ipynb
```diff
@@ -0,0 +1,364 @@
...
+    "# Each user is responsible for checking the content of datasets and the\n",
+    "# applicable licenses and determining if suitable for the intended use."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "23d9bf34",
+   "metadata": {},
+   "source": [
+    "<img src=\"https://developer.download.nvidia.com/notebooks/dlsw-notebooks/merlin_models_01-getting-started/nvidia_logo.png\" style=\"width: 90px; float: right;\">\n",
```
bschifferer, line 34: we might need to update the logo?

examples/pytorch/01-Getting-started.ipynb
```diff
@@ -0,0 +1,364 @@
+{
```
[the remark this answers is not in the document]
marcromeyn: This is something @edknv suggested, it ensures that the background thread gets removed. I am working on a way to see if we can move this inside our model/trainer code. Because I think the context manager approach works for single GPU, but I don't think it will work in a multi-GPU setting.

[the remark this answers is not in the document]
radekosmulski: good point! removed the cell for now and will add it once we have more examples

examples/pytorch/01-Getting-started.ipynb
radekosmulski: yes, absolutely! good point, created a new tracking logo

examples/pytorch/01-Getting-started.ipynb
```diff
@@ -0,0 +1,348 @@
...
```
rnyak, line 1: <pre class="ql-syntax" spellcheck="false">Line #4.    model.initialize(train_loader)
</pre><p>can we add some explanation why we do need <code>model.initialize()</code> step? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1180/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='10'/>

rnyak, line 1: <p>do we need this entire block again?</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/NVIDIA-Merlin/models/pull/1180/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='14'/>

radekosmulski: I am not sure -- was just copying over what we have on the TF side, we follow the same pattern there

PR #1185, opened 2023-07-05
Add DCN Model

edknv:
This PR adds a `DCNModel`.


merlin/models/torch/models/ranking.py
```diff
@@ -74,3 +78,76 @@ def __init__(
...
+    Model
+        An instance of Model class representing the fully formed DCN.
+
+    Example usage
+    -------------
+    >>> model = mm.DCNModel(
+    ...    schema,
+    ...    depth=2,
+    ...    deep_block=mm.MLPBlock([256, 64]),
+    ...    output_block=BinaryOutput(ColumnSchema("target")))
```
marcromeyn, line 115: Should we try to make our examples black-compatible? In that case I guess it would put the last `)` on a new line

merlin/models/torch/models/ranking.py
```diff
@@ -74,3 +78,76 @@ def __init__(
...
+            deep_block = MLPBlock([512, 256])
+
+        if input_block is None:
+            input_block = TabularInputBlock(schema, init="defaults")
+
+        if output_block is None:
+            output_block = TabularOutputBlock(schema, init="defaults")
+
+        if stacked:
+            cross_network = Block(CrossBlock.with_depth(depth), deep_block)
```
marcromeyn, line 146: `deep_block` seems optional so we should handle the case where it's None

marcromeyn (approved):

PR #1192, opened 2023-07-10
Introducing BatchBlock

marcromeyn:
### Goals :soccer:
`BatchBlock` will be used inside the `Model` to create the `Batch` object. It's also useful for things like masking & padding.

### DIFF
--- a/merlin/models/torch/batch.py
+++ b/merlin/models/torch/batch.py
@@ -277,6 +277,13 @@ class Batch:
 
         raise ValueError("Batch has multiple target, please specify a target name")
 
+    def flatten_as_dict(self, inputs: "Batch") -> Dict[str, torch.Tensor]:
+        raise NotImplementedError()
+
+    @staticmethod
+    def from_partial_dict(input: Dict[str, torch.Tensor], batch: "Batch") -> "Batch":
+        raise NotImplementedError()
+
     def __bool__(self) -> bool:
         return bool(self.features)
 
--- a/merlin/models/torch/batch.py
+++ b/merlin/models/torch/batch.py
@@ -295,6 +295,12 @@ class Batch:
 
         raise ValueError("Batch has multiple target, please specify a target name")
 
+    def inputs(self) -> Union[torch.Tensor, Dict[str, torch.Tensor]]:
+        if len(self.features) == 1 and "default" in self.features:
+            return self.features["default"]
+
+        return self.features
+
     def flatten_as_dict(self, inputs: Optional["Batch"]) -> Dict[str, torch.Tensor]:
         """
         Flatten features, targets, and sequences into a dictionary of tensors.
--- a/merlin/models/torch/batch.py
+++ b/merlin/models/torch/batch.py
@@ -65,6 +65,9 @@ class Sequence:
     def __contains__(self, name: str) -> bool:
         return name in self.lengths
 
+    def __bool__(self) -> bool:
+        return bool(self.lengths)
+
     def length(self, name: str = "default") -> torch.Tensor:
         """Retrieves a length tensor from a sequence by name.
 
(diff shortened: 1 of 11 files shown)


tests/unit/torch/test_block.py
```diff
@@ -377,3 +377,58 @@ def forward(self, inputs: Dict[str, torch.Tensor]) -> Tuple[torch.Tensor, torch.
...
+    def test_forward_with_tuple(self):
+        feat, target = torch.tensor([1, 2]), torch.tensor([3, 4])
+        outputs = module_utils.module_test(mm.BatchBlock(), feat, targets=target)
+
+        assert isinstance(outputs, mm.Batch)
+        assert torch.equal(outputs.feature(), feat)
+        assert torch.equal(outputs.target(), target)
+
+    def test_forward_exception(self):
+        with pytest.raises(RuntimeError):
```
oliverholworthy, line 404: I noticed that in some places we use the  `match=` argument of `pytest.raises` to specify which error we're expecting. I think this is a good pattern to adopt and will make things more robust. Since this could start failing from a different place accidentally and still pass the test, changing what the test is checking.

tests/unit/torch/test_batch.py
```diff
@@ -186,6 +186,69 @@ def test_device(self):
...
+        )
+
+        # with inputs
+        input_batch = Batch(
+            {"feature2": torch.tensor([11, 12])}, {"target1": torch.tensor([13, 14])}, sequences
+        )
+        result = batch.flatten_as_dict(input_batch)
+        assert len(result) == 9  # input keys are considered
+        assert (
+            len([k for k in result if k.startswith("inputs.")]) == 4
```
sararb, line 217: what is the difference between `inputs.` and `features.` keys ?

tests/unit/torch/test_block.py
```diff
@@ -377,3 +377,60 @@ def forward(self, inputs: Dict[str, torch.Tensor]) -> Tuple[torch.Tensor, torch.
...
+        feat, target = torch.tensor([1, 2]), torch.tensor([3, 4])
+        outputs = module_utils.module_test(mm.BatchBlock(mm.BatchBlock()), feat, targets=target)
+
+        assert isinstance(outputs, mm.Batch)
+        assert torch.equal(outputs.feature(), feat)
+        assert torch.equal(outputs.target(), target)
+
+    def test_in_parallel(self):
+        feat, target = torch.tensor([1, 2]), torch.tensor([3, 4])
+        outputs = module_utils.module_test(
```
sararb, line 419: +1 to how we can now create different preprocessing blocks for each subset of the inputs! I love it!!

sararb (approved):
The PR looks good to me! I've just left some minor comments/questions

merlin/models/torch/batch.py
```diff
@@ -277,6 +295,106 @@ def target(self, name: str = "default") -> torch.Tensor:
...
+            Another Batch object to include in the flattening process. The keys from the input
+            batch are also added with a prefix of "inputs.", by default None
+
+        Returns
+        -------
+        Dict[str, torch.Tensor]
+            A dictionary containing all the flattened features, targets, and sequences.
+        """
+        flat_dict: Dict[str, torch.Tensor] = self._flatten()
+        dummy_tensor = torch.tensor(0)
```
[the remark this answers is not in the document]
marcromeyn: We never use it, but we need to have it in order to keep the type of the `Dict[str, torch.Tensor]`. We could store the original value but that might take more memory, so that's why I added the dummy

merlin/models/torch/models/base.py
```diff
@@ -128,9 +140,11 @@ def forward(
         self, inputs: Union[torch.Tensor, Dict[str, torch.Tensor]], batch: Optional[Batch] = None
     ):
         """Performs a forward pass through the model."""
+        _batch = self.pre(inputs, batch=batch)
+
         outputs = inputs
         for block in self.values:
```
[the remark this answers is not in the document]
marcromeyn: Good point! Lemme update that

PR #1193, opened 2023-07-10
Adding CrossAttentionBlock

marcromeyn:
### Goals :soccer:
This PR adds `CrossAttentioBlock`. This would be a better fit to use with session-based models then broadcasting to sequence, since that leads to a bunch of redundant computation. Cross-attention is also a core building-block for multi-model models like Flamingo:

image

### DIFF
--- a/merlin/models/torch/blocks/attention.py
+++ b/merlin/models/torch/blocks/attention.py
@@ -13,6 +13,25 @@ class CrossAttentionBlock(Block):
     Cross Attention Block module which performs a multihead attention operation
     on a provided context and sequence.
 
+    Note this block assumes that the input and output tensors are provided as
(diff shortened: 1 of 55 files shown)


merlin/models/torch/blocks/attention.py
```diff
@@ -0,0 +1,149 @@
...
+
+import torch
+from torch import nn
+
+from merlin.models.torch.batch import Batch
+from merlin.models.torch.block import Block
+
+
+class CrossAttentionBlock(Block):
+    """
```
edknv, line 12: It would be nice to have some example usage in the docstring because things like having to set `batch_first=True` can only be learned by digging into unit tests.

edknv (approved):

PR #1194, opened 2023-07-10
Add functional-methods to make it easier to transform blocks & models.

marcromeyn:
### Goals :soccer:
This PR adds methods like `map`, `filter`, `walk` to our blocks. This makes it easier to transform models. This would enable things like adapters, something like:

```python
def llora(module):
    if isintance(module, (nn.Linear, nn.Embedding)):
        return LoRALayer(module)
​
    return module
​
with_adapters = Block(....).walk(llora)
```

### DIFF
--- a/tests/unit/torch/test_container.py
+++ b/tests/unit/torch/test_container.py
@@ -36,16 +36,16 @@ class TestContainerMixin:
         assert len(filtered) == 1
         assert isinstance(filtered[0], nn.ReLU)
 
-    # def test_filter_recurse(self, container):
-    #     def func(module):
-    #         return isinstance(module, nn.Linear)
+    def test_filter_recurse(self, container):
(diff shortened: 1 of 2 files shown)


tests/unit/torch/test_container.py
```diff
@@ -23,6 +25,182 @@
...
+        def func(module):
+            if isinstance(module, nn.Linear):
+                return nn.Conv2d(3, 3, 3)
+
+        chosen = container.choose(func)
+        assert isinstance(chosen, BlockContainer)
+        assert len(chosen) == 2
+        assert isinstance(chosen[0], nn.Conv2d)
+
+    # def test_choose_recurse(self, container):
```
edknv, line 142: Why was this test commented out?

edknv (approved):

PR #1198, opened 2023-07-11
Makes RetrievalModelV2 support item tower with transforms (e.g. pre-trained embeddings)

gabrielspmoreira:
### Goals :soccer:

- Currently the methods based on `batch_predict()` support only a `Dataset` as input. That prevents performing `batch_predict()` using a `Loader`, which might contain transforms that are important to the model
- A particular use case not supported because of that limitation is generating the item embeddings from the item tower of `RetrievalModelV2`, when among item features there are pre-trained embeddings loaded with `EmbeddingOperator` transform.
- This PR, adds support to `Loader` with `transforms` to all `batch_predict()` based methods (e.g. `to_top_k_encoder()`.

- The `batch_predict()` method and associated methods were changed to accept `Loader` besides `Dataset`. When a `Loader` is passed, it cascades any transforms (like the `EmbeddingOperator`) to the `Loader` that is created inside the `data_iterator_func`. 
- It also checks if the model schema matches  the Loader output schema (which is different from the dataset schema).

### Testing Details :mag:

- Added the `test_two_tower_model_topk_evaluation_with_pretrained_emb` to ensure that it is possible to have pre-trained embeddings in the item tower and convert the `retrieval_model.to_top_k_encoder()`.


rnyak:
@oliverholworthy hello. do you think you can review this PR? this is an important feature that'd be needed by users and us. thanks.

merlin/models/tf/models/base.py
```diff
@@ -1553,7 +1553,7 @@ def predict(
         return out
 
     def batch_predict(
-        self, dataset: merlin.io.Dataset, batch_size: int, **kwargs
+        self, dataset: Union[merlin.io.Dataset, Loader], batch_size: int, **kwargs
```
sararb, line 1556: Can you update the docstrings to expose to the user that we are supporting `Loader` as a possible input dataset

sararb (approved):
The changes lgtm. Thanks for the PR!!

PR #1202, opened 2023-07-13
Add TabularPredictNext for causal item prediction task

sararb:
### Goals :soccer:

- Create the batch-block transformation operator `TabularPredictNext` for defining autoregressive targets by shifting the input sequence feature defined as `target`.
 

- Implement the BatchBlock TabularPredictNext.
- Implement `TabulatSequenceTransform`, the base class that checks the input sequences and target features are in the correct format. 
- Implement `TabularPredictNextModule` responsible for shifting the targets and generating the related masking information. 

Example of usage: 
- For a group of sequences defined in `sequence_schema`, you can specify the target feature to use for creating the targets and all sequential inputs will be transformed accordingly.
```
 transform = TabularPredictNext(schema=sequence_schema, target="a")
batch_output = transform(sequence_batch)
```

- You can also define multiple transforms for different sub-schemas: 
```
        transform_1 = TabularPredictNext(schema=sequence_schema_1, target="a")
        transform_2 = TabularPredictNext(schema=sequence_schema_2, target="c")
        transform_block = mm.BatchBlock(
            mm.ParallelBlock({"transform_1": transform_1, "transform_2": transform_2})
        )
        batch_output = transform_block(sequence_batch)
```


merlin/models/torch/transforms/sequences.py
```diff
@@ -338,3 +338,179 @@ def get_seq_length(self, inputs: Dict[str, torch.Tensor]) -> int:
...
+        for k, v in batch.features.items():
+            if k in self.features:
+                new_inputs[k] = v[:, :-1]
+
+        # Generates information about the new lengths and causal masks
+        new_lengths, causal_masks = {}, {}
+        _max_length = new_target.shape[-1]  # all new targets have same output sequence length
+        for name in self.features:
+            new_lengths[name] = batch.sequences.lengths[name] - 1
+            causal_masks[name] = self._generate_causal_mask(new_lengths[name], _max_length)
```
marcromeyn, line 504: Shouldn't this be the same for all features?

merlin/models/torch/transforms/sequences.py
```diff
@@ -338,3 +338,179 @@ def get_seq_length(self, inputs: Dict[str, torch.Tensor]) -> int:
...
+            self.padding_operator = TabularPadding(
+                schema=self.schema, max_sequence_length=max_sequence_length
+            )
+
+    def initialize_from_schema(self, schema: Schema):
+        self.schema = schema
+        self.features: List[str] = self.schema.column_names
+
+    def _get_target(self, target):
+        return target.column_names
```
marcromeyn, line 437: Shouldn't this be `target.column_names[0]`?

merlin/models/torch/transforms/sequences.py
```diff
@@ -338,3 +338,179 @@ def get_seq_length(self, inputs: Dict[str, torch.Tensor]) -> int:
...
+        schema = Schema(["feature1", "feature2"])
+        next_item_op = TabularPredictNext(
+            schema=schema, target='feature1'
+        )
+        transformed_batch = next_item_op(Batch(features))
+    """
+
+    def __init__(
+        self,
+        target: Union[str, List[str], Tags],
```
marcromeyn, line 378: Might be better to make this of type `Selection` and default to `Tags.ID`?

merlin/models/torch/transforms/sequences.py
sararb: The target is expected to be a list of strings, in case there are multiple target features.

merlin/models/torch/transforms/sequences.py
sararb: There is a requirement in the TabularSequenceTransform block that ensures all features have the same length so the causal mask will be the same for all features. I am iterating over the features because we need to update the corresponding length information as well.

edknv (approved):

PR #1214, opened 2023-10-04
Handle ColumnSchema `target` in serialization of `SequenceTransform`

### LINKED ISSUE #1212: [BUG] Unable to serve a session-based  topk_model on Triton due to TypeError: Unable to serialize item_id-list to JSON. Unrecognized type <class 'merlin.schema.schema.ColumnSchema'>.
### Bug description

I created this bug [ticket](https://github.com/NVIDIA-Merlin/systems/issues/383) on systems and cross linking it here.

The issue might be solved from Merlin models side, not from Systems side. 

The main reason for this bug is that the topk model cannot be saved properly. we get an error like below:

```
...
File /usr/lib/python3.8/json/encoder.py:257, in JSONEncoder.iterencode(self, o, _one_shot)
    252 else:
    253     _iterencode = _make_iterencode(
    254         markers, self.default, _encoder, self.indent, floatstr,
    255         self.key_separator, self.item_separator, self.sort_keys,
    256         self.skipkeys, _one_shot)
--> 257 return _iterencode(o, 0)

TypeError: Unable to serialize item_id-list to JSON. Unrecognized type <class 'merlin.schema.schema.ColumnSchema'>.
```

Please run the code in this [gist](https://gist.github.com/rnyak/609a023b55d593992694f48ecbb90a73) to repro the issue.

### Expected behavior

We should be able to serve a session-based topk_model on Triton and be able to return topK scores and corresponding item-ids

### Environment details

- Merlin version:
- Platform:
- Python version:
- PyTorch version (GPU?):
- Tensorflow version (GPU?):

oliverholworthy:
Fixes #1212 

### Goals :soccer:

Enable serialization of `SequenceTransform` layers. with schema target.

Handle ColumnSchema `target` in serialization of `SequenceTransform`

### DIFF
--- a/merlin/models/tf/transforms/sequence.py
+++ b/merlin/models/tf/transforms/sequence.py
@@ -184,7 +184,10 @@ class SequenceTransform(TabularBlock):
     def get_config(self):
         """Returns the config of the layer as a Python dictionary."""
         config = super().get_config()
-        config["target"] = self.target
+        target = self.target
+        if isinstance(target, ColumnSchema):
+            target = schema_utils.schema_to_tensorflow_metadata_json(Schema([target]))
+        config["target"] = target
 
         return config
 
@@ -193,6 +196,10 @@ class SequenceTransform(TabularBlock):
         """Creates layer from its config. Returning the instance."""
         config = tf_utils.maybe_deserialize_keras_objects(config, ["pre", "post", "aggregation"])
         config["schema"] = schema_utils.tensorflow_metadata_json_to_schema(config["schema"])
+        if config["target"].startswith("{"):  # we have a schema
+            config["target"] = [
+                col for col in schema_utils.tensorflow_metadata_json_to_schema(config["target"])
+            ][0]
         schema = config.pop("schema")
         target = config.pop("target")
         return cls(schema, target, **config)


oliverholworthy:
While this change will enable saving a model with `SequenceTransform` layers with a `target` type of `ColumnSchema`. 

[The example code ](https://gist.github.com/rnyak/609a023b55d593992694f48ecbb90a73) linked in [the related bug](https://github.com/NVIDIA-Merlin/models/issues/1212) issue, shows that the `SequenceMaskLast` layer is only used as part of the `pre` of the `.evaluate` call. This layer then fails to serialize when calling `.save()` on the model, because keras finds a `ColumnSchema` object which it doesn't know what to do with. 

We should check if it's intended that the call to evaluate has a side-effect of adding this `SequenceMaskLast` layer to the model. Since if the `SequenceTransform` layers are  only used in the `pre`, and don't become part of the model, in that usecase, they may not strictly need to be saved/serialized to json correctly.

rnyak:
@gabrielspmoreira @sararb  fyi.

rnyak:
@gabrielspmoreira  and @oliverholworthy I tested this fix and now we can serve topk_model on Triton, however served topK_model returns only top 100 scores which does not make sense without item ids. we might need to fix the code so that it returns both the scores and corresponding item_ids.

rnyak (approved):

PR #1231, opened 2023-12-17
Fix sok import error in ut

EmmaQiaoCh:
SOK has refactored the code path and remove experiment. There are ut failure in tests/unit/tf/horovod/test_embedding.py due to below error:
ImportError: 'horovod' and 'sparse_operation_kit' are required to use SOKEmbedding.


edknv:
Duplicate: https://github.com/NVIDIA-Merlin/models/pull/1232.

PR #1232, opened 2023-12-17
Fix sok import error in ut

EmmaQiaoCh:
SOK has refactored the code path and remove experiment. There are ut failure in tests/unit/tf/horovod/test_embedding.py due to below error:
ImportError: 'horovod' and 'sparse_operation_kit' are required to use SOKEmbedding.

### DIFF
--- a/merlin/models/tf/distributed/backend.py
+++ b/merlin/models/tf/distributed/backend.py
@@ -22,7 +22,7 @@ if hvd_installed:
 
 if HAS_GPU:
     try:
-        from sparse_operation_kit import experiment as sok  # noqa: F401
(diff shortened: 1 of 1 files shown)


minseokl (changes requested):
Please just do `import sparse_operation_kit as sok`

minseokl (approved):

edknv (approved):
The failing tests in CI are unrelated to this PR and are fixed in separate PRs. The failure in `horovod-cpu` is fixed with https://github.com/NVIDIA-Merlin/models/pull/1235 and `docs-build` is fixed with https://github.com/NVIDIA-Merlin/models/pull/1236. `gpu-multi / tensorflow ` is expected to fail since we no longer have access to multi-gpu instances.

